PR #3, opened 2025-05-13
Add an example with prescribed sea surface temperature

glwagner:
In this case we compute momentum, heat, and moisture fluxes using constant coefficients.


glwagner:
https://github.com/user-attachments/assets/f779fc2f-c293-4855-b9d1-281c81af017d

PR #4, opened 2025-05-14
Update the flux boundary conditions to discrete form

mmr0:
The model solution looks the same so hopefully I haven't messed anything up too badly! 

The purpose of this is to enable outputting the surface fluxes - haven't started that part yet.

### DIFF
--- a/examples/prescribed_sst.jl
+++ b/examples/prescribed_sst.jl
@@ -133,9 +133,6 @@ function progress(sim)
     θ = sim.model.tracers.θ
     u, v, w = sim.model.velocities
 
-    # # something like this?
-    # update_surface_fluxes!(sim.model, parameters)
-
     umax = maximum(u)
     vmax = maximum(v)
     wmax = maximum(w)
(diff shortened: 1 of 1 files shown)


glwagner:
looks pretty good to me!

I added a utility to Oceananigans to help you out: https://github.com/CliMA/Oceananigans.jl/pull/4506

perhaps surprisingly, nobody has asked for this before

glwagner:
so for example to get the surface temperature flux you might write

```julia
Jθ = BoundaryConditionOperation(model.tracers.θ, :top, model)
```

the heat flux can then be formed by calculation,

```julia
ρ₀ = AquaSkyLES.base_density(buoyancy) # air density at z=0
cₚ = buoyancy.thermodynamics.dry_air.heat_capacity
Q = ρ₀ * cₚ * Jθ
```

There is also `BoundaryConditionField`, both in `Oceananigans.Models`.

glwagner:
although I am wondering if the heat flux actually requires using the mixture / moist heat capacity. but anyways

glwagner:
ok, the new feature will be available soon: https://github.com/JuliaRegistries/General/pull/130917

PR #6, opened 2025-05-21
Implement anelastic AtmosphereModel

glwagner:
This PR will initially implement an anelastic model with no radiation. Later, we can extend support to Boussinesq (eg providing overlapping capability as Oceananigans + MoistAirBuoyancy) and fully compressible formulations.

I'm using a potential temperature formulation, but we can discuss using a different formulation at some point in the future.

Radiation will be added later.

This is very much WIP but hope to get simulations going soon.


glwagner:
This PR likely requires https://github.com/CliMA/Oceananigans.jl/pull/4583

glwagner:
I'll merge this so we are synced up generally, but everything will continue to be developed

PR #34, opened 2025-10-23
[CI] Improve GitHub Actions setup

giordano:
* install Dependabot
* run CI only when relevant files are modified
* add concurrency group to cancel successive jobs within same PR
* add timeouts
* use default architecture for Julia instead of incorrectly forcing x86-64 everywhere
* update workflow versions
* actually upload code coverage data somewhere

### DIFF
--- a/examples/bomex.jl
+++ b/examples/bomex.jl
@@ -19,7 +19,7 @@ Lx = 6400
 Ly = 6400
 Lz = 3000
 
-arch = CPU() # try changing to GPU()
+arch = CPU() # if changing to GPU() add `using CUDA` above
 stop_time = 6hours
 
 if get(ENV, "CI", "false") == "true" # change values for CI
--- a/examples/thermal_bubble.jl
+++ b/examples/thermal_bubble.jl
@@ -7,7 +7,7 @@ using Printf
 using CairoMakie
 
 # Architecture and grid setup
-arch = CPU() # Change to GPU() for GPU acceleration
+arch = CPU() # if changing to GPU() add `using CUDA` above
 
 # Domain size - following typical thermal bubble studies
 Lx = 20e3  # 20 km horizontal domain


.github/workflows/CI.yml
```diff
@@ -3,39 +3,58 @@ name: CI
...
-      - uses: julia-actions/julia-buildpkg@v1
-
+      - uses: julia-actions/cache@v2
       - uses: julia-actions/julia-runtest@v1
         with:
           coverage: true
+      - uses: julia-actions/julia-processcoverage@v1
+      - uses: codecov/codecov-action@v5
+        with:
+          token: ${{ secrets.CODECOV_TOKEN }}
```
giordano: Someone needs to actually create this token at https://app.codecov.io/account/gh/numericalearth/org-upload-token/ and add it as as an [org-wide actions secret](https://github.com/organizations/NumericalEarth/settings/secrets/actions) called `CODECOV_TOKEN`

.github/workflows/CI.yml
```diff
@@ -3,39 +3,63 @@ name: CI
...
   test:
     name: Julia ${{ matrix.version }} - ${{ matrix.os }} - ${{ matrix.arch }} - ${{ github.event_name }}
     runs-on: ${{ matrix.os }}
+    timeout-minutes: 30
     strategy:
       fail-fast: false
       matrix:
         version:
-          - '1.10'  # Latest release
+          - "min"
```
giordano: For what is worth, tests pass on Julia v1.11 as well.  I know Oceananigans isn't fully compatible with v1.11 yet, but I believe testing multiple versions for a downstream package would still be a good idea, not to get stuck forever to an old version, and this may help keeping Oceananigans up to date without having to test the full upstream package with that version.

giordano: So I can add `"1"`?

glwagner: lets do it

glwagner: we'll find out if it doesn't work...

.github/workflows/CI.yml
giordano: Tokenless uploads are a mess in codecov: https://github.com/codecov/codecov-action/issues/1293

navidcy (approved):

Issue #21, opened 2025-07-17
No clouds are forming in the bomex validation case


mmr0:
We are attempting to use BOMEX observations/LES as a test case for Breeze (latest in [this branch](https://github.com/NumericalEarth/Breeze.jl/tree/glw/fiddle-bomex)). The experimental design is from [Siebesma et al. (2003)](https://doi.org/10.1175/1520-0469(2003)60<1201:ALESIS>2.0.CO;2). Briefly, these are 6-hour runs of marine shallow cumulus convection, with a well-mixed 500m-thick "dry" convective layer topped by a 1000m thick conditionally-unstable cloud layer. These clouds do not form in our Breeze run, and I'm not sure why. 

## Model setup
The model is forced with geostrophic winds, prescribed heat and vapor fluxes at the air-sea interface, and prescribed friction velocity. Other forcings include pressure gradient forcing for geostrophic balance, prescribed radiative cooling and drying profiles, as well as a subsidence forcing that introduces tendencies in u, v, pot. temp., and humidity. 

Initial conditions:

<img width="700" height="260" alt="Image" src="https://github.com/user-attachments/assets/34accb89-8ada-4d72-81e8-b388e8a6bc5a" />

The subsidence profile, and "radiative cooling" & "advective drying" forcings:

<img width="700" height="300" alt="Image" src="https://github.com/user-attachments/assets/f9a3e39f-0450-4703-9b9c-7d6ebe856267" />

The tendency terms due to subsidence (as calculated at the first timestep):

<img width="700" height="300" alt="Image" src="https://github.com/user-attachments/assets/02ffb8da-91c9-42b2-87a8-5e767558eafb" />

These all look reasonable. The v tendency is not shown as it is zero at the first timestep since v=0. The subsidence leads to warming/drying as expected. I haven't shown the geostrophic forcing here, but since the wind is steady in time (outside of the actively-mixing lower layer) I'm confident that this is working as expected. 

## Some actual output

This is plane- and time-averaged output (time-averaging over hours 3-6) of: potential temperature; specific humidity (vapor); velocity; specific humidity (liquid). 

<img width="600" height="500" alt="Image" src="https://github.com/user-attachments/assets/59b5ce17-969f-421c-8861-90d035ab4c43" />

Compared to the Siebesma et al paper:
- Our mixed layer (ML) is deeper
- We have no liquid water (i.e. no clouds)
- Our ML is too moist/cloud layer is too dry
[...]

PR #37, opened 2025-10-28
changed (reverted?) geostrophic forcing term 

mmr0:
The geostrophic forcing term should be constant (in time) and depend only on the initial velocity profiles $u_g$ and $v_g$. With this change u and v are quasi-steady.


glwagner:
Also if this is wrong, it is definitely my fault 😂

mmr0:
Hmmm OK definitely what you had is consistent with the text. But also leads to oscillating flow & doesn't (to my mind) make sense - if you have Coriolis and some initial velocity profile $u_g$ & want a geostrophic balance you need a tendency term $fu_g$ in the $y$ momentum equation. The term $-f(u_{avg} - u_g)$ will be zero, initially..

These plots were made with this branch - velocities look good to me? https://github.com/NumericalEarth/Breeze.jl/issues/21#issuecomment-3458724441 

I can look into it later and I'll actually read the paper this time (oops)

glwagner:
> The term will be zero, initially..

this reasoning makes sense to me and I understand why such a setup produces an inertial oscillation. Let us not be governed by ancient typos

glwagner:
@mmr0 feel free to merge

glwagner:
@kaiyuan-cheng see this PR for discussion about the geostrophic forcing term

PR #40, opened 2025-10-31
Add References via DocumenterCitations + few docs tweaks

navidcy:
I thought it's good practice to start adding proper citations in the docs early on.


glwagner (approved):

navidcy:
Given that we only modified docs, the CI didn't run so I'm merging.

PR #65, opened 2025-11-03
Notational tweak: be explicit about `ℒ(T=0)`; no need to define new subscript `ℒ_{0k}`

navidcy:
Question: Do we define cᵖᵛ - cᵖˡ to be Δcˡ or Δcᵖ? The docs/src code uses both. We should homogenise. I believe Δcᵖ.

### DIFF
--- a/docs/src/thermodynamics.md
+++ b/docs/src/thermodynamics.md
@@ -262,7 +262,7 @@ for an ideal gas
 
 where ``pᵛ⁺`` is saturation vapor pressure, ``T`` is temperature, ``Rᵛ`` is the specific
 gas constant for vapor, ``ℒ^β(T)`` is the latent heat of the transition from vapor to the
-``β`` phase (e.g. ``l ≡ β`` for vapor → liquid and ``i ≡ β`` for vapor to ice).
+``β`` phase (e.g., ``β = l`` for vapor → liquid and ``β = i`` for vapor to ice).
 
 For a thermodynamic formulation that uses constant (i.e. temperature-independent) specific
 heats, the latent heat of a phase transition is linear in temperature.


giordano:
Maybe only exclude windows with julia min for the time being? 1.12 is working OK so far.

src/Thermodynamics/vapor_saturation.jl
```diff
@@ -8,14 +8,17 @@ Compute the saturation vapor pressure ``pᵛ⁺`` over a fluid surface at
...
 The saturation vapor pressure ``pᵛ⁺`` is obtained after integrating the above from
 the triple point, i.e., ``p(Tᵗʳ) = pᵗʳ`` to get
 
 ```math
-pᵛ⁺(T) = pᵗʳ \\left ( \\frac{T}{Tᵗʳ} \\right )^{Δcᵖ / Rᵛ} \\exp \\left [ (1/Tᵗʳ - 1/T) ℒᵛ(T=0) / Rᵛ \\right ] .
+pᵛ⁺(T) = pᵗʳ \\left ( \\frac{T}{Tᵗʳ} \\right )^{Δcˡ / Rᵛ} \\exp \\left [ (1/Tᵗʳ - 1/T) ℒᵛ(T=0) / Rᵛ \\right ] .
 ```
+
+Note that latent heat ``ℒ₀`` is at the reference temperature ``T₀``.
+We can get ``ℒ(T=0) = ℒ₀ - Δcˡ T₀``.
```
glwagner:
```suggestion
Note that latent heat ``ℒ₀`` is at the reference temperature ``T₀``
and that ``ℒ(T=0) = ℒ₀ - Δcˡ T₀``.
```

glwagner:
Realized that notation for saturation vapor pressure should indicate that its over liquid. We'll also have a vapor pressure over ice, as well as a mixture of liquid and ice.

PR #70, opened 2025-11-05
Run tests with ParallelTestRunner

giordano:
This should slightly speed up running the tests as they can run in parallel, and you can also filter them with
```julia
using Pkg
Pkg.test("Breeze"; test_args=`atmosphere_model_unit`)
```
to run only a subset

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -59,9 +59,21 @@ jobs:
           arch: ${{ matrix.arch }}
       - uses: julia-actions/cache@v2
         id: julia-cache
+      - name: "Set test arguments"
+        shell: bash
+        run: |
+          # Run tests in verbose mode
+          TEST_ARGS=(--verbose)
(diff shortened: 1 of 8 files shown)


glwagner:
> This should slightly speed up running the tests as they can run in parallel, and you can also filter them with
> 
> ```julia
> using Pkg
> Pkg.test("Breeze"; test_args=`atmosphere_model_unit`)
> ```
> 
> to run only a subset

Oh, that's much better.

to clarify the test name needs to be enclosed in backticks? `test_name` ?

giordano:
Test jobs seem to be already ~30 seconds faster than on `main` :slightly_smiling_face: 

> to clarify the test name needs to be enclosed in backticks? `test_name` ?

Yes, backticks or double quotes, I think `test_args` accepts both of them.  And the name (or names, you can have more than one) is matched with `startswith` (the `.jl` extension is excluded).

PR #74, opened 2025-11-06
More quality assurance: `ExplicitImports`

giordano:


giordano:
> Does this auto-correct or should it be manually corrected?

I think you need to correct manually, I didn't see a way to automatically fix it, but one can use [pre-commit](https://github.com/JuliaTesting/ExplicitImports.jl/blob/dbed49f1bbe824d536b4b121580441437a1de11c/README.md#pre-commit-hooks) to be alerted when making a commit.  However I found
```julia
using ExplicitImports, Breeze
print_explicit_imports(Breeze)
```
super-useful for getting a clear list of things to correct (I ignored the non-public stuff for the time being), see the [output example](https://github.com/JuliaTesting/ExplicitImports.jl/blob/dbed49f1bbe824d536b4b121580441437a1de11c/README.md#example) in the README

> It might make sense to draw up some docs on what's expected from contributors and what is going to cause something like this to fail? So that people know what to do.

Ok, I can start some devdocs, but tomorrow or later this week at this point 😴

giordano:
I'm going to merge this now, but will add docs in a follow up PR.

PR #29, opened 2025-09-20
Refactor tests; implement anelastic pressure solver tests

glwagner:

PR #76, opened 2025-11-06
[Dycore] Implement tendency term for moist static energy following Pauluis (2008)

glwagner:
Yet another orphaned change from #29; if we want to evolve moist static energy, we need to include the buoyancy flux term in the energy tendency. See equation 15 in [Pauluis 2008](file:///Users/gregorywagner/Downloads/atsc-2007jas2475.1%20(6).pdf).

@kaiyuan-cheng probably worth revisiting the validation tests on this branch.

### DIFF
--- a/src/AtmosphereModels/anelastic_formulation.jl
+++ b/src/AtmosphereModels/anelastic_formulation.jl
@@ -146,7 +146,7 @@ function formulation_pressure_solver(anelastic_formulation::AnelasticFormulation
     reference_density = anelastic_formulation.reference_density
     tridiagonal_formulation = AnelasticTridiagonalSolverFormulation(reference_density)
 
-    solver = if grid isa Oceananigans.Grids.ImmersedBoundaryGrid
+    solver = if grid isa Oceananigans.ImmersedBoundaries.ImmersedBoundaryGrid
         # With this method, we are using an approximate solver that
         # will produce a divergent velocity field near terrain.
         FourierTridiagonalPoissonSolver(grid.underlying_grid; tridiagonal_formulation)


src/AtmosphereModels/update_atmosphere_model_state.jl
```diff
@@ -242,23 +249,43 @@ end
...
+                                 condensates,
+                                 microphysics)
+
+
+    ρᵣwbᶜᶜᶜ = ℑzᵃᵃᶜ(i, j, k, grid, ρᵣwbᶜᶜᶠ, velocities.w, reference_density,
+                    temperature, specific_humidity, formulation, thermo)
 
     return ( - div_Uc(i, j, k, grid, advection, velocities, energy)
-             + microphysical_energy_tendency(i, j, k, grid, formulation, microphysics, condensates)
+             + ρᵣwbᶜᶜᶜ
```
glwagner: @kaiyuan-cheng this is the necessary term

I think a free convection example is sufficient to reveal the necessity of this.

glwagner:
For documentation, I'm running this simple free convection case:

```julia
using Breeze
using Printf
using GLMakie

Nx = Nz = 128
Lz = 4 * 1024
grid = RectilinearGrid(size=(Nx, Nz), x=(0, 2Lz), z=(0, Lz), topology=(Periodic, Flat, Bounded))

e_bcs = FieldBoundaryConditions(bottom=FluxBoundaryCondition(1000))
model = AtmosphereModel(grid, advection=WENO(), boundary_conditions=(; e=e_bcs))

Tₛ = model.formulation.constants.reference_potential_temperature
g = model.thermodynamics.gravitational_acceleration
N² = 1e-6
dθdz = N² * Tₛ / g  # Background potential temperature gradient
θᵢ(x, z) = Tₛ + dθdz * z # background stratification
Ξᵢ(x, z) = 1e-2 * randn()
set!(model, θ=θᵢ, u=Ξᵢ, v=Ξᵢ)

simulation = Simulation(model, Δt=10, stop_iteration=500)
progress(sim) = @info string(iteration(sim), ": t = ", time(sim))
add_callback!(simulation, progress, IterationInterval(100))
run!(simulation)
heatmap(model.velocities.w)
```

on`main` this produces

Screenshot 2025-11-06 at 5 03 19 AM

this is wrong because we should not get convection/flow developing at the _top_ of the domain. Here we are heating from the bottom.

on this PR we get

Screenshot 2025-11-06 at 5 09 17 AM

Unfortunately while this is a bit improved (less convection at the top) it is still not right, so we have a little work to do still.

PR #77, opened 2025-11-06
[CI] Rework docs deployment

giordano:


giordano:
I have a fix :rocket: 

image

navidcy:
Great!

navidcy (approved):
Why lots/most of the doc src files are deleted? Is it still debugging mode?

docs/make.jl
```diff
@@ -29,9 +29,13 @@ makedocs(sitename="Breeze",
...
+deploydocs(
+    ;
     repo = "github.com/NumericalEarth/Breeze.jl",
+    deploy_repo = "github.com/NumericalEarth/BreezeDocumentation",
     devbranch = "main",
-    push_preview = true,
-    forcepush = true
+    # Only push previews if all the relevant environment variables are non-empty. This is an
+    # attempt to work around https://github.com/JuliaDocs/Documenter.jl/issues/2048.
+    push_preview = all(!isempty, (get(ENV, "GITHUB_TOKEN", ""), get(ENV, "DOCUMENTER_KEY", ""))),
```
giordano: I'd need to wait for a PR by dependabot to be sure this is working correctly, but looking at https://github.com/NumericalEarth/Breeze.jl/actions/runs/19032077021/job/54348250773?pr=61#step:5:5 `GITHUB_TOKEN` isn't empty, but `DOCUMENTER_KEY` is, so making sure they're both non-empty should be a decent approach.

giordano (commented):
Ok, I think this is ready now!  The commit status won't work until https://github.com/JuliaDocs/Documenter.jl/pull/2816 makes it into a new release, but we can't do too much about it for the time being (apart from temporarily using a fork, but I'd rather avoid that if possible)

.github/workflows/Documentation.yml
```diff
@@ -48,10 +48,16 @@ jobs:
...
+        id: generate_token
+        if: ${{ ! github.event.pull_request.head.repo.fork }}
+        with:
+          app-id: "${{ secrets.NUMTERRA_BOT_ID }}"
+          private-key: "${{ secrets.NUMTERRA_BOT_PRIVATE_KEY }}"
       - uses: julia-actions/julia-docdeploy@v1
         env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-          DOCUMENTER_KEY: ${{ secrets.DOCUMENTER_KEY }}
+          GITHUB_TOKEN: ${{ steps.generate_token.outputs.token }}
```
giordano: On a second thought, I'm not entirely sure an app is needed for this, pushing to the other repo happens via `DOCUMENTER_KEY`.  Let me try that.

glwagner:
eeeexcellent

PR #79, opened 2025-11-06
Large runner /w GPU workflow

bischtob:
Test the workflow


bischtob:
GPU runners work now! @giordano

PR #84, opened 2025-11-07
Fixes gravity so that it points down

navidcy:
This was a typo, right?

### DIFF
--- a/docs/src/dynamics.md
+++ b/docs/src/dynamics.md
@@ -1,19 +1,13 @@
 # Dynamics
 
-This section summarizes the governing equations behind Breeze’s atmospheric dynamics and the
-anelastic formulation used by [`AtmosphereModel`](@ref), following the thermodynamically consistent
-framework of [Pauluis2008](@citet).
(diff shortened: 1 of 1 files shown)


bischtob:
:D

navidcy:
@glwagner did you check the buoyancy term?

glwagner:
here's how we write it:

https://github.com/NumericalEarth/Breeze.jl/blob/ceb373e0e403a2172f526003564fa5ca17c45c9b/src/AtmosphereModels/dynamics_kernel_functions.jl#L9-L14

`α` is the specific volume or 1 / rho. When density is smaller, `α` is larger, and thus `b` is larger. This produces a positive acceleration in the vertical momentum equation:

https://github.com/NumericalEarth/Breeze.jl/blob/ceb373e0e403a2172f526003564fa5ca17c45c9b/src/AtmosphereModels/dynamics_kernel_functions.jl#L81

https://github.com/NumericalEarth/Breeze.jl/blob/ceb373e0e403a2172f526003564fa5ca17c45c9b/src/AtmosphereModels/dynamics_kernel_functions.jl#L16-L20

glwagner:
Note this is anelastic specific, so we will have a new way to write buoyancy for fully compressible. I think.

PR #85, opened 2025-11-07
Make quick start example an example again

glwagner:


giordano (approved):
For the record, see discussion at https://github.com/NumericalEarth/Breeze.jl/pull/76#discussion_r2502062103

PR #97, opened 2025-11-09
[docs] Tip about running tests without GPU

giordano:
Ref: https://github.com/NumericalEarth/Breeze.jl/pull/93#discussion_r2507378297.

### DIFF
--- a/docs/src/contributing.md
+++ b/docs/src/contributing.md
@@ -63,6 +63,22 @@ Pkg.test("Breeze"; test_args=`doctests`)
     Pkg.test("Breeze"; test_args=`--list`)
     ```
 
+### GPU tests
+
+When running the tests, if a CUDA GPU is detected, they automatically use the [`GPU` Oceananigans architecture](https://clima.github.io/OceananigansDocumentation/stable/appendix/library/#Oceananigans.Architectures.GPU), otherwise they run on [`CPU`](https://clima.github.io/OceananigansDocumentation/stable/appendix/library/#Oceananigans.Architectures.CPU).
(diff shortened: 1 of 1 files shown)


glwagner:
We had some issues with the "implicit run on GPU" method over at Oceananigans which is why we switched to an explicit env variable for the tests. Do we think maybe this won't be a concern here? I can't remember the details, @simone-silvestri maybe can.

glwagner (approved):
its helpful!

giordano:
Ok, I'm going to merge this, if we'll need a dedicated variable for enabling/disabling GPU tests it can be done in a follow up PR (and update the docs accordingly!).  Basically we'd need to revise the logic at https://github.com/NumericalEarth/Breeze.jl/blob/155344abedf5a8739202c5faac275c2ca2576680/test/runtests.jl#L18 But I think this PR is still helpful in pointing contributors to the use of `default_arch` in the tests

glwagner:
agree, progress!

glwagner:
I think the issue is addressed by the error. The big problem we had once was that the test stopped running on GPU silently, and we didn't realize it (bc CUDA stopped working on the ci machine, so the tests just silently ran on CPU). Since you put in a variable that is basically like "error if not on GPU", it achieves the same thing. ✔️

giordano:
In #93 I had also considered showing the value of `default_arch` at the beginning of the tests, but then changed my mind because I thought I was complicating my life (and `BREEZE_ENSURE_CUDA_FUNCTIONAL` was already a decent barrier).  I'm happy to try to do that again as well.

PR #98, opened 2025-11-09
Tweaks in latex rendering in Docs/Dynamics

navidcy:


docs/src/dynamics.md
giordano: Again unrelated to this PR, but I find the title of this page confusing because it's very different from what's written in the navigation sidebar: https://github.com/NumericalEarth/Breeze.jl/blob/155344abedf5a8739202c5faac275c2ca2576680/docs/make.jl#L25 When I followed the link I had to double check I was on the right page. Also, it took me a while to find the "dynamics" page from the sidebar, since there's no page ostensibly called like that. CC @glwagner

glwagner: Ah yes good point. @navidcy what do you think about calling this page "dycore equations and algorithms"?

I want the section to be more specific than "governing equations", because there will be a ton of equations associated with microphysics, radiation, and parameterizations, so we need to distinguish between them.

"Dycore" is short for "dynamical core", so "Dynamics" also works, eg "Dynamical equations and algorithms"

glwagner (approved):
Nice, thank you! I think we can also render the subscript "r". this is great though.

PR #100, opened 2025-11-09
[examples] Tropical Cyclone World via Cronin and Chavas (2019)

### LINKED ISSUE #234: A reference state close to the initial condition
The current design uses an isentropic atmospheric profile as the reference state, which may cause problems when simulating a non-isentropic atmosphere. (An isentropic profile is highly idealized and not representative of real atmospheric conditions.) I suspect this may be the root cause of why we aren’t getting the correct internal gravity wave, and possibly the same issue affecting the supercell case.

Could we instead use a reference state that is closer to the initial condition? For example, the model could take a user-defined atmospheric profile, such as a sounding or the horizontal average of a 3D realistic atmosphere, adjust it to be hydrostatically balanced, and use that as the reference state. I believe this is how CM1 handles the reference state in its anelastic solver.

glwagner:
This PR draws up an literate'd "TC world" example following [Cronin and Chavas 2019](https://journals.ametsoc.org/downloadpdf/view/journals/atsc/76/8/jas-d-18-0357.1.pdf) setup.

Getting the TC world example to run stably with Float32 required changing how we compute buoyancy for `AnelasticDynamics` -- it is now computed in a way that is friendlier to lower precision (cc @giordano). This fixes an issue where the example was running with F64 but blowing up with F32 and I believe improves the quality of the code overall (cc @bischtob, @nrb171 ).

This PR also added capabiltiies for moist reference states and computes the reference state by integrating from specified temperature and moisture profiles, determining density and pressure simultaneously. It therefore supercedes #185 .

It also adds the ability to compute the reference profile from the mean state of `AtmosphereModel`. I believe this can therefore close #234 -- the procedure to do this involves first setting the initial condition, and then calling `set_to_mean!(reference_state, model)`. cc @kaiyuan-cheng

### DIFF
--- a/.github/workflows/Documentation.yml
+++ b/.github/workflows/Documentation.yml
@@ -37,7 +37,7 @@ jobs:
     container:
       image: 'ghcr.io/numericalearth/breeze-docker-images:docs-julia_1.12.4'
       options: --gpus=all
-    timeout-minutes: 60
+    timeout-minutes: 75
     steps:
       - uses: actions/checkout@v6
       - name: Set BREEZE_BUILD_ALL_EXAMPLES
@@ -62,7 +62,7 @@ jobs:
           git config --global --add safe.directory ${PWD}
       - name: Build documentation
         shell: bash
-        timeout-minutes: 45
+        timeout-minutes: 60
         run:
           earlyoom -m 3 -s 100 -r 300 --prefer 'julia' &
           julia --color=yes --project=docs --threads 1,$(nproc) docs/make.jl
@@ -72,7 +72,7 @@ jobs:
         with:
           name: documentation-build
           path: docs/build
-          retention-days: 10
+          retention-days: 15
 
   deploy-docs:
     # We want to deploy docs if:


examples/tropical_cyclone_world_v2.jl
```diff
@@ -0,0 +1,628 @@
...
+# resolution and a 28 km model top. We use reduced resolution for faster runs.
+
+arch = GPU()
+
+# Domain size (paper values)
+Lx = Ly = 1152e3  # 1152 km
+H  = 28e3         # 28 km model top
+
+# Resolution (reduced for CI/documentation — increase for science runs)
+Nx = Ny = 576     # Paper resolution: 2 km spacing
```
glwagner: I think we could consider running a 10km version of this case as an example.

giordano:
The docs build worked and the preview is at https://numericalearth.github.io/BreezeDocumentation/previews/PR100/literated/tropical_cyclone_world/.  However the [visualisation](https://numericalearth.github.io/BreezeDocumentation/previews/PR100/literated/tropical_cyclone_world/#Visualization) didn't go very well, because in Literate a "cell" can't end in the middle of an `if` (otherwise the expression would be incomplete), so that part needs to be rearranged a little bit.

giordano:
With latest version I get
```
[ Info: Iter 26000, t = 1.077 days, Δt = 3.654 seconds, max|u,v,w| = (398.7, 723.2, 20.8) m/s, θ ∈ [218.3, 2313.5] K
[ Info: Iter 26500, t = 1.098 days, Δt = 3.300 seconds, max|u,v,w| = (118.0, 853.0, 23.1) m/s, θ ∈ [216.0, 2275.7] K
[ Info: Iter 27000, t = 1.119 days, Δt = 3.567 seconds, max|u,v,w| = (426.4, 716.5, 22.3) m/s, θ ∈ [216.7, 2274.6] K
[ Info: Iter 27500, t = 1.139 days, Δt = 3.353 seconds, max|u,v,w| = (786.8, 396.5, 26.3) m/s, θ ∈ [214.6, 2303.0] K
[ Info: Iter 28000, t = 1.159 days, Δt = 3.569 seconds, max|u,v,w| = (1027.7, 127.9, 23.4) m/s, θ ∈ [220.3, 2325.3] K
[ Info: time = NaN, iteration = 28300: NaN found in field ρu. Stopping simulation.
```
`NaN` time? 🤨

Also, these don't look great, I think:

https://github.com/user-attachments/assets/a6f85c31-7e0e-4444-b940-ac5526e3d9bd

tc_world_profiles
tc_world_surface_winds

glwagner:
I AM INSTALLING THE PRECOMMIT HOOK

giordano:
Looking better now: https://numericalearth.github.io/BreezeDocumentation/previews/PR100/literated/tropical_cyclone_world/#Results:-mean-profile-evolution? 👀

docs/make.jl
```diff
@@ -28,19 +28,20 @@ struct Example
...
+    # Example("Shallow cumulus convection (BOMEX)", "bomex", true),
+    # Example("Precipitating shallow cumulus (RICO)", "rico", false),
+    # Example("Convection over prescribed sea surface temperature (SST)", "prescribed_sea_surface_temperature", true),
+    # Example("Inertia gravity wave", "inertia_gravity_wave", true),
+    # Example("Single column radiation", "single_column_radiation", true),
+    # Example("Stationary parcel model", "stationary_parcel_model", true),
+    # Example("Rising parcel: adiabatic ascent", "rising_parcels", true),
+    # Example("Acoustic wave in shear layer", "acoustic_wave", true),
+    # Example("Cloud formation in prescribed updraft", "kinematic_driver", true),
+    # Example("Splitting supercell", "splitting_supercell", false),
```
giordano: Reminder to revert this before merging

src/Thermodynamics/reference_states.jl
```diff
@@ -107,12 +120,100 @@ the density of dry air at the reference pressure and temperature.
...
+        @inbounds ρᵣ[1, 1, k] = pᵏ / RᵐTᵏ
+
+        z⁻   = zᵏ
+        RᵐT⁻ = RᵐTᵏ
+        p⁻   = pᵏ
+    end
+end
+
+"""
+    compute_hydrostatic_reference!(ref::ReferenceState, constants)
```
giordano:
```suggestion
$(TYPEDSIGNATURES)
```

src/Thermodynamics/reference_states.jl
```diff
@@ -124,28 +225,114 @@ Keyword arguments
...
+# set! and fill_halo_regions! are no-ops for ZeroField
+set_reference_field!(field, value) = (set!(field, value); fill_halo_regions!(field); nothing)
+set_reference_field!(::ZeroField, value) = nothing
+
+#####
+##### Recompute reference state from profiles
+#####
+
+"""
+    compute_reference_state!(reference_state, T̄, q̄ᵛ, q̄ˡ, q̄ⁱ, constants)
```
giordano:
```suggestion
$(TYPEDSIGNATURES)
```

src/Thermodynamics/reference_states.jl
```diff
@@ -124,28 +225,114 @@ Keyword arguments
...
+    fill_halo_regions!(ref.temperature)
+    set_reference_field!(ref.vapor_mass_fraction, q̄ᵛ)
+    set_reference_field!(ref.liquid_mass_fraction, q̄ˡ)
+    set_reference_field!(ref.ice_mass_fraction, q̄ⁱ)
+    compute_hydrostatic_reference!(ref, constants)
+    return nothing
+end
+
+"""
+    compute_reference_state!(reference_state, T̄, q̄ᵗ, constants)
```
giordano:
```suggestion
$(TYPEDSIGNATURES)
```

examples/tropical_cyclone_world.jl
```diff
@@ -0,0 +1,377 @@
...
+    z  = [0, 1000, 3500, H],
+    Δz = [Δz_fine, Δz_fine, Δz_coarse, Δz_coarse])
+
+Nz = length(z) - 1
+
+grid = RectilinearGrid(arch; size = (Nx, Ny, Nz), halo = (5, 5, 5),
+                       x = (0, Lx), y = (0, Ly), z,
+                       topology = (Periodic, Periodic, Bounded))
+
+@show grid
```
giordano: Can probably remove this `@show`?  The grid would be shown anyway since it's at the end of a cell.
```suggestion
```

glwagner: done. yes, I added this for debugging. I am running the case again, will push when I am happy with figures.

glwagner:
> Looking better now: https://numericalearth.github.io/BreezeDocumentation/previews/PR100/literated/tropical_cyclone_world/#Results:-mean-profile-evolution? 👀

way better!

PR #104, opened 2025-11-09
Retest old method for saturation adjustment initial guess for Boussinesq model

glwagner:


glwagner:
So, this change causes our test to fail: https://github.com/NumericalEarth/Breeze.jl/actions/runs/19212135251/job/54915944368?pr=104#step:6:420

I am confused how, surely there are not multiple roots to the adjustment equation? There could be a mistake in the residual expression, or an inconsistency with how the thermodynamics are implemented.

PR #107, opened 2025-11-09
[CI] Split docs building job in two parts

giordano:
In this way we can have the global concurrency group only for the docs publishing part, and let the docs building job run freely instead of waiting forever for everyone else.  The advantage of splitting the jobs is also that if only the deploy fails/is cancelled, restarting it should be quick because we reuse the artifact, instead of having to rebuild the world.

Ref: https://github.com/NumericalEarth/Breeze.jl/pull/102#issuecomment-3508702176.

### DIFF
--- a/.github/workflows/Documentation.yml
+++ b/.github/workflows/Documentation.yml
@@ -88,6 +88,7 @@ jobs:
       - uses: actions/download-artifact@v5
         with:
           name: documentation-build
+          path: docs/build
       - name: Instantiate docs environment
         shell: julia --color=yes {0}
         run: |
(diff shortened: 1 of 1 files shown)


giordano:
And the deploy job is now correctly waiting for another job in the concurrency group

image

But the build artifact is already available at https://github.com/NumericalEarth/Breeze.jl/actions/runs/19214186805?pr=107, so one can immediately see the output (and whether docs building is successful or not) instead of waiting forever

navidcy (approved):

giordano:
https://numericalearth.github.io/BreezeDocumentation/previews/PR107/

navidcy:
wooooooo!

PR #109, opened 2025-11-09
Move docstring from kernel to method for the pressure_correction update

navidcy:


giordano (approved):
I was a bit surprised there's was a docstring attached to a `@kernel` function with an internal looking name, makes sense!

Sorry, I cancelled the docs job because I really want to get #107 in as soon as possible, with multiple PRs all going on at the same time docs building is awful, will restart afterwards!

PR #117, opened 2025-11-10
Update thermal bubble example

navidcy:
This PR uses a bit more atmosphere-specific grid + runs the example for a bit longer.

preview at https://numericalearth.github.io/BreezeDocumentation/previews/PR117/literated/thermal_bubble/


giordano:
> We need to trigger the Docs build when we modify the examples?

Now yes, they weren't used for the docs before.

PR #118, opened 2025-11-10
Add high-res thermal bubble animation in README

navidcy:
Added example output links for higher resolution.


glwagner:
I suggest getting rid of free convection and having this as the showcase

navidcy:
Done!

PR #120, opened 2025-11-10
Fix bug in saturation adjustment solver for MoistAirBuoyancy

glwagner:
This PR fixes a bug in the saturation adjustment solver that only gets hit under certain conditions: when the secant iteration accesses unsaturated states during the course of finding a saturated state. Before this PR, we assume that during adjustment the state was saturated. However, this assumption does not hold if "intermediate" (eg test) states that are accessed during the iteration are not saturated. The result of the bug is a spurious change in the total moisture fraction. It is fixed in this PR.


glwagner:
i think we should build regression tests from the canonical cases (eg bomex, and there are another dozen that we will implement)

navidcy (approved):
@glwagner, your call whether to merge or not!

PR #122, opened 2025-11-10
[CI] Temporarily use dev version of Documenter

giordano:
This is mainly for testing for the moment


giordano:
Ok, this was mainly to confirm that https://github.com/JuliaDocs/Documenter.jl/pull/2816 will still work after reorganising the docs jobs in #107, and yes, it works: 
image
The `documenter/deploy` link points to https://numericalearth.github.io/BreezeDocumentation/previews/PR122/ 🥳

giordano:
I really want this feature to make it easier to find the link for the docs, but new version of Documenter is being delayed, so I'm going to merge this PR.  Will revert after new version of Documenter is released

PR #124, opened 2025-11-10
Unified adjustment solver for MoistStaticEnergyState and PotentialTemperatureState + atmos bomex case

glwagner:
This PR unifies the saturation adjustment solver for the two thermodynamic states we support: `MoistStaticEnergyState` and `PotentialTemperatureState`. This is useful, because we now correctly set initial conditions in `AtmosphereModel` based on potential temperature input. This PR also fixes `set!`, which was wrong for moist initial states, and sets up an anelastic bomex case. I think we should work on actually running / validating the bomex case in a future PR.

### DIFF
--- a/test/atmosphere_model_unit.jl
+++ b/test/atmosphere_model_unit.jl
@@ -39,7 +39,7 @@ end
 
     # Initialize with potential temperature and dry air
     θᵢ = CenterField(grid)
-    set!(θᵢ, (x, y, z) -> randn())
+    set!(θᵢ, (x, y, z) -> θ₀ + rand())
     set!(model; θ=θᵢ)
 
     θ_model = Breeze.AtmosphereModels.PotentialTemperatureField(model)


.github/workflows/Validation.yml
giordano: I think you want to change this as well?

glwagner: copy that!

src/Thermodynamics/dynamic_states.jl
```diff
@@ -42,7 +60,25 @@ end
 
 @inline Base.eltype(::MoistStaticEnergyState{FT}) where FT = FT
 @inline total_moisture_mass_fraction(state::MoistStaticEnergyState) = total_moisture_mass_fraction(state.moisture_mass_fractions)
+@inline is_absolute_zero(𝒰::MoistStaticEnergyState) = 𝒰.moist_static_energy == 0
 
 @inline function with_moisture(𝒰::MoistStaticEnergyState, q::MoistureMassFractions)
     return MoistStaticEnergyState(𝒰.moist_static_energy, q, 𝒰.height, 𝒰.reference_pressure)
 end
+
+@inline function temperature(𝒰::MoistStaticEnergyState, thermo::ThermodynamicConstants)
```
kaiyuan-cheng: Duplication? There is a temperature function above already.

kaiyuan-cheng: Apologize for my oversight

test/atmosphere_model_unit.jl
giordano: To get the additional context for the failures 🙂 
```suggestion
for p₀ in (101325, 100000), θ₀ in (288, 300)
    @testset let p₀ = p₀, θ₀ = θ₀
```

glwagner: ahahah

PR #125, opened 2025-11-10
 Change opening docs page to use AtmosphereModel

glwagner:
Let's see if this works

### DIFF
--- a/src/Thermodynamics/reference_states.jl
+++ b/src/Thermodynamics/reference_states.jl
@@ -75,7 +75,7 @@ Keyword arguments
 - `base_pressure`: Default: 101325
 - `potential_temperature`: Default: 288
 """
-function ReferenceState(grid, thermo;
+function ReferenceState(grid, thermo=ThermodynamicConstants(eltype(grid));
                         base_pressure = 101325,
                         potential_temperature = 288)
 


docs/src/index.md
```diff
@@ -24,13 +30,12 @@ Pkg.add("https://github.com/NumericalEarth/Breeze.jl.git")
 
 ## Quick Start
 
-A basic free convection simulation:
+A basic free convection simulation with `AtmosphereModel`:
```
navidcy:
```suggestion
A basic free convection simulation with an `AtmosphereModel`:
```

docs/src/index.md
```diff
@@ -24,13 +30,12 @@ Pkg.add("https://github.com/NumericalEarth/Breeze.jl.git")
...
 
 ```@example
-using Oceananigans
+using Breeze
 using Oceananigans.Units
```
navidcy: I'm wondering if it makes sense to somehow make Breeze re-export some things. E.g., would it make sense if here we could just write:

```suggestion
using Breeze.Units
```

cc @giordano

giordano: Honestly I'm not a fan of that, then you're tying the API of one package A to another package B, with unintended consequences, like code breaking when B gets upgraded without any changes in A

navidcy:
I think this is good to go. I don't know why Windows Julia 1.12 fails but I'm happy to ignore it.

giordano:
That's happening from time to time, no clue either 😞

navidcy:
merge?

Issue #90, opened 2025-11-08
Build docs on GPU + harness more resources?


glwagner:
The documentation is currently the longest running test, and this is going to continue to get worse I think as we add examples. We can move some heavy things to GPU to help, but I believe that it will still be a bottleneck.

Opening this issue to track progress transferring the docs build to GPU. I am also wondering a few additional things:

* Can we pull down more powerful CPU for the job / somehow harness more resources?
* This might be harder, but I am also curious whether it is possible to assemble the docs using a few different jobs. Then for example we can run some examples on parallel jobs, and then somehow transfer the artifacts from those jobs into the Documenter build. I am not sure that Documenter supports this right now (but the situation may have changed since the last time I looked into it)
* We can also consider adding a label like we do for ClimaOcean "build docs", and only running the docs build if (1) the label exists or (2) the PR is merged into main.

glwagner:
> We can also play with the `draft` option, which skips the examples:
> 
> [Breeze.jl/docs/make.jl](https://github.com/NumericalEarth/Breeze.jl/blob/6c386a7de923d0d785284952634a2141fc3e5c0d/docs/make.jl#L33)
> 
> Line 33 in [6c386a7](/NumericalEarth/Breeze.jl/commit/6c386a7de923d0d785284952634a2141fc3e5c0d)
> 
>  draft = false, 
> For example we could make it conditional on the value of an env var, which could be set in CI based on a label (and always `true` when on `main`). This can be done independently of the suggestions above.

This is a cool idea! We may want to distinguish between the "literate examples" (heavy heavy computations a la Oceananigans examples) and the more vanilla docs `@example` blocks which I think we should try to keep cheaper. But if we can set env variables based on labels, we can do both in fact with two different labels perhaps.

navidcy:
Was this resolved with #235?

glwagner:
resolved!

PR #134, opened 2025-11-10
Change code block to use @example syntax

navidcy:


docs/src/index.md
```diff
@@ -26,7 +26,7 @@ Pkg.add("https://github.com/NumericalEarth/Breeze.jl.git")
 
 A basic free convection simulation:
 
-```julia
+```@example
```
giordano: This was changed in #105 by @glwagner, I'm still not entirely sure what's the intended use of this 😅

giordano: In particular, see https://github.com/NumericalEarth/Breeze.jl/pull/105#discussion_r2508319005.  If running all the examples is too bothersome in all PRs we can disable them as suggested at https://github.com/NumericalEarth/Breeze.jl/issues/90#issuecomment-3506631109, and we can have some (label-based?) mechanism to run the full docs build on demand in PRs (and every time on `main` anyway)

navidcy: I believe it was changed by accident to compile the docs locally faster? @glwagner?

glwagner (approved):

PR #140, opened 2025-11-11
Support turbulence closures from Oceananigans

glwagner:
Undoubtedly this needs a fair amount of work (thanks a lot chat 🫤) but it is a start towards supporting diffusive fluxes / turbulence closures in AtmosphereModel. Basically we can hook into Oceananigans turbulence closures, provided we multiply by density in the appropriate place. We also have to think about which fields we pass into the turbulence closure functions. For example, Oceananigans expects to get velocities (which as far as I can tell, is also what we want). However, this means that we need to pass auxiliary fields, and not just prognostic fields, into the closures (in Oceananigans, we simply pass prognostic fields).

We will also eventually want to implement atmosphere-specific turbulence closures.

### DIFF
--- a/src/AtmosphereModels/dynamics_kernel_functions.jl
+++ b/src/AtmosphereModels/dynamics_kernel_functions.jl
@@ -5,6 +5,16 @@ using Oceananigans.Operators: ∂xᶠᶜᶜ, ∂yᶜᶠᶜ, ∂zᶜᶜᶠ, ℑz�
 @inline ∂ⱼ_𝒯₁ⱼ(i, j, k, grid, args...) = zero(grid)
 @inline ∂ⱼ_𝒯₂ⱼ(i, j, k, grid, args...) = zero(grid)
 @inline ∂ⱼ_𝒯₃ⱼ(i, j, k, grid, args...) = zero(grid)
+
+"""
+    ∇_dot_Jᶜ(i, j, k, grid, density, closure::AbstractTurbulenceClosure, closure_fields,
+             id, c, clock, model_fields, buoyancy)
+
+Return the discrete divergence of the dynamic scalar flux `Jᶜ = ρ jᶜ`,
+where `jᶜ` is the "kinematic scalar flux", using area-weighted differences divided by cell volume.
+Similar to Oceananigans' `∇_dot_qᶜ` signature with the additional density factor `ρ`, where in
+Oceananigans `qᶜ` is the kinematic tracer flux.
+"""
 @inline ∇_dot_Jᶜ(i, j, k, grid, args...) = zero(grid)
 
 #####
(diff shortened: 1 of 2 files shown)


src/AtmosphereModels/atmosphere_model.jl
```diff
@@ -112,7 +114,7 @@ function AtmosphereModel(grid;
 
     density = materialize_density(formulation, grid)
     velocities, momentum = materialize_momentum_and_velocities(formulation, grid, boundary_conditions)
-    tracers = NamedTuple(n => CenterField(grid, boundary_conditions=boundary_conditions[n]) for name in tracers)
+    tracers = NamedTuple(n => CenterField(grid, boundary_conditions=boundary_conditions[n]) for n in tracers)
```
giordano: Let me guess, this wasn't covered by tests before? Is it now?

glwagner: i can add

docs/src/appendix/notation.md
```diff
@@ -44,6 +44,9 @@ inline math, while the Unicode column shows the exact glyphs used in code.
 | ``p_n``                       | `pₙ`          | Nonhydrostatic pressure (projection/correction potential) |
 | ``\Delta t``                  | `Δt`          | Time step |
 | ``\Delta z``                  | `Δz`          | Vertical grid spacing |
+| ``\boldsymbol{\tau}``         | `τ`           | Kinematic subgrid/viscous stress tensor (per unit mass) |
+| ``\boldsymbol{\mathcal{T}}``  | `𝒯`           | Dynamic stress tensor used in anelastic momentum, ``\mathcal{T} = ρᵣ \, \tau`` |
+| ``\boldsymbol{J}``            | `J`           | Dynamic diffusive flux for scalars, ``\boldsymbol{J} = ρᵣ \, \boldsymbol{\tau}`` |
```
glwagner:
```suggestion
| ``\boldsymbol{J}``            | `J`           | Dynamic diffusive flux for scalars |
```

test/turbulence_closures.jl
```diff
@@ -0,0 +1,26 @@
...
+
+const TC = Breeze.TurbulenceClosures
+
+@testset "Time stepping with TurbulenceClosures [$(FT)]" for FT in (Float32, Float64)
+    Oceananigans.defaults.FloatType = FT
+    grid = RectilinearGrid(default_arch; size=(8, 8, 8), x=(0, 100), y=(0, 100), z=(0, 100))
+
+    # Try two closures from Oceananigans
+    for closure in (SmagorinskyLilly(), AnisotropicMinimumDissipation())
+        @testset let closure=closure
```
giordano: Very good! 🤓

test/turbulence_closures.jl
```diff
@@ -0,0 +1,26 @@
...
+            model = AtmosphereModel(grid; closure=closure)
+            @test try
```
giordano: Why may this fail? Don't we want to see the error?

test/turbulence_closures.jl
glwagner: i'm a good student 📚

test/turbulence_closures.jl
glwagner: whats a good pattern? shoudl we just atke the time-step and then write `@test true` after? basically i want to test that it doesn't throw an error, is all

test/turbulence_closures.jl
```diff
@@ -0,0 +1,45 @@
...
+        set!(model; ρe=ρe₀)
+        time_step!(model, 1)
+
+        ϵ = sqrt(eps(FT))
+        @test model.momentum.ρu ≈ XFaceField(grid) atol=ϵ
+        @test model.momentum.ρv ≈ YFaceField(grid) atol=ϵ
+        @test model.momentum.ρw ≈ ZFaceField(grid) atol=100ϵ # large bc of non-zero buoyancy
+        @test model.moisture_density ≈ CenterField(grid) atol=ϵ
+        @test model.tracers.ρc ≈ CenterField(grid) atol=ϵ
+        @test all(isapprox.(interior(model.energy_density), ρe₀, atol=100ϵ))
```
giordano: I'm always of the opinion that we should use rtol, almost never atol (only at 0) 🙂 `eps` makes sense only around 1

PR #141, opened 2025-11-11
Actually respect user forcing and add a test

glwagner:
This PR adds support for forcing to AtmosphereModel and adds a few tests.

### DIFF
--- a/docs/src/breeze.bib
+++ b/docs/src/breeze.bib
@@ -50,3 +50,14 @@
     doi = {10.1029/2019MS002025},
     year = {2020}
 }
+
+@article{Khairoutdinov2022,
+  title={Global system for atmospheric modeling: Model description and preliminary results},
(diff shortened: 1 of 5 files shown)


test/forcing.jl
```diff
@@ -0,0 +1,60 @@
...
+    forcings = [
+        Returns(one(FT)),
+        Forcing(Returns(one(FT)), discrete_form=true),
+        Forcing(Returns(one(FT)), field_dependencies=:ρu, discrete_form=true),
+        Forcing(Returns(one(FT)), field_dependencies=(:ρe, :ρqᵗ, :ρu), discrete_form=true),
+    ]
+
+    Δt = convert(FT, 1e-6)
+
+    @testset "Forcing forcing ($FT, $(typeof(forcing)))" for forcing in forcings
```
giordano: Is "forcing forcing" intended?

glwagner: no

PR #142, opened 2025-11-11
Build abstraction for computing saturation vapor pressure over mixed phase surfaces

glwagner:
This PR fleshes out our abstractions for computing the saturation vapor pressure. It adds the concept of a "surface" over which the saturation vapor pressure is to be computed, and builds support for

* `PlanarLiquidSurface`
* `PlanarIceSurface`
* `PlanarMixedPhaseSurface` with `PlanarMixedPhaseSurface.liquid_fraction`

This encodes an expectation that we may also want to compute saturation vapor pressure over curved surfaces in the future.

Once we add docs and test for this, we can then start work on mixed phase saturation adjustment, which will use this new feature to compute the saturation specific humidity in a mixed phase situation.

### DIFF
--- a/src/Thermodynamics/vapor_saturation.jl
+++ b/src/Thermodynamics/vapor_saturation.jl
@@ -76,10 +76,11 @@ struct PlanarMixedPhaseSurface{FT}
     liquid_fraction :: FT
 end
 
-@inline specific_heat_difference(thermo, ::PlanarIceSurface) = specific_heat_difference(thermo, thermo.ice)
-@inline absolute_zero_latent_heat(thermo, ::PlanarIceSurface) = absolute_zero_latent_heat(thermo, thermo.ice)
-@inline absolute_zero_latent_heat(thermo, ::PlanarLiquidSurface) = absolute_zero_latent_heat(thermo, thermo.liquid)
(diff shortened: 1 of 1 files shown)


src/Thermodynamics/vapor_saturation.jl
```diff
@@ -1,46 +1,99 @@
...
+struct PlanarIceSurface end
+
+struct PlanarMixedPhaseSurface{FT}
+    liquid_fraction :: FT
+end
+
+@inline specific_heat_difference(thermo, ::PlanarIceSurface) = specific_heat_difference(thermo, thermo.ice)
+@inline absolute_zero_latent_heat(thermo, ::PlanarIceSurface) = absolute_zero_latent_heat(thermo, thermo.ice)
+@inline absolute_zero_latent_heat(thermo, ::PlanarLiquidSurface) = absolute_zero_latent_heat(thermo, thermo.liquid)
+@inline specific_heat_difference(thermo, ::PlanarLiquidSurface) = specific_heat_difference(thermo, thermo.liquid)
```
kaiyuan-cheng: Reorder them for a better look?

glwagner:
better tests in a future PR

PR #144, opened 2025-11-11
Update thermal bubble animation

navidcy:


giordano (approved):
I like the "we" style, more involving!

PR #148, opened 2025-11-11
Builds draft version of docs when the PR is a draft

giordano:
So we don't need to manually edit https://github.com/NumericalEarth/Breeze.jl/blob/c0ad90fb5af6c8900cba3548562e947ee6ccb01b/docs/src/index.md?plain=1#L29 anymore 🙂 Also, see https://github.com/NumericalEarth/Breeze.jl/issues/90#issuecomment-3506732252.

I'm happy to elaborate the conditions under which the full/draft documentation is built (for example if you want to use labels to control that), but this is a start, then we can iterate over as necessary.

Opening the PR as a draft to demonstrate the functionality.

### DIFF
--- a/docs/src/contributing.md
+++ b/docs/src/contributing.md
@@ -63,7 +63,7 @@ Pkg.test("Breeze"; test_args=`doctests`)
     Pkg.test("Breeze"; test_args=`--list`)
     ```
 
-### GPU tests
+### Tests on GPU
 
 When running the tests, if a CUDA GPU is detected, they automatically use the [`GPU` Oceananigans architecture](https://clima.github.io/OceananigansDocumentation/stable/appendix/library/#Oceananigans.Architectures.GPU), otherwise they run on [`CPU`](https://clima.github.io/OceananigansDocumentation/stable/appendix/library/#Oceananigans.Architectures.CPU).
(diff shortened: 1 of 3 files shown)


giordano:
We can definitely separate literate stuff vs `@example`s.  From within `@example` I don't think `Documenter` has the granularity to pick which examples to run and which to skip.

One possibility though is that we manually hack the ```` ```@example ```` annotations during the build to replace them with ```` ```julia ````, but we need to figure out how (sadly it looks like any sort of comment on the ```` ```@example ```` line messes up with the document generation).

navidcy:
You are a magician @giordano!

docs/src/contributing.md
```diff
@@ -117,3 +119,25 @@ Pkg.add("LiveServer") # this is necessary only the first time, to install LiveSe
 using LiveSever: serve
 serve(; dir="docs/build")
 ```
+
+### Draft documentation
+
+If you want to quickly build a draft copy of the documentation (i.e. without running the doctests or all the examples, which can be very time-consuming), set the environment variable `BREEZE_DOCS_DRAFT=true` and run again the `docs/make.jl` script:
+
+```sh
+export BREEZE_DOCS_DRAFT=true
```
glwagner:
```suggestion
export BREEZE_DRAFT_DOCS=true
```

my only excuse being that you did write "draft version of the docs", so IO would say these are "draft docs" :-D :-D

Issue #149, opened 2025-11-12
Thermal Bubble in the docs and in Readme look different


bischtob:
Didn't check the exact details, but was wondering if they should be the same :).

glwagner:
@navidcy very recently recreated the README, but I am wondering if maybe there was a little something extra in that script. I agree it should be the case that just cranking up resolution on the example should produce something different (which is what the video used to have)

giordano:
@navidcy on what machine did you run this? The examples in the docs are generated on x86_64-linux, I mentioned yesterday on slack that I see different results on aarch64-darwin and aarch64-linux (all three of them are different, but within each system I get consistent results). This is probably worth investigating.

navidcy:
I rerun, now I don't think they are way different: https://github.com/NumericalEarth/Breeze.jl/pull/151

glwagner:
> is there some aliasing in both sims?

do you mean spectral aliasing or another kind of aliasing? This code is finite volume, so there's no spectral aliasing issues... however we can get spurious oscillations for sure. The WENO advection scheme is designed to tame those, but doesn't always provide sufficient dissipation (esp for high order WENO, and strongly nonlinear dynamics that produce sharp gradients quickly)

PR #151, opened 2025-11-12
Update simulation stop time from 15 to 25 minutes

navidcy:
This makes the example identical to the one used for the README animation (except only the resolution 128 vs 512).

Related to #149.

### DIFF
--- a/docs/src/thermodynamics.md
+++ b/docs/src/thermodynamics.md
@@ -480,8 +480,8 @@ cˡ = thermo.liquid.heat_capacity
 Δcˡ = cᵖᵛ - cˡ
 ```
 
-This difference ``\Delta c^l ≈`` -2296 J/(kg⋅K) is negative because
-water vapor has a lower heat capacity than liquid water.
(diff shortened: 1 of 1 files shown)


glwagner:
It still doesn't look like the README https://numericalearth.github.io/BreezeDocumentation/previews/PR151/literated/thermal_bubble/

navidcy:
Let me run it again with 512 and see

navidcy:
@glwagner I updated the animation in the README in this PR.
Looks much more similar to docs, but only higher resolution!

glwagner:
its beautiful. any idea what was going on before?

glwagner (approved):

navidcy:
> any idea what was going on before?

perhaps I used an experimental branch yesterday by accident.... don't know

> also do you think gif or mp4 is better?

I like gif but it's usually much much much bigger file size and GitHub has a 10MB limit.

navidcy:
Does this PR close #149?

PR #153, opened 2025-11-12
Add DocStringExtensions to Project

glwagner:
Is this necessary? I was getting an error trying to run code from the repo environment without it:

```julia
Precompiling Breeze...
Info Given Breeze was explicitly requested, output will be shown live 
┌ Warning: You are using Julia v1.11 or later!"
│ Oceananigans is currently tested on Julia v1.10."
│ If you find issues with Julia v1.11 or later,"
│ please report at https://github.com/CliMA/Oceananigans.jl/issues/new
└ @ Oceananigans ~/.julia/packages/Oceananigans/A56vm/src/Oceananigans.jl:125
ERROR: LoadError: ArgumentError: Package Breeze does not have DocStringExtensions in its dependencies:
- You may have a partially installed environment. Try `Pkg.instantiate()`
  to ensure all packages in the environment are installed.
- Or, if you have Breeze checked out for development and have
  added DocStringExtensions as a dependency but haven't updated your primary
  environment's manifest file, try `Pkg.resolve()`.
- Otherwise you may need to report an issue with Breeze
```

I resolved and updated packages several times.


giordano:
We can have a look together at your issue, but I don't think this is necessary, Breeze dependencies should be resolved automatically.

PR #154, opened 2025-11-12
Generalized SaturationAdjustment to include a MixedPhaseEquilibrium

glwagner:
This PR fleshes out the design for a mixed phase saturation adjustment scheme, in which the an equilibrium state (wrt to condensate nucleation) is defined that produces both liquid and ice.

To do this, we implement two levels of abstraction. The low level abstraction is designed to represent the computation of a thermodynamically equilibrated saturation vapor pressure over planar surface which are either all liquid, all ice, or a mixture of ice and liquid. By "thermodyanmically equilibrated", we mean that we apply the Clausius-Claperyon relation to the _system_ rather than to the phases individually. This results in a particular form of the saturation vapor pressure.

One level up, this PR also implements an abstraction for the "equilibrium state" which `SaturationAdjustment` err... adjusts to. The equilibrium state could be a "warm phase equilibrium" or a "mixed phase equilibrium". When the equilibrium state is a mixed phase, the user has to parameterize how the liquid and ice phases are partitioned within the model equilibrium (note, no such equilibrium exists in reality, it is purely a model construction...)

We provide one such parameterization of a mixed phase equilibrium in which the fraction of supercooled liquid linearly varies from the freezing temperature to the temperature of homogenous ice nucleation.

### DIFF
--- a/src/Thermodynamics/dynamic_states.jl
+++ b/src/Thermodynamics/dynamic_states.jl
@@ -7,9 +7,6 @@ struct PotentialTemperatureState{FT} <: AbstractThermodynamicState{FT}
     reference_pressure :: FT
 end
 
-@inline PotentialTemperatureState(θ::FT, q::MoistureMassFractions{FT}, p₀::FT, pᵣ::FT) where FT =
-    PotentialTemperatureState{FT}(θ, q, p₀, pᵣ)
-
 @inline is_absolute_zero(𝒰::PotentialTemperatureState) = 𝒰.potential_temperature == 0
 
 @inline function exner_function(𝒰::PotentialTemperatureState, thermo::ThermodynamicConstants)


src/Microphysics/saturation_adjustment.jl
```diff
@@ -0,0 +1,278 @@
...
+    qᵛ⁺₁ = saturation_specific_humidity(T₁, ρ₁, thermo, equilibrium)
+    qᵗ <= qᵛ⁺₁ && return 𝒰₁
+
+    # Re-initialize first guess assuming saturation
+    𝒰₁ = adjust_state(𝒰₀, T₁, thermo, equilibrium)
+
+    # 𝒰₁ = with_moisture(𝒰₀, q₁)
+    # qᵛ⁺₁ = adjustment_saturation_specific_humidity(T₁, pᵣ, qᵗ, thermo, equilibrium)
+    # q₁ = equilibrated_moisture_mass_fractions(T₁, qᵗ, qᵛ⁺₁, equilibrium)
+    # 𝒰₁ = with_moisture(𝒰₀, q₁)
```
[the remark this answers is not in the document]
glwagner: yes I think so. I'm getting a gPU errror so was experimenting with ways to help the compiler infer things here (eg splitting into smaller functions). I made progress and I think this helped, so I will delete the comments.

src/Microphysics/saturation_adjustment.jl
```diff
@@ -0,0 +1,278 @@
...
+    materialize_microphysical_fields,
+    moisture_mass_fractions
+
+abstract type AbstractEquilibrium end
+
+"""
+    WarmPhaseSaturationAdjustment(reference_state, thermodynamics)
+
+Simple warm-phase saturation adjustment microphysics that computes temperature
+via a saturation adjustment.
```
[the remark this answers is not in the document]
glwagner: ill fix this

src/Microphysics/Microphysics.jl
```diff
@@ -1,168 +1,10 @@
...
-                                         moisture_mass_fraction)
-    @inbounds begin
-        qᵛ = microphysical_fields.specific_humidity[i, j, k]
-        qˡ = microphysical_fields.liquid_mass_fraction[i, j, k]
-    end
-    return MoistureMassFractions(qᵛ, qˡ, zero(qᵛ))
-end
-
-end # module Microphysics
+end # module Microphysics
```
giordano:
```suggestion
end # module Microphysics

```

.cursorrules
```diff
@@ -0,0 +1,229 @@
...
+    - explicit time-stepping
+    - acoustic substepping
+- A way to use total energy with the anelastic model
+- open boundaries following the Oceananigans.NonhydrostaticModel implementation
+- microphysics schemes from CloudMicrophysics
+- superdroplet microphysics schemes (reference tbd, but will use LagrangianParticleTracking)
+- MPI / distributed GPU support
+- Many, many more canonical LES validation cases
+- Terrain-following coordinate (may require upstream development in Oceananigans + using MutableVerticalCoordinate)
+- Cut-cells ImmersedBoundaryGrid implementation
```
giordano:
```suggestion
- Cut-cells ImmersedBoundaryGrid implementation
- 

```

docs/src/microphysics/mixed_phase_saturation_adjustment.md
```diff
@@ -0,0 +1,102 @@
...
+```math
+\lambda(T) = \frac{T' - T^f}{T^h - T^f} ,\qquad \lambda\in[0,1] .
+```
+
+A value of `λ = 1` corresponds to a pure liquid surface (warm side), and `λ = 0` corresponds to a pure
+ice surface (cold side). This model yields a mixed-phase surface with effective latent heat and heat-capacity
+difference that are linear blends of the pure-phase values, consistent with Pressel et al. (2015).
+
+In Breeze, the equilibrium surface is constructed internally via `PlanarMixedPhaseSurface(λ)` using
+`MixedPhaseEquilibrium`, and is accessed by
```
giordano:
```suggestion
[`MixedPhaseEquilibrium`](@ref Breeze.Microphysics.MixedPhaseEquilibrium), and is accessed by
```

glwagner:
I think we can ignore the error at the intersection of Float32 + GPU possibly and open an issue about it. We don't need Float32 immediately, but it would be nice to understand what's happening (and also support Float32 in the long run)

PR #163, opened 2025-11-13
Add tests for `set!`

glwagner:
I am expecting the GPU+Float32 to fail here, but not sure.

### DIFF
--- a/test/atmosphere_model_unit.jl
+++ b/test/atmosphere_model_unit.jl
@@ -22,19 +22,19 @@ end
             model = AtmosphereModel(grid; thermodynamic_constants=constants, formulation, microphysics)
             
             set!(model; qᵗ = 1e-2)
-            @test @allowscalar model.specific_moisture ≈ constant_field(grid, 1e-2)
+            @test model.specific_moisture ≈ constant_field(grid, 1e-2)
             
             ρᵣ = model.formulation.reference_state.density
-            @test @allowscalar model.moisture_density ≈ ρᵣ * 1e-2
+            @test model.moisture_density ≈ ρᵣ * 1e-2
 
             set!(model; u = 1, v = 2)
-            @test @allowscalar model.velocities.u ≈ constant_field(grid, 1)
-            @test @allowscalar model.velocities.v ≈ constant_field(grid, 2)
-            @test @allowscalar model.momentum.ρu ≈ ρᵣ
-            @test @allowscalar model.momentum.ρv ≈ ρᵣ * 2
+            @test model.velocities.u ≈ constant_field(grid, 1)
+            @test model.velocities.v ≈ constant_field(grid, 2)
+            @test model.momentum.ρu ≈ ρᵣ
+            @test model.momentum.ρv ≈ ρᵣ * 2
             
             ρᵣ = model.formulation.reference_state.density
-            @test @allowscalar model.moisture_density ≈ ρᵣ * 1e-2
+            @test model.moisture_density ≈ ρᵣ * 1e-2
             
             # test set! for a dry initial state
             ρᵣ = model.formulation.reference_state.density


glwagner:
We do have an abstraction called `ConstantField(c)` that might work here, I will try it.

glwagner:
We also need scalar indexing for ordinary approx right?

Or maybe I am confused. Do we hit the same codepath into GPUArrays with Field magically?

glwagner:
`ConstantField` does not work because it does not have a `grid` and therefore no notion of `size`. That is pretty easily changed though

glwagner:
> > ok so we also need to remove all the allowscalar!
> 
> How was it? Great minds...? [5a8233d](https://github.com/NumericalEarth/Breeze.jl/commit/5a8233d3a47158a159b7017de9a74e2137c47bd3)

😂

src/Microphysics/saturation_adjustment.jl
```diff
@@ -17,8 +20,6 @@ using DocStringExtensions: TYPEDSIGNATURES
 
 import ..Thermodynamics: saturation_specific_humidity
 
-abstract type AbstractEquilibrium end
-
```
giordano: I'm assuming that this wasn't meant to be removed: it disappeared in a [merge commit](https://github.com/NumericalEarth/Breeze.jl/commit/4e642d144cee2203aa9c0a5f2662d0e106b12c7d), which feels unlikely to be done on purpose?

src/AtmosphereModels/set_atmosphere_model.jl
```diff
@@ -130,8 +130,6 @@ end
         θ = potential_temperature[i, j, k]
     end
 
-    g = thermo.gravitational_acceleration
-    z = znode(i, j, k, grid, c, c, c)
```
giordano: These lines also got removed in another merged commit: 4e642d144cee2203aa9c0a5f2662d0e106b12c7d.  Was that done on purpose?

src/Microphysics/saturation_adjustment.jl
```diff
@@ -185,15 +197,21 @@ end
     return ϵᵈᵛ * (1 - qᵗ) * pᵛ⁺ / (pᵣ - pᵛ⁺)
 end
 
-@inline function adjust_state(𝒰₀, T, thermo, equilibrium)
+@inline function adjust_state(𝒰₀::AbstractThermodynamicState{FT}, T::FT,
+                              thermo::ThermodynamicConstants{FT},
+                              equilibrium::MixedPhaseEquilibrium{FT}) where FT
```
giordano: This stricter signature caused
```
MethodError: no method matching adjust_state(::Breeze.Thermodynamics.PotentialTemperatureState{Float64}, ::Float64, ::ThermodynamicConstants{Float64, CondensedPhase{Float64}, CondensedPhase{Float64}}, ::WarmPhaseEquilibrium)
The function `adjust_state` exists, but no method is defined for this combination of argument types.

Closest candidates are:
  adjust_state(::Breeze.Thermodynamics.AbstractThermodynamicState{FT}, ::FT, ::ThermodynamicConstants{FT}, !Matched::MixedPhaseEquilibrium{FT}) where FT
   @ Breeze ~/work/Breeze.jl/Breeze.jl/src/Microphysics/saturation_adjustment.jl:200
```
Do we need to define a different method for `WarmPhaseEquilibrium`?  Or the last argument can be `AbstractEquilibrium`?

giordano:
```patch
diff --git a/src/Microphysics/saturation_adjustment.jl b/src/Microphysics/saturation_adjustment.jl
index 54bd973..393258f 100644
--- a/src/Microphysics/saturation_adjustment.jl
+++ b/src/Microphysics/saturation_adjustment.jl
@@ -199,7 +199,7 @@ end
 
 @inline function adjust_state(𝒰₀::AbstractThermodynamicState{FT}, T::FT,
                               thermo::ThermodynamicConstants{FT},
-                              equilibrium::MixedPhaseEquilibrium{FT}) where FT
+                              equilibrium::AbstractEquilibrium) where FT
 
     pᵣ = 𝒰₀.reference_pressure
     qᵗ = total_specific_moisture(𝒰₀)
@@ -210,7 +210,7 @@ end
 
 @inline function saturation_adjustment_residual(T::FT, 𝒰₀::AbstractThermodynamicState{FT},
                                                 thermo::ThermodynamicConstants{FT},
-                                                equilibrium::MixedPhaseEquilibrium{FT}) where FT
+                                                equilibrium::AbstractEquilibrium) where FT
 
     𝒰₁ = adjust_state(𝒰₀, T, thermo, equilibrium)
     T₁ = temperature(𝒰₁, thermo)
```
would fix that issue and make all the `saturation_adjustment` tests pass, but I'm not sure this is the intended design, so I'm leaving that to you.

src/Microphysics/saturation_adjustment.jl
```diff
@@ -68,6 +73,7 @@ Return `WarmPhaseEquilibrium` representing an equilibrium between water vapor an
 """
 struct WarmPhaseEquilibrium <: AbstractEquilibrium end
 @inline equilibrated_surface(::WarmPhaseEquilibrium, T) = PlanarLiquidSurface()
+convert_equilibrium(FT, ::WarmPhaseEquilibrium) = WarmPhaseEquilibrium()
```
giordano: Is this method necessary?  I presume this is already covered by https://github.com/NumericalEarth/Breeze.jl/blob/aa22c7b13cbe1b6e0a58f88a02934dc99fa1ddce/src/Microphysics/saturation_adjustment.jl#L63?

src/Microphysics/saturation_adjustment.jl
giordano: I tentatively pushed 705c0b9888b2844a1efc0c4576bb6edfa3d1ff14 to implement the above, let me know if you didn't want it.

PR #165, opened 2025-11-13
Add/enhance some docstrings

navidcy:
The Clausius-Clapeyron derivation is included in two docstrings at the moment. I think we should just keep it in one, probably the `saturation_vapor_pressure` and in the `ThermodynamicConstants` just link to the other docstring? What do you think?


glwagner (approved):
<3

PR #181, opened 2025-11-20
Add compute_diffusivities to `update_state!` for `AtmosphereModel`

glwagner:
This does not work right now for all closures, because they make assumptions about the properties of `model`. One option is to update the properties of `AtmosphereModel` (this could be a quick fix). A better fix is to generalize how closures work in Oceananigans, which will require a new version.

### DIFF
--- a/ext/BreezeCloudMicrophysicsExt.jl
+++ b/ext/BreezeCloudMicrophysicsExt.jl
@@ -1,6 +1,6 @@
 module BreezeCloudMicrophysicsExt
 
-using CloudMicrophysics
+using CloudMicrophysics: CloudMicrophysics
 using CloudMicrophysics.Parameters: Parameters0M, Rain, Snow, CloudIce, CloudLiquid, CollisionEff
 using CloudMicrophysics.Microphysics0M: remove_precipitation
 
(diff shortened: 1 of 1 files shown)


glwagner:
Over on Oceananigans, https://github.com/CliMA/Oceananigans.jl/pull/4967 tackles the problem of defining a `buoyancy_force` interface. We can use this to support proper SmagorinskyLilly that includes a computation of the buoyancy gradient.

glwagner:
https://github.com/CliMA/Oceananigans.jl/blob/d3f0df86233828fb7c93b67591690c77102c3479/src/TurbulenceClosures/turbulence_closure_implementations/Smagorinskys/lilly_coefficient.jl#L137

to support `SmagorinskyLilly`, we need to implement `buoyancy_force(model)` and `∂z_b` so that

```julia
buoyancy = buoyancy_force(model)
∂z_b(i, j, k, grid, buoyancy, model.tracers) # returns the buoyancy gradient
```

this requires encapsulating some things within `buoyancy_force`.

glwagner:
I will merge this so that we have a temporary way to use `AnisotropicMinimumDissipation`

kaiyuan-cheng (approved):
Looks good.

PR #136, opened 2025-11-11
[CI] Remove JLD2 files before publishing docs

giordano:
These can be very heavy (40+ MB, the rest of the website is ~0.5 MB), and are useless for the purpose of the website.

PR #190, opened 2025-11-22
Add an example of cloudy Kelvin Helmholtz instability

glwagner:
In this case, microphysics plays a very strong role in modulating the evolution of the instability.

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,26 +56,23 @@ julia> using Pkg; Pkg.instantiate()
 
 Now we are ready to run any of the examples!
 
-For instance, if we run
+For instance, by increasing the resolution of the cloudy Kelvin-Helmholtz instability to `Nx=1536` and `Nz=1024` and running
 
 ```julia
-julia> include("examples/thermal_bubble.jl")
-```
-
-but after we tweak the spatial resolution of the grid to `size = (1024, 512)`, we get
-
-https://github.com/user-attachments/assets/aaca693c-57fe-46bf-8ff7-6646f6e5eebe
-
-By increasing the resolution of the cloudy Kelvin-Helmholtz instability to `Nx=1536` and `Nz=1024`, we run
-
-```julia
-julia> include("examples/thermal_bubble.jl")
+julia> include("examples/wave_clouds.jl")
 ```
 
 and get
 
-
 https://github.com/user-attachments/assets/010766c6-ed86-445f-87dc-e3b90c626589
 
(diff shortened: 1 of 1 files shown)


glwagner:
top: "u"
middle: vorticity
bottom: liquid mass fraction

image

navidcy:
I'm about to push some changes... But have a look at this:

https://github.com/user-attachments/assets/9c5e42be-77d0-4234-aeff-2856a1a8a099

docs/Project.toml
```diff
@@ -9,7 +9,7 @@ Oceananigans = "9e8cae18-63c1-5223-a75c-80ca9d6e9a09"
 Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
 
 [sources]
-Breeze = {path = ".."}
+Breeze = {path = "/Users/navid/Library/CloudStorage/OneDrive-TheUniversityofMelbourne/Documents/Research/Breeze.jl"}
```
giordano:
```suggestion
Breeze = {path = ".."}
```

src/Breeze.jl
```diff
@@ -14,6 +14,7 @@ export
     PotentialTemperature,
     PotentialTemperatureField,
     TemperatureField,
+    PotentialTemperatureField,
```
giordano: Duplicate?
```suggestion
```

docs/make.jl
```diff
@@ -62,3 +64,25 @@ makedocs(
...
+Return list of filepaths within `directory` that contains the `pattern::Regex`.
+"""
+function recursive_find(directory, pattern)
+    mapreduce(vcat, walkdir(directory)) do (root, dirs, filenames)
+        matched_filenames = filter(contains(pattern), filenames)
+        map(filename -> joinpath(root, filename), matched_filenames)
+    end
+end
+
+@info "Cleaning up temporary .jld2 and .nc output created by doctests or literated examples..."
```
giordano: Note that we're already cleaning up jld2 files before deploying the website (see #136), but I guess it's good to remove them after the build anyway if they aren't necessary (but perhaps they might be useful if someone wants to quickly do a new plot locally without rerunning the simulation?)

navidcy: Yes, the GH action removes the outputs from literate examples. But sometimes also doctests produce output (e.g. Oceananigans output writers doctests do that) so this was my way to cover all bases. If you have a cleaner suggestion it'd be great!

PR #192, opened 2025-11-22
[example] Internal wave in saturated conditions

glwagner:
This PR adds a new example that builds an internal wave in saturated conditions. It also adds some features that help compute the saturation specific humidity in various configurations:

* given the current state of the model
* under the assumptions of saturation adjustment, when the total specific humidity is known and vapor is assumed saturated
* under the condition that the total specific humidity is equal to the saturation specific humidity (ie rh=100%)

the addition of a utility for the last case should be useful for idealized problems I think.

### DIFF
--- a/examples/moist_internal_wave.jl
+++ b/examples/moist_internal_wave.jl
@@ -23,7 +23,7 @@ using Printf
 # ## Grid and AtmosphereModel
 
 Nx, Nz = 128, 128
-Lx, Lz = 30_000, 12_000 # 30 km
+Lx, Lz = 30_000, 12_000
 
 grid = RectilinearGrid(size = (Nx, Nz),
                        x = (-Lx/2, Lx/2),
@@ -48,7 +48,7 @@ cᵖᵈ = thermo.dry_air.heat_capacity
 ρ₀ = p₀ / (Rᵈ * θ₀)
 
 N = 0.012
-θᵇ(x, z) = θ₀ * exp(N² * z / g)
+θᵇ(x, z) = θ₀ * exp(N^2 * z / g)
 set!(model, θ=θᵇ)
 
 # Let's plot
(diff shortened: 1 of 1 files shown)


navidcy:
I'm trying to fix this example because the AI generation doesn't even run ... it's not that good tbh.

What do we want to demonstrate? How a small-amplitude wavepacket moves and what dispersion relationship would follow?

examples/moist_internal_wave.jl
```diff
@@ -0,0 +1,235 @@
...
+axislegend(ax_ts; position = :rt)
+
+ax_freq = Axis(fig[3, 1]; xlabel = "", ylabel = "ω (s⁻¹)",
+               title = "Dominant frequency comparison",
+               xticks = (1:3, ["observed", "dry", "saturated"]))
+barplot!(ax_freq, 1:3, [ω_observed, ω_dry, ω_moist];
+         color = (:gray50, :royalblue, :firebrick))
+
+fig
+=#
```
giordano: This is messing up with the rendering of the documentation: https://numericalearth.github.io/BreezeDocumentation/previews/PR192/literated/moist_internal_wave/###-Diagnostics:-observed-vs-predicted-buoyancy-frequency

navidcy: I'll sort it... it's a WIP atm

navidcy:
@glwagner I think we are committing over each other. I can't see the things I was doing yesterday...

I'll leave it to you?

giordano:
I don't think Greg has pushed in a while: https://github.com/NumericalEarth/Breeze.jl/activity?ref=glw/saturated-internal-wave 🤔

giordano:
Maybe?  There are [no force-pushes at all on this branch](https://github.com/NumericalEarth/Breeze.jl/activity?ref=glw/saturated-internal-wave&activity_type=force_push), and you're the only one who has pushed over the last couple of days 😅

glwagner:
this PR will have to be restarted if there's interest, too many conflicts. We did lift some of the saturation specific humidity utilities from here into other PRs though, so it was helpful

PR #200, opened 2025-11-24
Add inertia gravity wave validation case

kaiyuan-cheng:
This PR adds a new validation case that simulates the propagation of a gravity wave in an idealized setting, following [Skamarock and Klemp (1994)](https://journals.ametsoc.org/view/journals/mwre/122/11/1520-0493_1994_122_2623_eaaotk_2_0_co_2.xml).

The goal is to reproduce Fig. 1 from Skamarock and Klemp (1994), as shown below:
image

**Update on Dec 9, 2025**
We now got the correct wave number by changing the prognostic variable from moist static energy to potential energy.
image

### DIFF
--- a/examples/inertia_gravity_wave.jl
+++ b/examples/inertia_gravity_wave.jl
@@ -64,7 +64,7 @@ grid = RectilinearGrid(CPU(), size = (Nx, Nz), halo = (5, 5),
 # We use the anelastic formulation with liquid-ice potential temperature thermodynamics:
 
 constants = ThermodynamicConstants()
-reference_state = ReferenceState(grid, constants, base_pressure=p₀, potential_temperature=θ₀)
+reference_state = ReferenceState(grid, constants; surface_pressure=p₀, potential_temperature=θ₀)
 formulation = AnelasticFormulation(reference_state, thermodynamics=:LiquidIcePotentialTemperature)
 advection = WENO(minimum_buffer_upwind_order=3)
 model = AtmosphereModel(grid; formulation, advection)


examples/inertia_gravity_wave.jl
```diff
@@ -0,0 +1,114 @@
...
+                       z = (0, Lz),
+                       halo = (5, 5),
+                       topology = (Periodic, Flat, Bounded))
+
+
+# Boundary conditions
+free_slip_bcs = FieldBoundaryConditions(
+    top = FluxBoundaryCondition(nothing),
+    bottom = FluxBoundaryCondition(nothing)
+)
```
glwagner: these are the defaults, so you shouldn't have to set them

kaiyuan-cheng: Got it. I thought the default was no slip.

bischtob:
Just wondering if the default reference state that is set in the formulation that is set as default during constructions could interfere with the results?

examples/inertia_gravity_wave.jl
```diff
@@ -0,0 +1,108 @@
...
+g = model.thermodynamics.gravitational_acceleration
+θ̄ᵦ(z) = θ₀ * exp(N² * z / g)
+# Save initial potential temperature without perturbation to compute anomaly later
+θᵢ₀ = Field{Center, Nothing, Center}(grid)
+set!(θᵢ₀, (x, z) -> θ̄ᵦ(z) )
+
+# Perturbation
+function θᵢ(x, z)
+    θ′ = Δθ₀ * sin(π * z / Lz)  / (1 + ((x - x_c)/a)^2)
+    return θ̄ᵦ(z)+ θ′
```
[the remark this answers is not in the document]
glwagner: This function runs on the CPU anyways... so for this, we want to prioritize readability / interpretability and computational cost doesn't matter.

That said, `^2` is a "literal pow" which exapnds to

```julia
(x - x_c) / a * (x - x_c) / a
```

So I think that with pow on the inside, there are two divisions. Whereas the pow on the outside, there is only one division. And the division will dominate the cost I think.

glwagner:
@kaiyuan-cheng is there an analytical solution that we can use to plot contours on top of our heatmap?

As for wrong wavenumber for static energy... I guess that could point to a bug in the buoyancy computation? It could also be an initialization issue.

glwagner (approved):

examples/inertia_gravity_wave.jl
```diff
@@ -0,0 +1,181 @@
...
+grid = RectilinearGrid(CPU(), size = (Nx, Nz), halo = (5, 5),
+                       x = (0, Lx), z = (0, Lz),
+                       topology = (Periodic, Flat, Bounded))
+
+# ## Atmosphere model setup
+#
+# We use the anelastic formulation with liquid-ice potential temperature thermodynamics:
+
+constants = ThermodynamicConstants()
+reference_state = ReferenceState(grid, constants, base_pressure=p₀, potential_temperature=θ₀)
```
giordano:
```suggestion
reference_state = ReferenceState(grid, constants; surface_pressure=p₀, potential_temperature=θ₀)
```

PR #225, opened 2025-12-04
Add a supercell test case and fix inconsistencies in the hydrometeor variables.

kaiyuan-cheng:
This PR adds a new validation case that simulates an idealized supercell case proposed by [Klemp et al. (2015)](https://agupubs.onlinelibrary.wiley.com/doi/full/10.1002/2015MS000435).

**Update on Dec 15, 2025**
The supercell’s development looks reasonable, and we have a reasonable updraft speed when we add a turbulence closure. As pointed out by Klemp et al. (2015), the simulation of the supercell is sensitive to dissipation (diffusion).  

image


examples/supercell.jl
```diff
@@ -0,0 +1,151 @@
...
+                       z = (0, Lz),
+                       halo = (5, 5, 5),
+                       topology = (Periodic, Periodic, Bounded))
+
+
+# Problem parameters and initial conditions
+pᵣ, θᵣ = 100000, 300
+thermo = ThermodynamicConstants()
+reference_state = ReferenceState(grid, thermo, base_pressure=pᵣ, potential_temperature=θᵣ)
+formulation = AnelasticFormulation(reference_state)
```
glwagner: should we use `thermodynamics=:LiquidIcePotentialTemperature` ?

probably we should make this default too

kaiyuan-cheng: Sounds good to me!

examples/supercell.jl
```diff
@@ -0,0 +1,421 @@
...
+T_host = Array(parent(T))
+qᵛᵢ_host = similar(ph_host)
+
+# Compute initial water vapor from relative humidity and saturation mixing ratio:
+
+for k in axes(qᵛᵢ_host, 3), j in axes(qᵛᵢ_host, 2), i in axes(qᵛᵢ_host, 1)
+    z = znode(i, j, k, grid, Center(), Center(), Center())
+    T_eq = @inbounds T_host[i, j, k]
+    p_eq = @inbounds ph_host[i, j, k]
+    qᵛ_sat = 380 / p_eq * exp(17.27 * ((T_eq - 273) / (T_eq - 36)))
```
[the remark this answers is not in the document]
glwagner: touche 😂 

is this used in more than one place? we can add it to the source code if so.

If not in source, it should be made into a function with a nice name. This would allow it to be copy-pasted around which could be useful

kaiyuan-cheng:
> I think we should merge this and keep working on the supercell case in a new example! IF you agree @kaiyuan-cheng then feel free to merge.

Agreed. We should merge this first, and then revise/examine the code on top of it.

Issue #132, opened 2025-11-10
Package registration


glwagner:
I think we should register Breeze and then announce it to the community on slack + other places around the same time.

What do folks feel like is a good set of baseline features for this first announcement? I think mixed phase saturation adjustment, plus some basic validation (eg bomex, maybe GATE too) is a good start.

Or we can register now. Curious to hear thoughts.

kaiyuan-cheng:
I think we need at least one microphysics scheme and open boundary conditions so that the severe weather community can simulate an idealized supercell to study storm dynamics efficiently!

kaiyuan-cheng:
@glwagner The native implementation of the Kessler microphysics scheme is nearly complete. There are still a few points we need to discuss regarding the microphysics interface, and we should go over them tomorrow. Once that’s done, we can make an apples-to-apples comparison of the supercell case with the literature.

Another issue I want to highlight is that the simulated gravity waves still do not have the correct wavenumber. I’m hoping that using a better reference state may resolve this.

Let's aim for registering Breeze this week.

PR #228, opened 2025-12-06
Allow "energy forcing" of potential temperature + polish examples

glwagner:
This PR adds a forcing term to the potential temperature equation of the form:

$$ \partial_t \left ( \rho \theta \right ) = \frac{1}{c^{pm} \Pi} F_{\rho e} + \cdots $$

where $F_{\rho e}$ is the "energy forcing" specified via `forcing = (; ρe=ρe_forcing)` kwarg to `AtmosphereModel`. This is needed because energy flux divergences must be normalized by the Exner function $\Pi$ and mixture heat capacity $c^{pm}$ which are both state-dependent (so it would be complicated to build this in a user script).

I then used this to implement the radiation forcing for the Bomex case, also removing the explicit turbulence closure and reverting to `WENO(order=9)`. It looks decent I think:

Screenshot 2025-12-06 at 8 38 38 AM

I also cleaned up the cloud thermal bubble example.

### DIFF
--- a/examples/bomex.jl
+++ b/examples/bomex.jl
@@ -372,14 +372,14 @@ axq = Axis(fig[1, 2], xlabel="qᵛ (kg/kg)", ylabel="z (m)")
 axuv = Axis(fig[2, 1], xlabel="u, v (m/s)", ylabel="z (m)")
 axqˡ = Axis(fig[2, 2], xlabel="qˡ (kg/kg)", ylabel="z (m)")
 
-default_colours = Makie.wong_colors()
-colors = [default_colours[mod1(i, length(default_colours))] for i in 1:Nt]
-
 times = θt.times
 Nt = length(times)
 
+default_colours = Makie.wong_colors()
+colors = [default_colours[mod1(i, length(default_colours))] for i in 1:Nt]
+
 for n in 1:Nt
-    t_max = Int(times[n] / 60)
+    t_max = Int(times[n] / minute)
     label = n == 1 ? "initial condition" : "mean over $(t_max - 20)-$t_max min"
 
     lines!(axθ, θt[n], color=colors[n], label=label)


navidcy:
So atm there is a `bomex.jl` and an `anelastic_bomex.jl`; what's the difference?

navidcy:
When I run up to 12h, output every 20min and averaged the outputs from 9h-12h I get:

bomex_profiles_avg_9h-12h

Note the order of magnitude for $q_l$; agrees with Siebesma et al. (2003); figure 3.

navidcy:
Here's when I average the last 3hrs properly with

```Julia
schedule = AveragedTimeInterval(3hours; window=3hours, stride=5)
```

bomex_profiles_3hrs

navidcy:
The v-velocity seems a bit larger?

Screenshot 2025-12-07 at 10 17 32 am

navidcy:
Oh... btw I just noticed that the results above are for the Nx = Ny = 32!

glwagner:
Thanks for the analysis @navidcy ! I think we should add these plots to the example. I'd also like to include snapshots in addition to the average

glwagner:
I reduced the resolution to 32^2 out of fear that higher would take too long. We only have CPU resources for CI right now. Hopefully that will change.

navidcy:
> I reduced the resolution to 32^2 out of fear that higher would take too long. We only have CPU resources for CI right now. Hopefully that will change.

Yeah that's fine! I just missed that when I did the simulations to compare with the paper.

navidcy:
Here are the results with 64 x 64 x 40 and up to 6hours (averaged every 1h).

bomex_profiles_1hrs

glwagner:
I noticed that Opus 4.5 decided to stp using my energy forcing feature. that means the radiation forcing is a little too weak? which might explain too strong clouds. 

I will push a fix

examples/bomex.jl
```diff
@@ -0,0 +1,405 @@
...
+end
+
+add_callback!(simulation, progress, IterationInterval(100))
+
+outputs = merge(model.velocities, model.tracers, (; θ, qˡ, qᵛ))
+averaged_outputs = NamedTuple(name => Average(outputs[name], dims=(1, 2)) for name in keys(outputs))
+
+filename = "bomex.jld2"
+simulation.output_writers[:averages] = JLD2Writer(model, averaged_outputs; filename,
+                                                  schedule = AveragedTimeInterval(20minutes, window=20minutes, stride=1),
```
glwagner:
```suggestion
                                                  schedule = AveragedTimeInterval(20minutes),
```

will this still save the profile at t=0? I think its nice to have the initial condition

examples/bomex.jl
```diff
@@ -0,0 +1,405 @@
...
+
+θt = FieldTimeSeries(filename, "θ")
+qᵛt = FieldTimeSeries(filename, "qᵛ")
+qˡt = FieldTimeSeries(filename, "qˡ")
+ut = FieldTimeSeries(filename, "u")
+vt = FieldTimeSeries(filename, "v")
+
+for fts in (qᵛt, qˡt)
+    fts .*= 1000 # convert kg/kg -> g/kg
+end
```
glwagner: I prefer not to do this

examples/bomex.jl
```diff
@@ -0,0 +1,405 @@
...
+
+axθ = Axis(fig[1, 1], xlabel="θ (K)", ylabel="z (m)")
+axq = Axis(fig[1, 2], xlabel="qᵛ (g/kg)", ylabel="z (m)")
+axuv = Axis(fig[2, 1], xlabel="u, v (m/s)", ylabel="z (m)")
+axqˡ = Axis(fig[2, 2], xlabel="qˡ (g/kg)", ylabel="z (m)")
+
+for n in 1:Nt
+    t_max = Int(times[n] / 60)
+
+    label = n == 1 ? "t = $t_max min" : "mean $(t_max - 20)-$t_max min"
```
glwagner: I'm confused about this label, what is it trying to say?

examples/bomex.jl
glwagner: also might be good to isolate the variable `time_interval = 20minutes` since we can then change it to 60 easily

examples/bomex.jl
navidcy: [39172a1](https://github.com/NumericalEarth/Breeze.jl/pull/228/commits/39172a1e365fc83399258636907e248c8a8f86d0) ?

examples/bomex.jl
```diff
@@ -299,7 +299,7 @@ averaged_outputs = NamedTuple(name => Average(outputs[name], dims=(1, 2)) for na
 
 filename = "bomex.jld2"
 simulation.output_writers[:averages] = JLD2Writer(model, averaged_outputs; filename,
-                                                  schedule = TimeInterval(20minutes),
+                                                  schedule = AveragedTimeInterval(20minutes, window=20minutes, stride=1),
```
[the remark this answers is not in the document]
navidcy: ok!

examples/bomex.jl
glwagner: yes i like that better!

examples/bomex.jl
navidcy: No need to isolate it since now it only appears in one place?

glwagner:
image

navidcy (approved):

navidcy:
The `anelastic_bomex.jl` is needed? Delete?

navidcy:
Does it have more stuff in it? I'm happy to keep it, but update it to use the potential temperature formulation?

glwagner:
I think what matters here is the initial condition. So we can try to correct the initial condition by 

$$ \left ( \frac{p_0}{10^5} \right )^{R^m / c^{pm}} $$

or we can change our definition of the Exner function. This would add another parameter which we have to put somewhere.

navidcy:
So for the BOMEX case it would make more sense to use 101500 as reference pressure (not 1e5), right?

glwagner:
Well basically there are two different reference pressures. There is the pressure at $z=0$ which we use to form the refernce pressure profile. Then there is a second parameter which is used in the definition of hte Exner function.

glwagner:
I thought it made no sense to introduce a second parameter so I didn't. Probably it is useful though for analyzing observations rather than always having to tell people what your reference pressure is.

navidcy:
So a 0.4% difference in reference pressure has an impact on the results??

I might be under-appreciating how much the reference state matters for the atmosphere (or for the anelastic). I think I'm in a bit  of an ocean-centric thinking?

glwagner:
No I agree it seems preposterous!

glwagner:
Ok, so here is what I get when I adjust the initial condition:

image

glwagner:
image

this is what we get for the approximation

```julia
using Breeze.Thermodynamics: dry_air_gas_constant, vapor_gas_constant
Rᵈ = dry_air_gas_constant(constants)
cᵖᵈ = constants.dry_air.heat_capacity
p₀ = reference_state.base_pressure
χ(z) = (p₀ / 1e5)^(Rᵈ/  cᵖᵈ)
```

glwagner:
I'm worried it will perpetually annoy people that they have to subtract 1.2 deg to compare with other model results.

navidcy:
Also perhaps we run the full Bomex and include in the README before the package registration/announcement? Ref #132 ?

glwagner:
> Also perhaps we run the full Bomex and include in the README before the package registration/announcement? Ref #132 ?

that's a good idea. We should also make a visualization. It's in vogue to visualize the clouds in 3D. We can crank up the resolution.

giordano:
As far as I understand, [codecov is only pointing out](https://github.com/NumericalEarth/Breeze.jl/pull/228#issuecomment-3620614571) that the new lines added to `src/AtmosphereModels/anelastic_formulation.jl` (the`Base.show` method) aren't tested, and that's a large-ish fraction of all the changes to `src/` (the ones which affect coverage).  I'd say the `Base.show` method isn't super critical, so this is good to go for me.

glwagner:
> As far as I understand, [codecov is only pointing out](https://github.com/NumericalEarth/Breeze.jl/pull/228#issuecomment-3620614571) that the new lines added to `src/AtmosphereModels/anelastic_formulation.jl` (the`Base.show` method) aren't tested, and that's a large-ish fraction of all the changes to `src/` (the ones which affect coverage). I'd say the `Base.show` method isn't super critical, so this is good to go for me.

Okay, that makes sense. We can address that by adding a doctest perhaps in a future PR.

PR #230, opened 2025-12-07
[docs] Run Literate examples concurrently

giordano:
The idea is to make building the docs, and in particular running the Literate examples, a little bit faster by running them concurrently.  However I'm not 100% happy with this solution because

* it turns out [Literate isn't thread safe](https://github.com/fredrikekre/Literate.jl/issues/284), so we can't just run all the examples in the same Julia process
* I worked around the above limitation by running the examples in separate processes, which ensures isolation, however this means all the simulations have to be fully recompiled from scratch, while when running them serially we can take advantage of previously compiled functions in the same processes, so the time saving isn't as good as it could be if Literate itself was thread-safe
* because of having to to run Literate in separate processes, for some reason I don't quite understand Literate also decided to start embedding all SVG images as raw HTML in the markdown files, making the resulting HTML files too large to be rendered by Documenter (the filesize limit could be bumped, but the fact that the images were included as blobs in the HTML isn't great on its own), so I used `DisplayAs`, as suggested in [Literate documentation](https://fredrikekre.github.io/Literate.jl/v2/tips/#notebook-filesize)

As I mentioned, time saving isn't as good as it could be if Literate was thread-safe, but I verified this would shave off 5 minutes when building the docs for #228 (14 minutes instead of 19), which isn't too bad.


glwagner:
Oceananigans uses distributed processes for this, which could be more appropriate. Some of the examples will be heavy, and benefit from multithreading. So I think we expose more parallelism with processes rather than threads. I think this also might mesh better when we have hybrid GPU/CPU resources (eg some docs builds running on CPU while others are concurrently running on GPU). Is this valid, what's your take?

glwagner:
> > Oceananigans uses distributed processes for this, which could be more appropriate.
> 
> What do you mean exactly?

Oceananigans does its docs build with two processes `using Distributed`:

https://github.com/CliMA/Oceananigans.jl/blob/main/docs/make.jl

It builds / executes the Literate files with the two processes and then `rmprocs`: https://github.com/CliMA/Oceananigans.jl/blob/58c6b7a7b4760abc4da76ddef1ae8d198ea82b6b/docs/make.jl#L68

If I recall correctly, this essentially chopped the docs build in half. It is completely dominated by the example build time (Oceananigans has 12 examples)

> > I think this also might mesh better when we have hybrid GPU/CPU resources (eg some docs builds running on CPU while others are concurrently running on GPU). Is this valid, what's your take?
> 
> It's possible that GPU would help, but we need to verify it.

GPU may not help with the _existing_ examples.

But that's not what I mean. More importantly, it will allow us to implement many _more_ examples, which we would like to now, but cannot because we don't have the resources. Even the existing bomex case, we wish we could use the canonical resolution and run it for 6 hours. However due to limitations we chop the resolution by 4x and run it for only 1 hour (6x less than we want). Oceananigans is able to run simulations that have 16x more grid points than the current bomex, for a handful of cases. ClimaOcean runs many very intensive examples (100s of times more DoF). 

> > Some of the examples will be heavy, and benefit from multithreading.
> 
> I've tried using multiple Julia threads to build the docs before, but haven't seen any significant improvements. Of all the solutions I tried, the one presented here is the one which delivered the largest speedup.

I am guessing however that you only ran the existing cases. The test that has to be done is to increase the resolution / dimensionality (2D to 3D) and see if multithreading benefits those more expensive cases. We have intentionally been conservative implementing examples because we know that we don't have many resources. In other words we are already working around computational limitations.

glwagner:
Perhaps i should say, I have been running Breeze on GPU for research purposes, so I can anecdotally say that GPU yields a massive speed up in time-to-solution. It is something in the 20-100x range depending on problem size. We should benchmark this though.

giordano:
> Oceananigans does its docs build with two processes using Distributed:
> 
> CliMA/Oceananigans.jl@main/docs/make.jl
> 
> It builds / executes the Literate files with the two processes and then rmprocs: CliMA/Oceananigans.jl@58c6b7a/docs/make.jl#L68

Thanks for the link, that's pretty much equivalent to what I suggested here.  The orchestration of the processes is different (using `Distributed` vs `@spawn`-ing separate processes), but the end result (running the examples in separate processes) is pretty much the same.

giordano:
> Ah ok, so the code inside `Threads.@spawn` can itself run multithreaded? If so then this is a simpler solution!

`Threads.@spawn` is independent from Julia threading, and the way I implemented it here I could specify a different number of threads for the subprocesses running the Literate examples than the number of threads used for running the `make.jl` script since they're fully independent from each other.

examples/bomex.jl
```diff
@@ -19,6 +19,7 @@ using Oceananigans.Units
 using AtmosphericProfilesLibrary
 using Printf
 using CairoMakie
+using DisplayAs #hide
```
glwagner: what does this do?

giordano: > because of having to to run Literate in separate processes, for some reason I don't quite understand Literate also decided to start embedding all SVG images as raw HTML in the markdown files, making the resulting HTML files too large to be rendered by Documenter (the filesize limit could be bumped, but the fact that the images were included as blobs in the HTML isn't great on its own), so I used DisplayAs, as suggested in [Literate documentation](https://fredrikekre.github.io/Literate.jl/v2/tips/#notebook-filesize)

giordano:
I think this is still worthwhile, but I wanted to get other stuff in first.  Will propose this change again afterwards.

PR #235, opened 2025-12-08
[docs] Build literate examples on GPU

giordano:
Alternative/related to #230.  This will actually take a little bit longer, but allows us to run larger/longer simulations.

PR #238, opened 2025-12-08
[docs] Render CairoMakie figures as PNG

giordano:


giordano:
Yeah, so, it looks like Literate really likes to embed the PNGs as base64-encoded binary blobs in the raw HTML, we don't have separate output images anymore, see https://numericalearth.github.io/BreezeDocumentation/previews/PR238/literated/dry_thermal_bubble/ as an example.  I guess this may still an improvement in the sense that if we have higher resolution images (as in #235) hopefully the image size doesn't balloon, like if we have SVGs.  What do you think?

PR #241, opened 2025-12-09
Native implementation of DCMIP2016 Kessler microphysics

kaiyuan-cheng:
The original Fortran implementation of Kessler can be found [here](https://gitlab.in2p3.fr/ipsl/projets/dynamico/dynamico/-/blob/master/src/dcmip2016_kessler_physic.f90).

### DIFF
--- a/src/Microphysics/dcmip2016_kessler.jl
+++ b/src/Microphysics/dcmip2016_kessler.jl
@@ -1,12 +1,12 @@
 using ..Thermodynamics:
     MoistureMassFractions,
     MoistureMixingRatio,
-    mixture_heat_capacity,
+    PlanarLiquidSurface,
     mixture_gas_constant,
-    total_mixing_ratio,
-    total_specific_moisture,
+    mixture_heat_capacity,
     saturation_specific_humidity,
-    PlanarLiquidSurface
+    total_mixing_ratio,
+    total_specific_moisture
 
 using ..AtmosphereModels:
     dynamics_density,
@@ -14,17 +14,14 @@ using ..AtmosphereModels:
     surface_pressure
 
 using Oceananigans: CenterField, Field
+using Oceananigans.AbstractOperations: KernelFunctionOperation
 using Oceananigans.Architectures: architecture
-using Oceananigans.Grids: znode, Center
+using Oceananigans.Grids: Center, znode
 using Oceananigans.Utils: launch!
 
-using KernelAbstractions: @kernel, @index
-
-using Oceananigans.AbstractOperations: KernelFunctionOperation
-
 using Adapt: Adapt, adapt
-
 using DocStringExtensions: TYPEDSIGNATURES
+using KernelAbstractions: @index, @kernel
 
 """
     struct DCMIP2016KesslerMicrophysics


src/AtmosphereModels/update_atmosphere_model_state.jl
```diff
@@ -154,7 +159,7 @@ end
 
     update_microphysical_fields!(microphysical_fields, microphysics,
                                  i, j, k, grid,
-                                 ρ, 𝒰₁, constants)
+                                 ρ, 𝒰₁, p′, constants, Δt)
```
glwagner: can you explain why the perturbation pressure is needed and what it represents? Can it be computed from the density?

More generally we want to collapse the thermodynamic state into as few objects as possible, so I am wondering if this warrants a better design.

Also in what sens is this pressure a "perturbation"?

src/Microphysics/saturation_adjustment.jl
```diff
@@ -139,13 +139,13 @@ center_field_tuple(grid, names...) = NamedTuple{names}(CenterField(grid) for nam
 materialize_microphysical_fields(::WPSA, grid, bcs) = center_field_tuple(grid, :qᵛ, :qˡ)
 materialize_microphysical_fields(::MPSA, grid, bcs) = center_field_tuple(grid, :qᵛ, :qˡ, :qⁱ)
 
-@inline function update_microphysical_fields!(μ, ::WPSA, i, j, k, grid, ρ, 𝒰, constants)
+@inline function update_microphysical_fields!(μ, ::WPSA, i, j, k, grid, ρ, 𝒰, p′, constants, Δt)
```
glwagner: what would the perturbation pressure be for a fully compressible formulation?

src/AtmosphereModels/update_atmosphere_model_state.jl
```diff
@@ -15,11 +15,15 @@ const AnelasticModel = AtmosphereModel{<:AnelasticFormulation}
 
 function update_state!(model::AnelasticModel, callbacks=[]; compute_tendencies=true)
     tracer_density_to_specific!(model) # convert tracer density to specific tracer distribution
+
     fill_halo_regions!(prognostic_fields(model), model.clock, fields(model), async=true)
     compute_auxiliary_variables!(model)
     compute_forcings!(model)
```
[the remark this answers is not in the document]
kaiyuan-cheng: @glwagner I do think the order matters, and I think `microphysics_model_update!` should be called at the end of each cycle. Could you advise where it should go?

src/Microphysics/dcmip2016_kessler.jl
```diff
@@ -0,0 +1,627 @@
...
+                # The Fortran Kessler scheme updates θ (standard potential temperature) as:
+                #   θ_new = θ + ℒᵥ * (condensation - ern) / (cₚ * Π)
+                # where condensation and ern are in mixing ratio and represent PHASE CHANGES ONLY.
+                #
+                # For liquid-ice potential temperature θˡⁱ, the relationship is:
+                #   T = Π * θˡⁱ + ℒˡᵣ * qˡ / cₚ
+                #   θˡⁱ = (T - ℒˡᵣ * qˡ / cₚ) / Π
+                #
+                # The temperature change from latent heating (PHASE CHANGES ONLY) is:
+                #   ΔT = ℒᵥ * (condensation - ern) / cₚ
```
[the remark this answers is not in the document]
kaiyuan-cheng: @copilot open a new pull request to apply changes based on [this feedback](https://github.com/NumericalEarth/Breeze.jl/pull/241#discussion_r2616088075)

src/AtmosphereModels/update_atmosphere_model_state.jl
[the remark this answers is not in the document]
glwagner: you can think of `compute_tendencies!` as the begnining of the next time-step. Are you sure that you want to call the update after computing tendencies? I think that you probably want to call them before. Then, the tendencies are computed and will be used to advance fields to n+1.

It is a big confusing, but the reason the code is organized this way is so that communication and computation can be overlapped. The tendencies are the heaviest part of the cycle, and so if they are computed in another function (outside `update_state!`) we would not be able to leverage such a strategy

let me konw if that makes sense!

src/AtmosphereModels/microphysics_interface.jl
```diff
@@ -65,7 +65,7 @@ Return the microphysical velocities associated with `microphysics` and `name`.
 
 Must be either `nothing`, or a NamedTuple with three components `u, v, w`.
 """
-@inline microphysical_velocities(microphysics::Nothing, name) = nothing
+@inline microphysical_velocities(microphysics::Nothing, name, microphysical_fields) = nothing
```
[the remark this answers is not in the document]
glwagner: heh ok. then I guess it doesn't matter yet!

src/Microphysics/dcmip2016_kessler.jl
```diff
@@ -0,0 +1,627 @@
...
+                qᵛ_field[i, j, k]  = rᵛ_new
+                qᶜˡ_field[i, j, k] = rᶜ_final
+                qʳ_field[i, j, k]  = rʳ_final
+
+                #####
+                ##### Update liquid-ice potential temperature
+                #####
+                # The Fortran Kessler scheme updates θ (standard potential temperature) as:
+                #   θ_new = θ + ℒᵥ * (condensation - ern) / (cₚ * Π)
+                # where condensation and ern are in mixing ratio and represent PHASE CHANGES ONLY.
```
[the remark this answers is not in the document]
glwagner: it seems like it tried its best but is unable to write unicode 😂

glwagner:
oh no lots of conflicts 😱 @kaiyuan-cheng let me know if you need help

kaiyuan-cheng:
> oh no lots of conflicts 😱 @kaiyuan-cheng let me know if you need help

I was fixing the conflicts. Let me double-check if the code is still working correctly.

src/Microphysics/dcmip2016_kessler.jl
```diff
@@ -0,0 +1,628 @@
...
+using Oceananigans.Architectures: architecture
+using Oceananigans.Grids: znode, Center, Nothing as GridNothing
+using Oceananigans.Utils: launch!
+
+using KernelAbstractions: @kernel, @index
+
+using DocStringExtensions: TYPEDSIGNATURES
+
+"""
+    struct KesslerMicrophysics <: AbstractMicrophysics
```
giordano: The docstring doesn't match the implementation: this struct isn't actually a subtype of `AbstractMicrophysics`, was it supposed to?

src/Microphysics/dcmip2016_kessler.jl
```diff
@@ -0,0 +1,628 @@
...
+Return the names of prognostic microphysical fields for the Kessler scheme.
+
+# Fields
+- `:ρqᶜˡ`: Density-weighted cloud liquid mass fraction (\$kg/m^3\$).
+- `:ρqʳ`: Density-weighted rain mass fraction (\$kg/m^3\$).
+"""
+prognostic_field_names(::KM) = (:ρqᶜˡ, :ρqʳ)
+
+"""
+    materialize_microphysical_fields(::KesslerMicrophysics, grid, boundary_conditions)
```
giordano:
```suggestion
$(TYPEDSIGNATURES)
```

src/Microphysics/dcmip2016_kessler.jl
```diff
@@ -0,0 +1,628 @@
...
+        qᶜˡ = μ.ρqᶜˡ[i, j, k] / ρ
+        qʳ  = μ.ρqʳ[i, j, k] / ρ
+    end
+    qˡ = qᶜˡ + qʳ
+    qᵛ = qᵗ - qˡ
+    return MoistureMassFractions(qᵛ, qˡ)
+end
+
+"""
+    maybe_adjust_thermodynamic_state(𝒰, ::KesslerMicrophysics, μ, qᵗ, constants)
```
giordano:
```suggestion
$(TYPEDSIGNATURES)
```

src/Microphysics/dcmip2016_kessler.jl
```diff
@@ -0,0 +1,628 @@
...
+
+Convert mass fraction \$q\$ to mixing ratio \$r\$.
+\$q^t\$ is the total mass fraction (sum of all moisture species).
+
+The conversion is: \$r = q / (1 - q^t)\$
+"""
+@inline mass_fraction_to_mixing_ratio(q, qᵗ) = q / (1 - qᵗ)
+
+"""
+    mixing_ratio_to_mass_fraction(r, rᵗ)
```
giordano:
```suggestion
$(TYPEDSIGNATURES)
```

src/Microphysics/dcmip2016_kessler.jl
```diff
@@ -0,0 +1,628 @@
...
+        end
+    end
+end
+
+#####
+##### Interface stub for update_microphysical_fields!
+#####
+
+"""
+    update_microphysical_fields!(μ, ::KesslerMicrophysics, i, j, k, grid, ρ, 𝒰, constants)
```
giordano:
```suggestion
$(TYPEDSIGNATURES)
```

src/AtmosphereModels/microphysics_interface.jl
giordano: I believe the main problem which caused all tests to fail is that you changed the signature of this function, without changing its uses, for example https://github.com/NumericalEarth/Breeze.jl/blob/07e2ec9e953b82540382cd6e0b05c1358b416b96/src/AtmosphereModels/dynamics_kernel_functions.jl#L139 but there are more.  Also, I don't understand why this was changed in the first place, in the rest of this PR it doesn't look like you make use of this 3-argument signature anywhere.

src/Microphysics/Microphysics.jl
```diff
@@ -20,10 +21,11 @@ import ..AtmosphereModels:
     materialize_microphysical_fields,
     microphysical_velocities,
     compute_moisture_fractions,
-    microphysical_tendency
+    microphysical_tendency,
+    microphysics_model_update!
 
 include("saturation_adjustment.jl")
 include("bulk_microphysics.jl")
-include("microphysics_diagnostics.jl")
```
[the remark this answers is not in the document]
giordano: Bringing back the inclusion of `microphysics_diagnostics.jl` would resolve the test failures in the quality assurance tests, but if this file is considered obsolete it should be deleted and the stale imports should be removed.

kaiyuan-cheng:
Really appreciate the hard work @giordano put into this PR. I apologize for not making sure the unit test worked. I was focused on the scheme itself.

test/dcmip2016_kessler.jl
```diff
@@ -90,164 +92,151 @@ function kessler_fortran(theta, qv, qc, qr, rho, pk, dt, z, nz)
...
-    pk_f = (p ./ 100000.0).^0.2875 # Match Kessler xk
-    
-    # Initial conditions (Mass Fractions for Breeze)
-    # Convert mixing ratios to mass fractions
+
+    # Initial conditions (mixing ratios for reference implementation)
+    qv_ref = 0.01 * ones(Nz)
+    qc_ref = 0.001 * ones(Nz)
+    qr_ref = 0.0005 * ones(Nz)
+    θ_ref = 300.0 * ones(Nz)
```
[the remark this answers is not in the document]
glwagner: notation is incorrect as well

src/Microphysics/dcmip2016_kessler.jl
```diff
@@ -0,0 +1,627 @@
...
+    kessler_terminal_velocity(rʳ, ρ, ρ_bottom)
+
+Compute liquid water terminal velocity (\$m/s\$) following Klemp and Wilhelmson (1978) eq. 2.15.
+
+Uses the three-argument form with explicit reference density.
+`ρ_bottom` is the reference density at the lowest vertical level (\$kg/m^3\$).
+"""
+@inline function kessler_terminal_velocity(rʳ, ρ, ρ_bottom)
+    rhalf = sqrt(ρ_bottom / ρ)
+    return 36.34 * (rʳ * 0.001 * ρ)^0.1364 * rhalf
```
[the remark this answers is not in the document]
glwagner: I agree with copilot here --- add these constants to `DCMIP2016KesslerMicrophysics`, and then invoke them here. I see two constants --- also why are units converted? Can we ensure that `r` has correct units?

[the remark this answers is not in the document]
glwagner: or perhaps 0.001 is also a parameter?

src/Microphysics/dcmip2016_kessler.jl
```diff
@@ -54,8 +57,18 @@ instead, it is diagnosed from the total specific moisture `qᵗ` and the liquid
...
+- `f2x :: FT = 17.27`: Clausius-Clapeyron exponent coefficient (empirical constant for saturation)
+- `p₀ :: FT = 100000.0`: Reference sea level pressure in Pascals
+
+The saturation adjustment coefficient `f5 = 237.3 × f2x × ℒˡᵣ / cᵖᵈ` is computed dynamically
+from `f2x` and the thermodynamic constants.
 """
-struct DCMIP2016KesslerMicrophysics end
+Base.@kwdef struct DCMIP2016KesslerMicrophysics{FT}
+    f2x :: FT = 17.27
+    p₀  :: FT = 100000.0
```
glwagner: This parameter should be available as `standard_pressure(dynamics)` -- should we pass `dynamics` into the microphysics tendency functions? We can also move this parameter to `ThermodynamicsConstants` (but I felt that it did not belong there, since it is really associated with the definition of potential temperature rather than a thermodynamic property)

src/Microphysics/Microphysics.jl
```diff
@@ -13,14 +13,20 @@ export
...
     RelativeHumidityField
 
-using ..AtmosphereModels: AtmosphereModels, compute_moisture_fractions,
-    materialize_microphysical_fields, update_microphysical_fields!
+using ..AtmosphereModels: AtmosphereModels
+
+import ..AtmosphereModels:
+    compute_moisture_fractions,
+    materialize_microphysical_fields,
+    update_microphysical_fields!
```
giordano: In the spirit of getting rid of all `import`s (it's a sort of spooky action at a distance, see #380 and #296), this change can be reverted by qualifying the method definitions with the `AtmosphereModels` module (see the above suggestions).

src/Microphysics/dcmip2016_kessler.jl
```diff
@@ -0,0 +1,769 @@
...
+"""
+function AtmosphereModels.microphysics_model_update!(microphysics::DCMIP2016KM, model)
+    grid = model.grid
+    arch = architecture(grid)
+    Nz = grid.Nz
+    Δt = model.clock.last_Δt
+
+    # Skip microphysics update if timestep is zero, infinite, or invalid
+    # (e.g., during model construction before any time step has been taken)
+    (isnan(Δt) || isinf(Δt) || Δt ≤ 0) && return nothing
```
giordano: Negating [`isfinite`](https://docs.julialang.org/en/v1/base/numbers/#Base.isfinite) covers both `isnan` and `isinf`
```suggestion
    (!isfinite(Δt) || Δt ≤ 0) && return nothing
```

test/dcmip2016_kessler.jl
```diff
@@ -0,0 +1,413 @@
...
+    # Use Float64 for accurate comparison
+    FT = Float64
+    
+    # 1. Setup shared column data
+    nz = 40
+    z_min = FT(0)
+    z_max = FT(4000)
+    
+    # Create grid
+    grid = RectilinearGrid(CPU(), size=(1, 1, nz), x=(0, 100), y=(0, 100), z=(z_min, z_max), topology=(Periodic, Periodic, Bounded))
```
giordano: Any specific reason for forcing the CPU architecture and not use
```suggestion
    grid = RectilinearGrid(default_arch, size=(1, 1, nz), x=(0, 100), y=(0, 100), z=(z_min, z_max), topology=(Periodic, Periodic, Bounded))
```
(`default_arch` is `GPU()` when an Nvidia GPU is available) instead?

src/Microphysics/dcmip2016_kessler.jl
```diff
@@ -0,0 +1,769 @@
...
+            qᵗ = ρqᵗ[i, j, k] * inv_ρ
+            qᶜˡ = max(ρqᶜˡ[i, j, k] * inv_ρ, zero(FT))
+            qʳ  = max(ρqʳ[i, j, k] * inv_ρ, zero(FT))
+            qˡ_sum = qᶜˡ + qʳ
+            qᵗ = max(qᵗ, qˡ_sum)  # Prevent negative vapor
+            qᵛ = qᵗ - qˡ_sum       # Diagnose vapor
+
+            # Convert to mixing ratios for Kessler physics
+            # Precompute common denominator: 1/(1 - qᵗ)
+            inv_one_minus_qᵗ = inv(one_FT - qᵗ)
```
glwagner: this optimization is probably unnecesssary in this context

test/dcmip2016_kessler.jl
```diff
@@ -0,0 +1,413 @@
...
+- `rho`: dry air density (kg/m³)
+- `pk`: Exner function (p/p0)^(R/cp)
+- `dt`: time step (s)
+- `z`: heights of thermodynamic levels (m)
+- `constants`: ThermodynamicConstants
+
+Returns:
+- `precl`: precipitation rate (m_water/s)
+"""
+function fortran_kessler!(theta, qv, qc, qr, rho, pk, dt, z, constants)
```
glwagner: the notation in this function should be changed to match Breeze. I understand that it has been formulated to match a fortran implementation (which is not in this repo), but for future maintainers this can only cause confusion.

src/Microphysics/dcmip2016_kessler.jl
```diff
@@ -389,15 +388,21 @@ end
...
     FT = eltype(grid)
     one_FT = one(FT)
     surface = PlanarLiquidSurface()
 
+    # Extract microphysical fields from μ
+    ρqᶜˡ = μ.ρqᶜˡ
+    ρqʳ = μ.ρqʳ
+    qᵛ_field = μ.qᵛ
+    qᶜˡ_field = μ.qᶜˡ
+    qʳ_field = μ.qʳ
```
glwagner:
```suggestion
    ρqᶜˡ = μ.ρqᶜˡ[i, j, k]
    ρqʳ = μ.ρqʳ[i, j, k]
    qᵛ_ = μ.qᵛ[i, j, k]
    qᶜˡ = μ.qᶜˡ[i, j, k]
    qʳ = μ.qʳ[i, j, k]
```

an idea to keep the notation compact

AGENTS.md
```diff
@@ -37,6 +37,8 @@ Breeze interfaces with ClimaOcean for coupled atmosphere-ocean simulations.
    - Mark functions inside kernels with `@inline`.
    - **Never use loops outside kernels**: Always replace `for` loops that iterate over grid points
      with kernels launched via `launch!`. This ensures code works on both CPU and GPU.
+   - **Use literal zeros**: Write `max(0, a)` instead of `max(zero(FT), a)`. Julia handles type
+     promotion automatically, and `0` is more readable. The same applies to `min`, `clamp`, etc.
```
giordano: I'm actually not sure about this suggestion, `max(0, a)` would be type-unstable. And the claim about automatic type promotion is technically incorrect (although there's _some_ automatism): https://docs.julialang.org/en/v1/manual/conversion-and-promotion/#conversion-and-promotion. The convenience of this is context-dependent. @glwagner @navidcy what do you think?

PR #242, opened 2025-12-09
Import prettysummary from Oceananigans.Utils

navidcy:
Atm tests break on main


giordano (commented):
Yay! Note that this should only break the tests in ExplicitImports.jl, but the package should work fine.

giordano (approved):
I meant to approve!

PR #244, opened 2025-12-09
[docs] Including info about computing environment in Literate examples

giordano:
Today I got a complaint from @matt-graham that in Literate examples in Oceananigans it's hard to figure out the versions of all the various packages used in the examples, which makes reproducing them hard.  I'm first addressing this complaint here in Breeze because it's a simpler playground for experimenting with a solution, and then we can do the same in Oceananigans as a follow up.

The approach I suggest here is to add a common postamble to all the literate examples which shows the Julia version and the top-level packages installed in the environment, in addition to including the `Project.toml` and `Manifest.toml` files so that people can actually fully reproduce the same computer environment (at least on the Julia software side, if the example uses a GPU...unfortunately we can't easily give one to you!).  PR best reviewed by [disabling whitespace changes](https://github.com/NumericalEarth/Breeze.jl/pull/244/files?w=1).

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -44,6 +44,9 @@ Breeze interfaces with ClimaOcean for coupled atmosphere-ocean simulations.
    - If an implementation is awkward, don't hesitate to suggest an upstream feature (eg in Oceananigans)
      that will make something easier, rather than forcing in low quality code.
 
+6. **Extending functions**
+   - Almost always extend functions in source code, not in examples
+
 ### Oceananigans ecosystem best practices
 
 1. **General coding style**
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -39,7 +39,7 @@ example_scripts = [
     "cloudy_thermal_bubble.jl",
     "cloudy_kelvin_helmholtz.jl",
     "bomex.jl",
-    # "prescribed_sst.jl", # this is a WIP
+    "prescribed_sst.jl",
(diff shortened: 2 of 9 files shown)


glwagner:
Interesting! It's a nice idea! @matt-graham should open issues on Oceananigans about it. I am guessing this is not the only complaint 😂

giordano:
I see. Locally I had a manifest in the docs directory, but may have been an old one.

Ok, I'm getting convinced that shipping all the environment files may not be a practical solution, because they rely on the specific layout of the package directory: that's useful if one wants to reproduce the full docs build inside this repository, but not for a standalone example.

Showing top-level packages should at least address the high level concern Matt had about not knowing the versions of the direct dependencies.

giordano:
@glwagner see the bottom of https://numericalearth.github.io/BreezeDocumentation/previews/PR244/literated/dry_thermal_bubble/ as an example

navidcy (approved):
I like this!!
I've been thinking about something like this for long now..

matt-graham:
Thanks @giordano for being so quick at looking in to this 😲! 

> Interesting! It's a nice idea! @matt-graham should open issues on Oceananigans about it. I am guessing this is not the only complaint 😂

It was more of a casual comment at the pub to @giordano than a complaint 😅, in general my experience of using Oceananigans so far has been great!

PR #248, opened 2025-12-10
Create abstractions for bulk flux boundary conditions

glwagner:
I am anticipating that idealized bulk boundary conditions will be useful and adding abstractions for them.

The plan is still to use ClimaOcean (whatever it becomes) to do bona fide Monin-Obhukov theory. There's a lot of scope for idealized bulk boundary conditions in my opinion, which can be computed directly without requiring a solver.

I'll implement these in the prescribed SST example.

### DIFF
--- a/test/forcing_and_boundary_conditions.jl
+++ b/test/forcing_and_boundary_conditions.jl
@@ -71,9 +71,9 @@ end
 
 @testset "Bulk boundary conditions [$FT]" for FT in (Float32, Float64)
     grid = RectilinearGrid(default_arch, FT; size=(4, 4, 4), x=(0, 100), y=(0, 100), z=(0, 100))
-    const Cᴰ = 1e-3
-    const gustiness = 0.1
-    const T₀ = 290
+    Cᴰ = 1e-3
+    gustiness = 0.1
+    T₀ = 290
 
     @testset "BulkDrag construction and application [$FT]" begin
         # Test construction with default parameters


src/BoundaryConditions/BoundaryConditions.jl
```diff
@@ -0,0 +1,571 @@
...
+    surface = PlanarLiquidSurface()
+    qᵛ₀ = Field{Center, Center, Nothing}(grid)
+    
+    # Create a function that computes qᵛ₀ pointwise
+    qᵛ₀_func(i, j) = saturation_specific_humidity(T₀[i, j, 1], ρ₀, constants, surface)
+    
+    # Set the field using the interior indices
+    for i in 1:size(grid, 1), j in 1:size(grid, 2)
+        qᵛ₀[i, j, 1] = qᵛ₀_func(i, j)
+    end
```
giordano: GPU doesn't like the pointwise computation

glwagner: sorry i should be nice

glwagner: this will probably require a kernel

glwagner:
@navidcy sorry if I wrote over a lot of your changes, i had a refactor in the works. I am stepping out now so commit away!

navidcy:
Don't worry!

PR #249, opened 2025-12-10
[docs] Run Literate examples concurrently

giordano:
Retry of #230


navidcy (approved):

giordano:
Now that we run all the examples in isolation we could even remove https://github.com/NumericalEarth/Breeze.jl/blob/805fc3f3149259478a2a5e9710f89218b8177249/examples/bomex.jl#L424-L426 if that's considered too noisy

glwagner:
MM yeah I thikn that would be best to remove!

glwagner (approved):

PR #259, opened 2025-12-11
Add proper boundary condition to reference pressure and density

### LINKED ISSUE #258: Implement correct surface/bottom boundary condition for the reference pressure
Right now we use the default boundary conditions for reference pressure:

https://github.com/NumericalEarth/Breeze.jl/blob/1393b3b1f6686319ec10a8a64d49a5268d8a6a55/src/Thermodynamics/reference_states.jl#L100

the default is zero gradient, so it means that the reference pressure evaluated at the first face (corresponding to z=0) will return the same value as the first cell.

This does not represent the physical reality, where the reference pressure evaluated at the first face should equal the parameter provided for `surface_pressure`. We can implement this by using a `ValueBoundaryCondition(surface_pressure)`.

This could be useful at some points because we can correctly extract the surface pressure given just the reference pressure profile, without requiring the additional surface pressure parameter.

the same principle holds for the reference density

https://github.com/NumericalEarth/Breeze.jl/blob/1393b3b1f6686319ec10a8a64d49a5268d8a6a55/src/Thermodynamics/reference_states.jl#L96

put another way, I think we want

```julia
ℑzᵃᵃᶠ(i, j, 1, grid, pᵣ) == p₀
```

to be true.

glwagner:
Closes #258


navidcy (approved):
Does this imply that we don't need the [correction factor χ](https://numericalearth.github.io/BreezeDocumentation/dev/literated/bomex/#Profiles-from-AtmosphericProfilesLibrary) anymore (see, e.g., bomex example)? Should we modify the example then to drop it?

glwagner:
> Does this imply that we don't need the [correction factor χ](https://numericalearth.github.io/BreezeDocumentation/dev/literated/bomex/#Profiles-from-AtmosphericProfilesLibrary) anymore (see, e.g., bomex example)? Should we modify the example then to drop it?

no that will require a separate change. This change makes is so that we don't need to have if-statements to evaluate the surface pressure vs pressure at other levels, which is useful for the radiation code.

PR #260, opened 2025-12-12
[docs] Sprinkle some `const`s in examples

giordano:
I'm curious to see how much of an impact this has on docs building runtime.  Locally I see a decent benefit when executing this example (~20%, excluding the plotting), perhaps on longer-running simulation this will be less important, but it's an experiment.


giordano:
It did help locally for me: excluding all the plotting stuff, the first simulation run went from 45 to 36 seconds (second time takes ~1 second 😂), these timings are fully consistent and reproducible several times, but clearly on the CI system most of the compilation effort isn't affected that much by adding all these annotations, so at this point we can keep the code simpler.

PR #266, opened 2025-12-14
Improve the RICO example and enforce impenetrability on subsidence velocity

glwagner:
@kaiyuan-cheng I wonder if this was a source of trouble. I cleaned up the explanations in RICO too.

### DIFF
--- a/examples/rico.jl
+++ b/examples/rico.jl
@@ -221,7 +221,7 @@ set!(model, θ=θᵢ, qᵗ=qᵢ, u=uᵢ, v=vᵢ)
 # RICO typically requires longer integration times than BOMEX to develop
 # a quasi-steady precipitating state.
 
-simulation = Simulation(model; Δt=10, stop_time=12hours)
+simulation = Simulation(model; Δt=10, stop_time=6hour)
 conjure_time_step_wizard!(simulation, cfl=0.7)
 
 # ## Output and progress
@@ -295,7 +295,7 @@ slice_outputs = (
 
 filename = "rico_slices.jld2"
 simulation.output_writers[:slices] = JLD2Writer(model, slice_outputs; filename,
-                                                schedule = TimeInterval(1minute),
+                                                schedule = TimeInterval(30seconds),
                                                 overwrite_existing = true)
 
 # We're finally ready to run this thing,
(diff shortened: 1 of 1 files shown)


examples/rico.jl
```diff
@@ -70,43 +67,45 @@ formulation = AnelasticFormulation(reference_state,
...
-# ## Surface momentum flux (drag)
-
-Cᴰ = 1.229e-3
 ρu_bcs = FieldBoundaryConditions(bottom=BulkDrag(coefficient=Cᴰ))
 ρv_bcs = FieldBoundaryConditions(bottom=BulkDrag(coefficient=Cᴰ))
 
+# Within the canon of Monin-Obukhov similarity theory, these transfer
+# coefficients should be scaled if the vertical grid spacing is changed.
+# Here we can use the values from [vanZanten2011](@citet) verbatim because
+# we use the recommend vertical grid spacing of 40 m.
```
navidcy:
```suggestion
# we use the recommended vertical grid spacing of 40 m.
```

navidcy:
Before the last plot there is this:

> Convert precipitation rate to mm/day for more intuitive units P is in kg/kg/s, multiply by ρ~1 kg/m³ and 86400 s/day and 1000 mm/m gives ~86.4 factor But since P is specific (kg/kg/s), we'll just show it in 10⁻⁶ s⁻¹ for clarity

Seems wrong when I look at the figure. Is this left over from some other version? Delete?

navidcy:
> Before the last plot there is this:
> 
> > Convert precipitation rate to mm/day for more intuitive units P is in kg/kg/s, multiply by ρ~1 kg/m³ and 86400 s/day and 1000 mm/m gives ~86.4 factor But since P is specific (kg/kg/s), we'll just show it in 10⁻⁶ s⁻¹ for clarity
> 
> Seems wrong when I look at the figure. Is this left over from some other version? Delete?

deleted by https://github.com/NumericalEarth/Breeze.jl/pull/266/commits/86c578ba0cc44d5b50ab65dce97dd78b3867b54a; feel free to bring back if it was crucial.

glwagner:
I think the movie has the right framerate to be interpretable now, but the build is pretty expensive (is it like this on `main`, or is it the movie making fault?) If the movie making is slow maybe we can think of ways to speed it up while preserving the framerate... we could run a shorter simulation, perhaps.

navidcy (approved):

kaiyuan-cheng:
My test suggests that this fix doesn’t change the result, at least based on a visual inspection.

RICO:
image

RICO2:
image

glwagner:
> I thought you wanted to shorten the simulation because the example was taking too long to run.

The total time for the example includes 1) the time to run the simulation and 2) the time it takes to generate the movie. When 2) dominates, we can reduce the total time by making a shorter movie.

glwagner:
> My test suggests that this fix doesn’t change the result, at least based on a visual inspection.

Ok, you're right!

This helped me realize that enforcing impenetrability does nothing for fields that have zero gradient on the boundary. Because w * dz(phi) is already zero.

Also, this change probably would only affect the cells adjacent to the top boundary, not the profile throughout the domain.

So we should remove subsidence on potential temperature in this PR as well?

glwagner:
Oh I see, the LS forcing includes both radiation and horizontal advection for potential temperature, and corresponds to drying for q^v.

PR #268, opened 2025-12-14
Remove duplicate import of Oceananigans

navidcy:


giordano:
We should run `ExplicitImports` also for the examples 😈 (sadly I'm only joking, because `ExplicitImports` requires a module defined in a file on disk, we can't even simply define a module in memory 😢)

PR #279, opened 2025-12-15
[CI] Use self-hosted runners for GPU jobs

giordano:


glwagner:
docs build takes exactly the same amount of time!

PR #284, opened 2025-12-15
Implement NonEquilibriumCloudFormation for 1M microphysics

### LINKED ISSUE #275: Use 1M microphysics for RICO case
It would be nice to show and involves only warm rain processes. 0M is just so simplified.

@kaiyuan-cheng do we want to merge changes in https://github.com/NumericalEarth/Breeze.jl/pull/225 before tackling this?

glwagner:
Closes #275.

This PR refactors `BulkMicrophysics` to have a `cloud_formation` property (replacing `nucleation`). This better reflects the fact that saturation adjustment, or its alternatives, primarily model condensation and deposition onto existing particles rather than nucleation itself.

We then go on to provide `NonEquilibriumCloudFormation` which can describe non-adjustment processes.

Finally, we build out the CloudMicrophysics extension to use this. It changes the default for 1M microphysics from saturation adjustment to the nonequilibrium case. 

We also refactor of `maybe_adjust_thermodynamic_state` in order to adjust only cloud liquid + ice rather than total liquid + ice.

RICO seems to run and work, but need more testing.

### DIFF
--- a/examples/rico.jl
+++ b/examples/rico.jl
@@ -221,11 +221,12 @@ set!(model, θ=θᵢ, qᵗ=qᵢ, u=uᵢ, v=vᵢ)
 
 # ## Simulation
 #
-# We run the simulation for 12 hours with adaptive time-stepping.
+# We run the simulation for 8 hours with adaptive time-stepping.
 # RICO typically requires longer integration times than BOMEX to develop
-# a quasi-steady precipitating state.
+# a quasi-steady precipitating state, and should be run for 24 hours.
+# We choose 8 hours here to save computational costs in building the examples.
 
-simulation = Simulation(model; Δt=2, stop_time=12hour)
+simulation = Simulation(model; Δt=2, stop_time=8hour)
 conjure_time_step_wizard!(simulation, cfl=0.7)
 
 # ## Output and progress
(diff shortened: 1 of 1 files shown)


test/cloud_microphysics.jl
giordano: We should split this into multiple files for better load balancing of the tests, this file alone takes nearly 100% of walltime of tests on some platforms.

docs/make.jl
```diff
@@ -35,16 +35,22 @@ import Pkg
...
-    "Cloudy Kelvin-Helmholtz instability" => "literated/cloudy_kelvin_helmholtz.md",
-    "Shallow cumulus convection (BOMEX)" => "literated/bomex.md",
-    "Precipitating shallow cumulus (RICO)" => "literated/rico.md",
-    "Convection over prescribed sea surface temperature (SST)" => "literated/prescribed_sea_surface_temperature.md",
-    "Inertia gravity wave" => "literated/inertia_gravity_wave.md",
-    "Single column gray radiation" => "literated/single_column_radiation.md",
-    "Stationary parcel model" => "literated/stationary_parcel_model.md",
+struct Example
+    title::String
+    filename::String
```
navidcy: if we define a struct, and if this is "filename" then it makes sense to expect the `example.jl`, not the `.md`?

navidcy: true

PR #285, opened 2025-12-15
Updates radiation docs

navidcy:
Preview at https://numericalearth.github.io/BreezeDocumentation/previews/PR285/


docs/src/radiative_transfer.md
```diff
@@ -54,21 +54,23 @@ The [`RadiativeTransferModel`](@ref) model computes:
...
-τ_{lw} = α \frac{Δp}{p} \left[ f_l \frac{p}{p_0} + 4 (1 - f_l) \left(\frac{p}{p_0}\right)^4 \right] \left[ τ_e + (τ_p - τ_e) \sin^2 φ \right]
+τ_{lw} = α \frac{Δp}{p_0} \left[ f_l + 4 (1 - f_l) \left(\frac{p}{p_0}\right)^3 \right] \left[ τ_e + (τ_p - τ_e) \sin^2 φ \right]
 ```
 
 where ``φ`` is latitude and ``α``, ``f_l``, ``τ_e``, ``τ_p`` are empirical parameters.
 
 For shortwave:
 ```math
-τ_{sw} = 2 τ_0 \frac{p}{p_0} \frac{Δp}{p_0}
+τ_{sw} = 2 τ_0 \frac{Δp}{p_0} \frac{p}{p_0}
```
glwagner: what's our feeling about "lw" and "sw" for longwave and shortwave? We could also use LW and SW. I don't think we have subscripts in unicode so we may need superscripts?

navidcy: but are these used anywhere in the code?
if not, then do they need to be in the notation.md?

PR #291, opened 2025-12-16
Implement SSPRungeKutta3TimeStepper

glwagner:
This PR implements the "strong stability preserving" (SSP) 3rd order Runge-Kutta time stepper described by Shu and Osher and makes it the default scheme for AtmosphereModel. Future PRs will build on this one to support things like implicit time stepping for microphysical tendencies, vertically-implicit diffusion, etc. I believe this is a drop-in replacement for hte existing RK3 time-stepper we were using, with the same computational cost and slightly better stability properties (probably).

### DIFF
--- a/docs/src/api.md
+++ b/docs/src/api.md
@@ -77,6 +77,13 @@ Modules = [Breeze.TurbulenceClosures]
 Private = false
 ```
 
+### TimeSteppers
+
+```@autodocs
+Modules = [Breeze.TimeSteppers]
+Private = false
+```
+
 ### Forcings
 
 ```@autodocs
@@ -186,6 +193,13 @@ Public = false
 
 ```@autodocs
 Modules = [Breeze.Forcings]
+Public = false
+```
+
+### TimeSteppers
+
+```@autodocs
+Modules = [Breeze.TimeSteppers]
 Public = false
 ```
 


glwagner:
This isn't working, so I will be opening up a PR with a new attempt soon.

src/Advection.jl
```diff
@@ -26,6 +26,10 @@ using ..AtmosphereModels: AtmosphereModels, div_ρUc
 @inline tracer_mass_flux_z(i, j, k, grid, ρ, args...) =
     ℑzᵃᵃᶠ(i, j, k, grid, ρ) * _advective_tracer_flux_z(i, j, k, grid, args...)
 
+@inline tracer_mass_flux_x(i, j, k, grid, ρ, advection::BoundsPreservingWENO, args...) = zero(grid)
+@inline tracer_mass_flux_y(i, j, k, grid, ρ, advection::BoundsPreservingWENO, args...) = zero(grid)
+@inline tracer_mass_flux_z(i, j, k, grid, ρ, advection::BoundsPreservingWENO, args...) = zero(grid)
+
```
glwagner:
```suggestion
```

src/TimeSteppers/ssp_rk3.jl
```diff
@@ -0,0 +1,222 @@
...
+    compute_tendencies!,
+    compute_flux_bc_tendencies!,
+    compute_pressure_correction!,
+    make_pressure_correction!,
+    step_lagrangian_particles!
+
+import Oceananigans.TimeSteppers: time_step!
+
+"""
+    SSPRungeKutta3TimeStepper{FT, U0, TG, TI} <: AbstractTimeStepper
```
giordano:
```suggestion
$(TYPEDEF)
```

glwagner: what's your opinion about docstrings on structs? In some previous lifetime we decided they were not a good idea, since users have no use for them, and they pollute the help query.

giordano: I think they are still useful, at least for other contributors/developers if these structs aren't meant to be consumed by end-users.

I understand the argument of polluting the search results, but perhaps that points to a higher infrastructure missing feature in Documenter, but I don't think the result should be to document _fewer_ things.

glwagner: i'll merge this but i want to see the docs build first

PR #293, opened 2025-12-17
Add a default NaNChecker to AtmosphereModel

### LINKED ISSUE #283: Add `default_nan_checker`
I noticed that simulations that NaN run forever. This is because `AtmosphereModel` does not define `Oceananigans.Diagnostics.default_nan_checker`. We should improve the situation and define it. I think `model.momentum.ρu` is probably the right choice.

navidcy:
Closes #283 

cc @mmr0

### DIFF
--- a/test/atmosphere_model_unit.jl
+++ b/test/atmosphere_model_unit.jl
@@ -9,7 +9,7 @@ function run_nan_checker_test(arch; erroring)
     grid = RectilinearGrid(arch, size=(4, 2, 1), extent=(1, 1, 1))
     model = AtmosphereModel(grid)
     simulation = Simulation(model, Δt=1, stop_iteration=2)
-    model.momentum.ρu[1, 1, 1] = NaN
+    @allowscalar model.momentum.ρu[1, 1, 1] = NaN
     erroring && erroring_NaNChecker!(simulation)
 
     if erroring


src/AtmosphereModels/atmosphere_model.jl
```diff
@@ -391,3 +393,17 @@ function total_energy(model)
     e = static_energy(model) |> Field
     return k + e
 end
+
+# Check for NaNs in the first prognostic field (generalizes to prescribed velocities).
```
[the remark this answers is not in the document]
glwagner: yet

src/AtmosphereModels/atmosphere_model.jl
```diff
@@ -14,6 +15,7 @@ using Oceananigans.Utils: launch!, prettytime, prettykeys, with_tracers
 
 import Oceananigans: fields, prognostic_fields
 import Oceananigans.Advection: cell_advection_timescale
+import Oceananigans.Diagnostics: default_nan_checker
```
giordano: `import` 😢

navidcy: it must be `import` since I'm extending the method, right?

giordano: No, a better way to extend a function from another method is to qualify the function with the owning module:
```julia
Diagnostics.default_nan_checker(...) = ...
```
`import` is a bit evil because you may accidentally extend methods from other modules if don't realise you `import`ed the function before (it's a sort of spooky action at a distance).  With the `Module.function(...) = ...` syntax you make it super explicit, locally, that you're extending a function from another module.  In some cases `import` is actively harmful: https://github.com/CliMA/Oceananigans.jl/pull/5041#issuecomment-3639653393.  Just completely avoid `import` to extend methods from other modules and live happy.

navidcy: see https://github.com/NumericalEarth/Breeze.jl/pull/294

test/atmosphere_model_unit.jl
```diff
@@ -1,9 +1,27 @@
...
+    model = AtmosphereModel(grid)
+    simulation = Simulation(model, Δt=1, stop_iteration=2)
+    @allowscalar model.momentum.ρu[1, 1, 1] = NaN
+    erroring && erroring_NaNChecker!(simulation)
+
+    if erroring
+        @test_throws ErrorException run!(simulation)
+    else
+        run!(simulation)
+        @show @test model.clock.iteration == 1 # simulation stopped after one iteration
```
giordano: I presume this `@show` is a leftover?

navidcy: Yeah! Thanks for catching this. Removed in #294.

PR #294, opened 2025-12-17
Avoid import for extending methods

navidcy:
https://github.com/NumericalEarth/Breeze.jl/pull/293#discussion_r2626275066

### DIFF
--- a/src/AtmosphereModels/atmosphere_model.jl
+++ b/src/AtmosphereModels/atmosphere_model.jl
@@ -5,7 +5,7 @@ using Oceananigans: AbstractModel, Center, CenterField, Clock, Field,
 using Oceananigans.Advection: adapt_advection_order
 using Oceananigans.AbstractOperations: @at
 using Oceananigans.BoundaryConditions: FieldBoundaryConditions, regularize_field_boundary_conditions
-using Oceananigans.Diagnostics: Diagnostics, NaNChecker
+using Oceananigans.Diagnostics: Diagnostics as OceananigansDiagnostics, NaNChecker
 using Oceananigans.Grids: ZDirection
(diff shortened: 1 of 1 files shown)


giordano (approved):
🥳

src/AtmosphereModels/atmosphere_model.jl
```diff
@@ -5,7 +5,7 @@ using Oceananigans: AbstractModel, Center, CenterField, Clock, Field,
 using Oceananigans.Advection: adapt_advection_order
 using Oceananigans.AbstractOperations: @at
 using Oceananigans.BoundaryConditions: FieldBoundaryConditions, regularize_field_boundary_conditions
-using Oceananigans.Diagnostics: NaNChecker
+using Oceananigans.Diagnostics: Diagnostics, NaNChecker
```
giordano: Ah, there's already a submodule called `Diagnostics` in `AtmosphereModels`!  We have two options:

* do `using Oceananigans: Oceananigans`, and then define the method `Oceananigans.Diagnostics.default_nan_checker(...)`, or
* do `using Oceananigans.Diagnostics: Diagnostics as OceananigansDiagnostics, NaNChecker` (or some other alias), and then define the method `OceananigansDiagnostics.default_nan_checker(...)`

giordano:
> hm... locally tests passed

Uh, interesting it worked for you, because it did fail locally for me 🤔

giordano:
> There is a chance locally the tests pass at 9a9f876553ad11dd2ad0438d944647e42104d484 ... oops

Oh, in that case I suppose you may have defined a method in the `AtmosphericModels.Diagnostics` module 😬

In any case it's looking good now!

PR #295, opened 2025-12-17
Introduce a "standard pressure" parameter in the reference state

### LINKED ISSUE #229: Exner function definition differs from the standard one
We currently use 

$$ \Pi = \left ( \frac{p_r}{p_0} \right )^{R^m / c^{pm}} $$

where $p_0$ is the "Base pressure" of the formulation; ie the reference pressure at $z=0$.

I think the standard definition of the Exner function is 

$$ \Pi = \left ( \frac{p_r}{10^5} \right )^{R^m / c^{pm}} $$

although some even use the exponent $R^d / c^{pd}$ rather than the moist part (Pauluis 2008 argued that omitting moisture violates the first law)

_Originally posted by @glwagner in https://github.com/NumericalEarth/Breeze.jl/issues/228#issuecomment-3621585298_

glwagner:
This permits computing the standard form of potential temperature, at the expense of a little bit extra GPU parameter space, sadly.

Closes #229

### DIFF
--- a/.github/workflows/TagBot.yml
+++ b/.github/workflows/TagBot.yml
@@ -1,0 +1,20 @@
+name: TagBot
+
+on:
+  issue_comment:
+    types:
+      - created
+  workflow_dispatch:
+    inputs:
+      lookback:
+        default: "3"
+
+jobs:
+  TagBot:
+    if: github.event_name == 'workflow_dispatch' || github.actor == 'JuliaTagBot'
+    runs-on: ubuntu-latest
+    steps:
+      - uses: JuliaRegistries/TagBot@v1
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          ssh: ${{ secrets.DOCUMENTER_KEY }}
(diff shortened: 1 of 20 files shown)


docs/src/appendix/notation.md
```diff
@@ -80,10 +80,11 @@ The following table also uses a few conventions that suffuse the source code and
 | ``\mathcal{L}^i(T)``                | `ℒⁱ`   | `ice_latent_heat(T, constants)`     | Temperature-dependent latent heat of deposition                                |
 | ``θ₀``                              | `θ₀`   | `RS.potential_temperature`          | (Constant) reference potential temperature for the anelastic formulation       |
 | ``p₀``                              | `p₀`   | `RS.surface_pressure`               | Base (surface) reference pressure                                              |
```
navidcy: is there a default for p₀? shall we mention here?

glwagner: the default is 101325 which is the mean sea level pressure on earth

glwagner:
```suggestion
| ``p₀``                              | `p₀`   | `RS.surface_pressure`               | Surface reference pressure                                              |
```

PR #296, opened 2025-12-17
Remove all imports

giordano:
Follow up from #294.

@glwagner I put into separate commits the more contentious changes:

* `div_ρUc` has all main methods defined in `Advection` and that module even exports it, I don't see why it should be defined in `AtmosphereModels`
* `compute_forcing!` is defined in the forcing interface file in `AtmosphereModels`, but the generic method is defined in `Forcings`, I don't know, that feels backward to me 😅

### DIFF
--- a/src/BoundaryConditions/BoundaryConditions.jl
+++ b/src/BoundaryConditions/BoundaryConditions.jl
@@ -21,8 +21,7 @@ using Oceananigans.Fields: Field, set!
 using Oceananigans.BoundaryConditions: BoundaryConditions as OceananigansBC,
                                        BoundaryCondition,
                                        Flux,
-                                       FieldBoundaryConditions,
-                                       getbc
+                                       FieldBoundaryConditions
 
 using Oceananigans.Operators: ℑxyᶠᶜᵃ, ℑxyᶜᶠᵃ, ℑxᶜᵃᵃ, ℑyᵃᶜᵃ
 
--- a/src/AtmosphereModels/atmosphere_model.jl
+++ b/src/AtmosphereModels/atmosphere_model.jl
@@ -7,7 +7,7 @@ using Oceananigans.AbstractOperations: @at
 using Oceananigans.BoundaryConditions: FieldBoundaryConditions, regularize_field_boundary_conditions
 using Oceananigans.Diagnostics: Diagnostics as OceananigansDiagnostics, NaNChecker
 using Oceananigans.Grids: ZDirection
-using Oceananigans.Models: Models, boundary_condition_args, validate_model_halo, validate_tracer_advection
+using Oceananigans.Models: Models, validate_model_halo, validate_tracer_advection
 using Oceananigans.Models.HydrostaticFreeSurfaceModels: validate_momentum_advection
 using Oceananigans.Solvers: FourierTridiagonalPoissonSolver
 using Oceananigans.TimeSteppers: TimeStepper
--- a/src/MoistAirBuoyancies.jl
+++ b/src/MoistAirBuoyancies.jl
@@ -22,9 +22,7 @@ using DocStringExtensions: TYPEDSIGNATURES
 using Oceananigans: Oceananigans, Center, Field, KernelFunctionOperation
 using Oceananigans.BuoyancyFormulations: BuoyancyFormulations,
     AbstractBuoyancyFormulation,
-    buoyancy_perturbationᶜᶜᶜ,
-    ∂z_b,
-    required_tracers
+    buoyancy_perturbationᶜᶜᶜ
 using Oceananigans.Grids: AbstractGrid
 using Oceananigans.Operators: ∂zᶜᶜᶠ
 


glwagner:
> * `div_ρUc` has all main methods defined in `Advection` and that module even exports it, I don't see why it should be defined in `AtmosphereModels`
> * `compute_forcing!` is defined in the forcing interface file in `AtmosphereModels`, but the generic method is defined in `Forcings`, I don't know, that feels backward to me 😅

This design was an attempt to clarify / clean up the mess that has developed in Oceananigans. The idea is that AtmosphereModel defines the _interface_ for its computations, which are then (optionally) extended later on. This clarifies whether or not the model structure is "available" when a scheme is constructed.

The construction of an advection scheme or forcing is inevitably entangled with the model design. As an example, we use a method `materialize_forcing` to build the elements of a forcing function that depend on specific (hidden to the use) aspects of a model.

Modules that are input before AtmosphereModels become dependencies of AtmosphereModel. If we split those from Breeze, then Breeze depends on them. Whereas modules that come after become extensions which could be easily / cleanly split into other packages that depend on Breeze.

If we go this route then I think we have a situation where some model features, like advection and forcing, are dependencies of the model. Whereas others, like "microphysics", depend themselves on AtmosphereModel and can be viewed as extensions.

I'm ok with with this split design. I was just wondering if have a more uniform treatment of all model features might clarify the situation.

What are your thoughts @giordano ?

src/BoundaryConditions/BoundaryConditions.jl
```diff
@@ -12,13 +12,14 @@ export BulkDragFunction,
...
-import ..AtmosphereModels: regularize_atmosphere_model_boundary_conditions
-
+using ..AtmosphereModels: AtmosphereModels
 using ..Thermodynamics: saturation_specific_humidity, surface_density, PlanarLiquidSurface
 
+using Oceananigans.Architectures: Architectures, on_architecture
 using Oceananigans.Grids: Center, Face, XDirection, YDirection, AbstractGrid
 using Oceananigans.Fields: Field, set!
-using Oceananigans.BoundaryConditions: BoundaryCondition,
+using Oceananigans.BoundaryConditions: BoundaryConditions as OBC,
```
navidcy:
```suggestion
using Oceananigans.BoundaryConditions: BoundaryConditions as OceananigansBC,
```

src/BoundaryConditions/BoundaryConditions.jl
```diff
@@ -131,7 +129,7 @@ const YDBDF = YDirectionBulkDragFunction
     return - Cᴰ * Ũ² * ρu / U * (U > 0)
 end
 
-@inline function getbc(df::YDBDF, i::Integer, j::Integer, grid::AbstractGrid, clock, fields)
+@inline function OBC.getbc(df::YDBDF, i::Integer, j::Integer, grid::AbstractGrid, clock, fields)
```
navidcy:
```suggestion
@inline function OceananigansBC(df::YDBDF, i::Integer, j::Integer, grid::AbstractGrid, clock, fields)
```

src/BoundaryConditions/BoundaryConditions.jl
```diff
@@ -196,8 +194,8 @@ Base.summary(bf::BulkSensibleHeatFluxFunction) =
            ", gustiness=", bf.gustiness, ")")
 
 # getbc for BulkSensibleHeatFluxFunction
-@inline function getbc(bf::BulkSensibleHeatFluxFunction, i::Integer, j::Integer,
-                       grid::AbstractGrid, clock, fields)
+@inline function OBC.getbc(bf::BulkSensibleHeatFluxFunction, i::Integer, j::Integer,
+                           grid::AbstractGrid, clock, fields)
```
navidcy:
```suggestion
@inline function OceananigansBC.getbc(bf::BulkSensibleHeatFluxFunction, i::Integer, j::Integer,
                                      grid::AbstractGrid, clock, fields)
```

src/BoundaryConditions/BoundaryConditions.jl
```diff
@@ -274,7 +272,7 @@ Base.summary(bf::BulkVaporFluxFunction) =
 const BVFF = BulkVaporFluxFunction
 
 # getbc for BulkVaporFluxFunction
-@inline function getbc(bf::BVFF, i::Integer, j::Integer, grid::AbstractGrid, clock, fields)
+@inline function OBC.getbc(bf::BVFF, i::Integer, j::Integer, grid::AbstractGrid, clock, fields)
```
navidcy:
```suggestion
@inline function OceananigansBC.getbc(bf::BVFF, i::Integer, j::Integer, grid::AbstractGrid, clock, fields)
```

navidcy (approved):
just one suggestion but really it's aesthetic and feel free to ignore

glwagner:
> > This design was an attempt to clarify / clean up the mess that has developed in Oceananigans. The idea is that AtmosphereModel defines the interface for its computations, which are then (optionally) extended later on. This clarifies whether or not the model structure is "available" when a scheme is constructed.
> 
> Ok, I agree with the idea, but I feel like that if there needs to be a generic fallback method, like for `compute_forcing!`, it should be in the module defining the interface. What do you think?

I guess the central question _is_ whether AtmosphereModels should define the interface or whether Forcing should define the interface. It seems quite clear that AtmosphereModels must define the interface for microphysics, because it is fairly intricate and requires ~10 functions to be sprinkled throughout a time-step, just-so. For forcing, it is more ambiguous because we are only talking about a single function `compute_forcing`. For advection similarly, there are only a handful of functions that constitute the interface.

It is probably valid to think of AtmosphereModels as the user rather than owner of these very thin/small interfaces.

giordano:
> Should we open a dedicated issue for that question?

Issue created at #299.  Is this PR good to go now? 😁

glwagner (approved):

PR #300, opened 2025-12-18
Clean implementation of Kessler microphysics

kaiyuan-cheng:


glwagner:
Are we sure we want to maintain two versions of Kessler microphysics?

PR #302, opened 2025-12-18
[CI] Tweak command line flags to reduce precompilation work

giordano:


glwagner:
could make sense to run the tests with O0 -- we do that in oceananigans

PR #305, opened 2025-12-19
Make LiquidIcePotentialTemperature the default thermodynamics

glwagner:
this saves a few lines here and there.


navidcy (approved):
Happy for this. There is a small issue with some of the rest examples in the examples dir; are we sure they are also up to date? Perhaps we should only keep examples there that are included in the docs?

glwagner:
> Looks like this made the tests a lot slower? _**Edit**_: apparently only on macOS, on other platforms the difference is much more negligible:
> 
> * [main](https://github.com/NumericalEarth/Breeze.jl/actions/runs/20350577025/job/58473841424#step:7:1117) ([b436cfa](https://github.com/NumericalEarth/Breeze.jl/commit/b436cfa7356bc9646407bd7cb91b72c82e09abee)): 9m07.4s
> * [this PR](https://github.com/NumericalEarth/Breeze.jl/actions/runs/20362716162/job/58511040241#step:7:404): 16m18.8s

oof, i wonder if that means that the potential temperature formulation is slower for some reason (that would be surprising)

PR #306, opened 2025-12-19
Update compute_forcing! to handle multiple forcings

glwagner:
I think that forcings were not properly computed for multiple forcings before this.

### DIFF
--- a/src/AtmosphereModels/atmosphere_model.jl
+++ b/src/AtmosphereModels/atmosphere_model.jl
@@ -52,7 +52,7 @@ mutable struct AtmosphereModel{Frm, Arc, Tst, Grd, Clk, Thm, Mom, Moi, Mfr, Buy,
     timestepper :: Tst
     closure :: Cls
     closure_fields :: Cfs
-    radiative_transfer :: Rad
+    radiation :: Rad
 end
 
 # Note: Formulation-specific functions (default_formulation, materialize_formulation, etc.)
@@ -119,7 +119,7 @@ function AtmosphereModel(grid;
                          closure = nothing,
                          microphysics = nothing, # WarmPhaseSaturationAdjustment(),
                          timestepper = :RungeKutta3,
-                         radiative_transfer = nothing)
+                         radiation = nothing)
 
     if !(advection isa DefaultValue)
         # TODO: check that tracer+momentum advection were not independently set.
@@ -221,7 +221,7 @@ function AtmosphereModel(grid;
                             timestepper,
                             closure,
                             closure_fields,
-                            radiative_transfer)
+                            radiation)
 
     update_state!(model)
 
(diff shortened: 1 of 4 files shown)


src/Forcings/Forcings.jl
```diff
@@ -32,6 +32,11 @@ function AtmosphereModels.materialize_atmosphere_model_forcing(forcings::Tuple,
 end
 
 # Handle compute_forcing! for MultipleForcings
-AtmosphereModels.compute_forcing!(mf::MultipleForcings) = compute_forcing!(mf.forcings)
+function AtmosphereModels.compute_forcing!(mf::MultipleForcings)
+    for forcing in mf.forcings
+        compute_forcing!(forcing)
+    end
+    return nothing
+end
```
giordano: How did it work before? 🤔 Or it didn't?  Are we missing tests?

glwagner: i don't think it did.

giordano: But then we're missing relevant tests?

glwagner: yes for sure

there is a test but it is too basic / loose, so it passed without this being caught.

navidcy:
Oh gotcha

src/AtmosphereModels/atmosphere_model.jl
```diff
@@ -100,21 +100,20 @@ function AtmosphereModel(grid;
                          coriolis = nothing,
                          boundary_conditions = NamedTuple(),
                          forcing = NamedTuple(),
-                         advection = nothing,
-                         momentum_advection = nothing,
-                         scalar_advection = nothing,
+                         advection = DefaultValue(),
+                         momentum_advection = DefaultValue(),
+                         scalar_advection = DefaultValue(),
```
navidcy: add explanation for how to use these kwargs or which take precedence over which in the docstring?

glwagner: ok -- but scope creep! this was already present before this PR

src/AtmosphereModels/atmosphere_model.jl
```diff
@@ -100,21 +113,20 @@ function AtmosphereModel(grid;
...
                          closure = nothing,
                          microphysics = nothing, # WarmPhaseSaturationAdjustment(),
                          timestepper = :RungeKutta3,
-                         radiation = nothing,
```
giordano: Radiation was used in https://github.com/NumericalEarth/Breeze.jl/blob/2570353a3c96bdbb1f49a0090c26e708ecf2673f/test/radiative_transfer.jl

giordano: Makes sense, given that one was being set equal to the other.

PR #333, opened 2025-12-22
Clean up mountain wave case

### LINKED ISSUE #332: Clean up mountain wave cases
There are two mountain wave cases running the same simulation. one under ./examples and the other under ./validation/topography. I think we should keep just one to tidy up the codebase. The question is which folder we should use. For simplicity, I suggest using ./example for the following reasons. First, there is no need to revise the current documentation workflow to showcase the mountain wave case. Second, the mountain wave simulation we obtain is close but not fully correct, due to the numerical challenges in accurately and efficiently representing low-elevation terrain.

kaiyuan-cheng:
close #332

### DIFF
--- a/examples/two_dimensional_mountain_wave.jl
+++ b/examples/two_dimensional_mountain_wave.jl
@@ -291,7 +291,7 @@ run!(simulation)
 # \right]
 # ```
 
-hhat(k) = sqrt(π) * h₀ * a / 4 * (exp(-a^2 * (K + k)^2 / 4) +
+ĥ(k) = sqrt(π) * h₀ * a / 4 * (exp(-a^2 * (K + k)^2 / 4) +
                                 exp(-a^2 * (K - k)^2 / 4) +
                                 2exp(-a^2 * k^2 / 4))
 
@@ -326,11 +326,8 @@ Compute the 2-D linear vertical velocity `w(x,z)` from the analytical solution
 """
 function w_linear(x, z; nk=100)
     k = range(0, 10k★; length=nk)
-    m2 = m².(k)
-    ĥ = hhat.(k)
-
-    m_abs = sqrt.(abs.(m2))
-    integrand = @. k * ĥ * ifelse(m2 ≥ 0,
+    m_abs = @. sqrt(abs(m²(k)))
+    integrand = @. k * ĥ(k) * ifelse(m²(k) ≥ 0,
                                    sin(m_abs * z + k * x),
                                    exp(-m_abs * z) * sin(k * x))
 


examples/two_dimensional_mountain_wave.jl
```diff
@@ -0,0 +1,387 @@
...
+end
+
+# ## Results: Comparison with analytical solution
+#
+# We compare the simulated vertical velocity field at 2 hours with the linear
+# analytical solution. While the simulation reproduces the general gravity wave
+# pattern, noticeable discrepancies in wavenumber appear. The immersed boundary
+# method struggles to resolve the low-amplitude, fine-scale terrain corrugations
+# at this resolution. A terrain-following coordinate would be better suited for
+# this test case.
```
glwagner:
```suggestion
# at this resolution.
```

its probably true, but I think its best to avoid speculation

examples/two_dimensional_mountain_wave.jl
```diff
@@ -0,0 +1,386 @@
...
+
+# Derived atmospheric parameters:
+
+N² = g^2 / (cᵖᵈ * T₀)       # s⁻² - Brunt–Väisälä frequency squared
+N  = sqrt(N²)               # s⁻¹ - Brunt–Väisälä frequency
+β  = g / (Rᵈ * T₀)          # m⁻¹ - density scale parameter
+
+# ## Schär mountain parameters
+#
+# The mountain profile parameters following [Schar2002](@cite):
```
navidcy:
```suggestion
# The mountain profile parameters following [Schar2002](@citet):
```

examples/two_dimensional_mountain_wave.jl
```diff
@@ -0,0 +1,386 @@
...
+
+# band! requires arrays, not Fields
+x_ana = xnodes(h_analytical)
+h_ana = interior(h_analytical, :, 1, 1)
+band!(ax_terrain, x_ana, zeros(length(x_ana)), h_ana, color = (:brown, 0.2))
+xlims!(ax_terrain, -30e3, 30e3)
+axislegend(ax_terrain, position = :rt)
+
+save("mountain_wave_terrain.png", fig_terrain)
+fig_terrain
```
navidcy:
```suggestion
# The hill profile on a high-resolution grid:
x_finegrid = -L/3:L/1000:L/3
h_finegrid = hill.(x_finegrid)

# Discretized profile as represented in the model:
h_model = grid.immersed_boundary.bottom_height

fig = Figure(size=(900, 400))
ax = Axis(fig[1, 1],
          xlabel = "x (m)",
          ylabel = "Height (m)",
          title = "Schär Mountain Profile")

lines!(ax, x_finegrid, h_finegrid, linewidth = 1, color = :black, label = "Analytical")
lines!(ax, h_model, linewidth = 2, color = :brown, linestyle = :dash, label = "Model")
band!(ax, x_finegrid, zeros(length(x_finegrid)), h_finegrid, color = (:brown, 0.2))

axislegend(ax, position = :rt)
xlims!(ax, extrema(x_finegrid)...)
fig
```

examples/two_dimensional_mountain_wave.jl
```diff
@@ -0,0 +1,386 @@
...
+# ```math
+# \theta(z) = \theta_0 \exp\left(\frac{N^2 z}{g}\right)
+# ```
+
+θᵢ(x, z) = θ₀ * exp(N² * z / g)
+set!(model, θ = θᵢ, u = U)
+
+# ## Simulation
+#
+# Run for 2 hours with a fixed time step. A small time step is required for numerical stability with this test case:
```
navidcy: simulation blows up for me; even when I changed to 

```Julia
simulation = Simulation(model; Δt=0.5, stop_time=2hours)
conjure_time_step_wizard!(simulation, cfl=0.7)
```

I don't know why... :(

examples/two_dimensional_mountain_wave.jl
```diff
@@ -0,0 +1,386 @@
...
+# Calculate grid distribution:
+
+Nz_top = ceil(Int, (H - z_transition) / dz_top)     # cells in uniform region
+Nz_bottom = Nz - Nz_top                             # cells in stretched region
+
+# Construct hybrid vertical grid:
+
+z_stretched = ExponentialDiscretization(Nz_bottom, 0, z_transition, scale = z_transition / 8, bias=:left)
+z_uniform = range(z_transition + dz_top, H; length=Nz_top)
+z_faces = vcat(z_stretched.faces, collect(z_uniform))
```
giordano: This is printing the entire array: https://numericalearth.github.io/BreezeDocumentation/previews/PR333/literated/two_dimensional_mountain_wave/#Grid-configuration
```suggestion
z_faces = vcat(z_stretched.faces, collect(z_uniform))
nothing #hide
```

examples/two_dimensional_mountain_wave.jl
```diff
@@ -0,0 +1,386 @@
...
+#
+# Save velocity fields for post-processing and validation:
+
+filename = "mountain_waves"
+simulation.output_writers[:fields] = JLD2Writer(model, model.velocities;
+                                                filename,
+                                                schedule = TimeInterval(100),
+                                                overwrite_existing = true)
+
+run!(simulation)
```
giordano: Simulation is stopping because there are NaNs: https://numericalearth.github.io/BreezeDocumentation/previews/PR333/literated/two_dimensional_mountain_wave/#Output
```
[ Info: time = 600.0, iteration = 300: NaN found in field ρu. Stopping simulation.
```

kaiyuan-cheng: It’s weird that the simulation blows up in CI but not on my end. 😢

giordano: What system? Can you show the output of `versioninfo()` in the REPL? You can see at https://numericalearth.github.io/BreezeDocumentation/previews/PR333/literated/two_dimensional_mountain_wave/#Julia-version-and-environment-information the information about the CI machine.

kaiyuan-cheng: The CI system uses an AMD CPU, unlike the two machines I just tested. Could that be the issue?

giordano: One thing I did notice is that the RNG seems to generate different streams of numbers for different GPUs models/generations, but I don't think we're using random numbers here, right?

examples/two_dimensional_mountain_wave.jl
```diff
@@ -0,0 +1,374 @@
...
+# ability to capture terrain-induced gravity waves in a stably stratified atmosphere. The test
+# case features a bell-shaped mountain with superimposed small-scale corrugations, generating
+# both propagating and evanescent wave modes.
+#
+# ## References
+#
+# ```@bibliography
+# Pages = ["literated/two_dimensional_mountain_wave.md"]
+# Canonical = false
+# ```
```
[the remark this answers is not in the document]
navidcy: gotcha

examples/two_dimensional_mountain_wave.jl
giordano: I could replicate the NaNs with
```julia-repl
julia> versioninfo()
Julia Version 1.12.2
Commit ca9b6662be4 (2025-11-20 16:25 UTC)
Build Info:
  Official https://julialang.org release
Platform Info:
  OS: Linux (x86_64-linux-gnu)
  CPU: 192 × AMD Ryzen Threadripper PRO 7995WX 96-Cores
  WORD_SIZE: 64
  LLVM: libLLVM-18.1.7 (ORCJIT, znver4)
  GC: Built with stock GC
Threads: 1 default, 1 interactive, 1 GC (on 192 virtual cores)
```
with NVIDIA GeForce RTX 5090.

kaiyuan-cheng: > On the same machine I could also replicate the NaN with the CPU architecture (which I presume is what also @navidcy tested), i.e. by replacing `GPU()` at
> 
> https://github.com/NumericalEarth/Breeze.jl/blob/9743747540a39d4db5917a9685ce91a0b0798fc1/examples/two_dimensional_mountain_wave.jl#L145
> 
> to `CPU()`. @kaiyuan-cheng on the machines where you don't have NaNs, what do you get with the CPU simulation? It fails in 80 seconds for me, it doesn't run for long.

This case runs fine on both CPU and GPU on the two machines I tested. Very strange.

examples/two_dimensional_mountain_wave.jl
```diff
@@ -0,0 +1,374 @@
...
+#
+# The Fourier transform of the Schär mountain profile (Equation A8):
+#
+# ```math
+# \hat{h}(k) = \frac{\sqrt{\pi} h_0 a}{4} \left[
+#     e^{-a^2(K+k)^2/4} + e^{-a^2(K-k)^2/4} + 2e^{-a^2 k^2/4}
+# \right]
+# ```
+
+ĥ(k) = sqrt(π) * h₀ * a / 4 * (exp(-a^2 * (K + k)^2 / 4) +
```
giordano: @navidcy in 0e52c18ddd6f694bf16d77255f488222a983fab8 you renamed this function from `hhat`, but then you didn't change it in https://github.com/NumericalEarth/Breeze.jl/blob/b8744820380cdc101770057d58aafaaf2db19780/examples/two_dimensional_mountain_wave.jl#L330 but in that function there's already a local variable called `ĥ`, so there's a naming issue here 🙂

giordano: I'm reverting the name change just to get this working again, then we can discuss the naming issue
```suggestion
hhat(k) = sqrt(π) * h₀ * a / 4 * (exp(-a^2 * (K + k)^2 / 4) +
```

PR #347, opened 2025-12-24
[CI] Start julia process for running tests with `--check-bounds=yes`

giordano:
This should avoid precompiling the `CUDA_Runtime_jll` on the GPU jobs using the Docker image, because that one doesn't have cached that package for `--check-bounds=auto`

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -106,7 +106,7 @@ jobs:
               # We use earlyoom only on the GPU machines
               earlyoom -m 3 -s 100 -r 300 --prefer 'julia' &
           fi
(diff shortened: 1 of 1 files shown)


navidcy:
Hm... it wasn't?

navidcy (approved):

navidcy:
Only asking because, at least locally, when I was doing

```Julia
] test
```

I would get errors related with bounds checking.

giordano:
Note how that extra precompilation isn't present in this PR: https://github.com/NumericalEarth/Breeze.jl/actions/runs/20495660391/job/58894829563?pr=347#step:8:18.  Also, this is only a flag passed to the process launching the tests, the tests are run with `--check-bounds=yes` already.

Issue #351, opened 2025-12-25
Non-reproducible simulations


giordano:
Continuing from https://github.com/NumericalEarth/Breeze.jl/pull/344#issuecomment-3690538888, and somewhat similar to #149, in v0.2.1 this simulation https://github.com/NumericalEarth/Breeze.jl/blob/5d144a1ec2811c1df2a77e67e743914f2be56cc7/docs/src/index.md?plain=1#L52-L89 produces different output figures:

* [in the docs](https://numericalearth.github.io/BreezeDocumentation/v0.2/#Quick-Start) ([information about the system](https://numericalearth.github.io/BreezeDocumentation/v0.2/literated/dry_thermal_bubble/#Julia-version-and-environment-information)):
  <img width="1200" height="900" alt="Image" src="https://github.com/user-attachments/assets/a9266f96-7958-4d82-947b-9801387fb41c" />

* on my macbook, julia v1.11:
  ```julia-repl
  julia> versioninfo()
  Julia Version 1.11.8
  Commit cf1da5e20e3 (2025-11-06 17:49 UTC)
  Build Info:
    Official https://julialang.org/ release
  Platform Info:
    OS: macOS (arm64-apple-darwin24.0.0)
    CPU: 8 × Apple M1
    WORD_SIZE: 64
    LLVM: libLLVM-16.0.6 (ORCJIT, apple-m1)
  Threads: 1 default, 0 interactive, 1 GC (on 4 virtual cores)
  ```
  <img width="1200" height="900" alt="Image" src="https://github.com/user-attachments/assets/453498fb-51ec-44b8-afb3-af76e21ddea7" />
* on my macbook, julia v1.12:
  ```julia-repl
  julia> versioninfo()
  Julia Version 1.12.2
  Commit ca9b6662be4 (2025-11-20 16:25 UTC)
  Build Info:
    Official https://julialang.org release
  Platform Info:
    OS: macOS (arm64-apple-darwin24.0.0)
    CPU: 8 × Apple M1
    WORD_SIZE: 64
    LLVM: libLLVM-18.1.7 (ORCJIT, apple-m1)
    GC: Built with stock GC
  Threads: 1 default, 1 interactive, 1 GC (on 4 virtual cores)
  ```
  <img width="1200" height="900" alt="Image" src="https://github.com/user-attachments/assets/6f727933-4303-4e1d-8e7f-dda18b04c649" />
* on @navidcy's system:
[...]

PR #352, opened 2025-12-25
[docs] Reduce duration of quick start simulation for debugging

giordano:
This is only for investigating #351.  On my macbook, with both v1.11 and v1.12 stopping the simulation after 1 hour (1590 steps) I get 
image
I want to see if we get the same also on the CI machine.

### DIFF
--- a/docs/src/developer/microphysics_interface.md
+++ b/docs/src/developer/microphysics_interface.md
@@ -38,7 +38,7 @@ To illustrate the development of a new microphysics scheme, we implement a
 simple microphysics scheme that represents droplet and ice particle nucleation
 with constant-rate relaxation of specific humidity to saturation.
 
-```@example microphysics_interface
+```julia
 using Breeze
 
 struct ExplicitMicrophysics{FT}
@@ -52,7 +52,7 @@ end
 This scheme is fully prognostic, which means we must carry around vapor, liquid
 and ice density as prognostic variables,
 
-```@example microphysics_interface
+```julia
 import Breeze.AtmosphereModels: prognostic_field_names
 
 prognostic_field_names(::ExplicitMicrophysics) = (:ρqᵛ, :ρqˡ, :ρqⁱ)
@@ -67,7 +67,7 @@ prognostic_field_names(::ExplicitMicrophysics) = (:ρqᵛ, :ρqˡ, :ρqⁱ)
 When we materialize the microphysics fields, we must include all of the prognostic fields
 in addition to diagnostic fields (this behavior may change in the future):
 
-```@example microphysics_interface
+```julia
 import Breeze.AtmosphereModels: materialize_microphysical_fields
 
 function materialize_microphysical_fields(::ExplicitMicrophysics, grid, boundary_conditions)
@@ -80,7 +80,7 @@ end
 ```
 The tendencies for 
 
-```@example microphysics_interface
+```julia
 import Breeze.AtmosphereModels: microphysical_tendency
 
 using Breeze.Thermodynamics:
@@ -123,7 +123,7 @@ in addition to the three prognostic fields representing vapor, liquid and ice de
 
 ### Prognostic field names and materializing prognostic + diagnostic fields
 
-```@example microphysics_interface
+```julia
 import Breeze.AtmosphereModels:
     update_microphysical_fields!,
     compute_moisture_fraction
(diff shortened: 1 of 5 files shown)


giordano:
https://numericalearth.github.io/BreezeDocumentation/previews/PR352/#Quick-Start is the same as what I posted above (obtained with both v1.11 and v1.12), which confirms that there's some divergence emerging at a later point (around ~2200 iterations based on my previous experiments)

glwagner:
We are not bitwise reproducible across architectures right? Or should we be?

giordano:
In general, common sources of divergence across systems are calls to external libraries (BLAS is a major offender, with kernels written in assembly for different ISAs and architectures,they often do wildly different things), but I don't think we use much BLAS here, do we? For pure Julia IEEE-compliant code (e.g. which doesn't do fast-math-like optimisations) you should expect full reproducibility. SIMD optimisations can also cause differences (accumulation is done differently on different CPUs, especially if vector widths are different), because they slightly violate IEEE rules. Perhaps this could explain the divergence? But I'm not sure, I haven't pinpointed the culprit.

glwagner:
Which FFTW flag are you using? The plan generation process is not reproducible for all planner flags; some of them are empirical and built on-demand (so can change depending on the state of your machine)

glwagner:
> > Which FFTW flag are you using?
> 
> Whatever is the default flag in FFTW.jl (or Oceananigans.jl, if it uses other flags by default). But again, I don't think the FFT is relevant here because it doesn't explain differences across different Julia versions.

I think the default flag is `FFTW.PATIENT`, but we need `FFTW.ESTIMATE` for reproducibility. There's some discussion here: https://github.com/CliMA/Oceananigans.jl/discussions/2790

With `FFTW.PATIENT`, the results are not even reproducible from run to run on the same julia version and same architecture. So I think you will have to control the FFTW plan to be able to investigate if you are not bit reproducible across julia versions or architectures.

giordano:
Starting julia with `-O0` I get 
image
with both v1.11 and v1.12.  I want to see if I can reproduce the same on the CI machine under the same condition.

giordano:
> I think there also may be a fastmath call here?
> 
> [CliMA/Oceananigans.jl@2ea9a4d/src/Utils/newton_div.jl#L11](https://github.com/CliMA/Oceananigans.jl/blob/2ea9a4ddd4993e1031fef4dee5499e9f9456e376/src/Utils/newton_div.jl#L11)

In theory yes, but in practice on real input `@fastmath inv(x)` on CPU is just the same as `inv(x)`:
```julia-repl
julia> which(Base.FastMath.inv_fast, (Float64,))
inv_fast(xs...)
     @ Base.FastMath fastmath.jl:250
```
https://github.com/JuliaLang/julia/blob/571fa80059a6cdad3ed6a1304e54e08226988d24/base/fastmath.jl#L250 (The [CUDA.jl backend does something different](https://github.com/JuliaGPU/CUDA.jl/blob/5d9474ae73fab66989235f7ff4fd447d5ee06f8e/src/device/intrinsics/math.jl#L389), but that's not relevant here, all simulations are run on CPU).

navidcy:
> @navidcy now I'm curious to see if you can get [#352 (comment)](https://github.com/NumericalEarth/Breeze.jl/pull/352#issuecomment-3693389915) on your macbook when building the docs on this branch with `-O0 --threads=1` (or no `--threads`, which is pretty much the same)

Using `#main` I did:

```bash
julia --color=yes --project=docs/ -e 'using Pkg; Pkg.develop(PackageSpec(path=pwd())); Pkg.instantiate()'
```

and then
```Julia
/Users/navid/.juliaup/bin/julia +1.12.2 --color=yes -O0 --threads=1 --project=docs docs/make.jl
```

and got 

```Julia
ERROR: LoadError: ArgumentError: Semaphore size must be > 0
Stacktrace:
 [1] Base.Semaphore(sem_size::Int64)
   @ Base ./lock.jl:509
 [2] top-level scope
   @ ~/Library/CloudStorage/OneDrive-TheUniversityofMelbourne/Documents/Research/Breeze.jl-v3/docs/make.jl:47
 [3] include(mod::Module, _path::String)
   @ Base ./Base.jl:306
 [4] exec_options(opts::Base.JLOptions)
   @ Base ./client.jl:317
 [5] _start()
   @ Base ./client.jl:550
in expression starting at /Users/navid/Library/CloudStorage/OneDrive-TheUniversityofMelbourne/Documents/Research/Breeze.jl-v3/docs/make.jl:47
```

Then I removed all examples and commented out lines:
```julia
semaphore = Base.Semaphore(Threads.nthreads(:interactive))
@time "literate" @sync for example in examples
    script_file = example.basename * ".jl"
    script_path = joinpath(examples_src_dir, script_file)
    Threads.@spawn :interactive Base.acquire(semaphore) do
        run(`$(Base.julia_cmd()) --color=yes --project=$(dirname(Base.active_project())) $(joinpath(@__DIR__, "literate.jl")) $(script_path) $(literated_dir)`)
    end
end
```

and repeat and got:

Screenshot 2025-12-27 at 2 38 34 pm

PR #354, opened 2025-12-27
[docs] Add a note about reproducibility of the simulations

giordano:
Running code in documentation but getting different results than what's shown there is a bummer, because then you start wondering: did I do something wrong?  Is it expected?  What's the reason for the differences?  Can I control them?  Adding these notes, mentioning some sources of non-reproducibility (and how to limit them) is my attempt to "resolve" #351.

### DIFF
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -94,10 +94,10 @@ makedocs(
         "Dycore equations and algorithms" => "dycore_equations_algorithms.md",
         "Appendix" => Any[
             "Notation" => "appendix/notation.md",
+            "Reproducibility of Breeze.jl models" => "reproducibility.md",
         ],
         "References" => "references.md",
         "API" => "api.md",
-        "Reproducibility of Breeze.jl models" => "reproducibility.md",
         "Contributors guide" => "contributing.md",
     ],
     linkcheck = true,


docs/src/reproducibility.md
```diff
@@ -0,0 +1,32 @@
+# [Reproducibility of Breeze.jl models](@id reproducibility)
+
+`Breeze.jl` cannot guarantee full reproducibility of its atmospheric models, especially for long-running simulations, due to their [chaotic nature](https://en.wikipedia.org/wiki/Chaos_theory).
```
glwagner:
```suggestion
`Breeze.jl` cannot guarantee _bitwise_ reproducibility of atmospheric simulations across Julia versions and machine architectures.
For [chaotic](https://en.wikipedia.org/wiki/Chaos_theory) simulations, bitwise differences between two initially identical solutions can amplify, leading to qualitatively different trajectories after long simulation times.
```

docs/make.jl
```diff
@@ -97,6 +97,7 @@ makedocs(
         ],
         "References" => "references.md",
         "API" => "api.md",
+        "Reproducibility of Breeze.jl models" => "reproducibility.md",
```
glwagner: this could also go in "appendix", I am fine either way

giordano: Done!

PR #334, opened 2025-12-22
Clear sky `optics` for `RadiativeTransferModel` from `RRTMGP.jl`

glwagner:
This PR expands the `RRTMGP` extensions to support clear sky gas optics, in addition to gray atmosphere optics.

PR #361, opened 2025-12-30
[docs] Don't actually save unused figures in Literate examples

giordano:
By `save`-ing figures when running Literate examples, we store on the website figures which aren't actually used at all, which becomes a waste of space in the long run (see [Oceananigans experience](https://github.com/CliMA/Oceananigans.jl/issues/5027)).  Inspired by https://github.com/NumericalEarth/Breeze.jl/pull/334#discussion_r2652766036.

### DIFF
--- a/examples/bomex.jl
+++ b/examples/bomex.jl
@@ -359,7 +359,7 @@ text!(axuv, -8.5, 2200, text="solid: u\ndashed: v", fontsize=12)
 
 fig[0, :] = Label(fig, "BOMEX: Mean profile evolution (Siebesma et al., 2003)", fontsize=18, tellwidth=false)
 
-save("bomex_profiles.png", fig)
+save("bomex_profiles.png", fig) #src
 fig
 
 # The simulation shows the development of a cloudy boundary layer with:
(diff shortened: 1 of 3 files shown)


giordano:
Compare:

* current dev: https://github.com/NumericalEarth/BreezeDocumentation/tree/373ff1f7ba435ca3989fee733aabf3192a9a3519/dev/literated
* this PR: https://github.com/NumericalEarth/BreezeDocumentation/tree/373ff1f7ba435ca3989fee733aabf3192a9a3519/previews/PR361/literated

Note how in this PR the files `literated/bomex_profiles.png` and `literated/rico_profiles.png` are not present, as desired.

glwagner (approved):
thanks @giordano 

if we paste a link to the figures directly, then this does not constitute a waste -- is that correct? eg it is only a waste if we generate figures but do not embed the saved files (instead embedding using `display(fig)` or whatever)

I ask because it is sometimes convenient (eg to reduce filesize) to use bespoke display options for a single figure. In that case, it can be nice to use `save` rather than changing a global figure config just for one figure (perhaps there is another way to do this?)

giordano:
> if we paste a link to the figures directly, then this does not constitute a waste -- is that correct? eg it is only a waste if we generate figures but do not embed the saved files (instead embedding using `display(fig)` or whatever)

Yes, if you do use the files saved on disk in the docs then it's not a waste, but in all cases here the files are saved but never referenced in the docs (instead, the `Figure` object is always `display`ed, but that's rendered separately from the file saved to disk).  And then we have multiplies copies of these files, one for each version of the docs, and these things add up over time.

PR #367, opened 2025-12-31
Implement AllSkyOptics for RadiativeTransferModel

glwagner:
This PR extends radiation support to "all sky" optics. "All sky" is meant in opposition to "clear sky"; ie "all sky" includes the effects of clouds. It does not include aerosols, however.

I believe we can get away with a non-random / non-monte-carlo scheme because we are focused on LES. But we can evaluate this.

### DIFF
--- a/docs/src/api.md
+++ b/docs/src/api.md
@@ -77,6 +77,13 @@ Modules = [Breeze.TurbulenceClosures]
 Private = false
 ```
 
+### TimeSteppers
+
+```@autodocs
+Modules = [Breeze.TimeSteppers]
+Private = false
+```
+
 ### Forcings
 
 ```@autodocs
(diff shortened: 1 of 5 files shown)


src/AtmosphereModels/radiation_interface.jl
```diff
@@ -92,7 +104,7 @@ RadiativeTransferModel
 ```
 """
 function RadiativeTransferModel(grid::AbstractGrid, optics, args...; kw...)
-    msg = "Unknown optics $(optics). Valid options are GrayOptics(), ClearSkyOptics().\n" *
+    msg = "Unknown optics $(optics). Valid options are GrayOptics(), ClearSkyOptics(), AllSkyOptics().\n" *
```
giordano: It's slightly unfortunate we need to list the types explicitly here, but the only solution I see is using `InteractiveUtils.subtypes()`, but I'd rather avoid adding a dependency on `InteractiveUtils` if possible (although I see `GPUCompiler`, `KernelAbstractions`, and `Oceananigans` already do it, so it wouldn't be too bad after all)

PR #368, opened 2025-12-31
Add capability to set relative humidity directly

### LINKED ISSUE #282: Support setting `RelativeHumidity` in `set!`
It seems it will be convenient to develop a way to set the relative humidity in `set!`, eg something like

```julia
set!(model, θ=θᵢ, ℋ=ℋᵢ)
```

where `ℋ` is relative humidity (notation up for debate too, this is what Emmanuel 1994 uses).

this would be an alternative to setting the total water.

glwagner:
Closes #282

### DIFF
--- a/src/Microphysics/microphysics_diagnostics.jl
+++ b/src/Microphysics/microphysics_diagnostics.jl
@@ -1,15 +1,15 @@
 using Adapt: Adapt, adapt
+
 using Oceananigans.AbstractOperations: KernelFunctionOperation
 using Oceananigans.Fields: Field, Center
+using Oceananigans.Utils: Utils
 
 using Breeze.Thermodynamics:
     saturation_specific_humidity,
-    dry_air_gas_constant,
     vapor_gas_constant,
     density,
     saturation_vapor_pressure,
     equilibrated_surface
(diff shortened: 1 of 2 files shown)


glwagner:
This PR prompted a bit of a refactor. I moved the concept of "equilibrium" to Thermodynamics, and moved the SaturationSpecificHumidity diagnostic into AtmosphereModels. This allows is to be used during `set!`. I think it reflects a better division of concerns.

giordano:
Doctest failure is due to a change of results, is that expected?

glwagner:
> Doctest failure is due to a change of results, is that expected?

I'm not sure, we should def take a close look before merging

glwagner:
> > It's 1.11
> 
> No, you were right, we require v1.10:
> 
> https://github.com/NumericalEarth/Breeze.jl/blob/49b544a7dcff7e7a8b91bf08272563bd6c065c54/Project.toml#L38
> 
> I thought we were requiring v1.11 already, I wonder if that'd be any faster than v1.10 🤔

I finally understand that "min" means "minimum Julia version allowed by compat". I'm learning!

PR #373, opened 2025-12-31
Support 2M microphysics from CloudMicrophysics

glwagner:
This PR builds out support for two-moment microphysics from CloudMicrophysics. The "two-moment" scheme evolves both the mass fraction _and_ number density of both cloud liquid and rain particles. CloudMicrophysics' two-moment scheme only supports warm processes. To support ice processes, we need the P3 scheme (which uses the two-moment scheme for warm processes). So, this PR is a precursor to supporting the P3 scheme.

Note, we still haven't validated non-equilibrium single moment physics for e.g. RICO.

### DIFF
--- a/test/cloud_microphysics_2M.jl
+++ b/test/cloud_microphysics_2M.jl
@@ -18,7 +18,7 @@ using Oceananigans.BoundaryConditions: ImpenetrableBoundaryCondition
 ##### Two-moment microphysics tests
 #####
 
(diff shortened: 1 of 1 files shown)


glwagner:
well, tests pass, so i dont think we need to wait. #368 will still improve the situation.

giordano:
Yeah, it depends on the scheduling of the tests.

PR #379, opened 2026-01-04
Increase relative humidity in cloud kelvin helmholtz

glwagner:
On `main` there is no instability (because the relative humidity peak is 1.0); this PR tries to fix the example so it goes unstable again within the duration of the simulation.


glwagner:
Why does it seem like the final visualization and last frame of the movie are different?

navidcy (approved):

glwagner:
well, now when I look at the preview I indeed seem to see the correct movie. Not sure what was happening before, but probably I am just going insane as the most likely possibility

PR #382, opened 2026-01-04
[CI] Run tests with Julia v1.11

giordano:
Ref: https://github.com/NumericalEarth/Breeze.jl/pull/368#issuecomment-3704130094.  For the time being this is only an experiment, to see how long the CI jobs take to run.


giordano:
Not any improvement compared to v1.10.

PR #383, opened 2026-01-05
Precipitation flux conserves total water but modifies liquid–ice potential temperature

### LINKED ISSUE #364: Microphysics sedimentation does not change total water or liquid-ice potential temperature
In the current 1M microphysics, “sedimentation” is modeled by advecting the rain tracer ρqʳ with an added downward microphysical velocity wʳ (terminal velocity).

When rain sediments out of the bottom boundary (open BC), only ρqʳ changes. The thermodynamic and moisture prognostics remain unaffected because 1M does not implement microphysical tendencies for Val(:ρqᵗ) or Val(:ρθ) (defaults to zero).

kaiyuan-cheng:
close #364


test/sedimentation.jl
```diff
@@ -0,0 +1,180 @@
...
+    grid = model.grid
+    ρᵣ = model.dynamics.reference_state.density
+    qᵗ = model.specific_moisture
+    
+    Nx, Ny, Nz = size(grid)
+    Δx, Δy, Δz = grid.Lx / Nx, grid.Ly / Ny, grid.Lz / Nz
+    
+    total = zero(eltype(grid))
+    for k in 1:Nz, j in 1:Ny, i in 1:Nx
+        total += @allowscalar(ρᵣ[i, j, k] * qᵗ[i, j, k]) * Δx * Δy * Δz
```
glwagner: I think this whole function requires just one line:

```julia
total = Integral(model.moisture_density) |> Field
```

test/sedimentation.jl
```diff
@@ -0,0 +1,180 @@
...
+    end
+    return total
+end
+
+"""Compute column-integrated potential temperature density: ∫ρθ dV"""
+function column_integrated_rho_theta(model)
```
glwagner: please use `Field(Integral(ρθ))`

test/sedimentation.jl
```diff
@@ -0,0 +1,180 @@
...
+#####
+##### Conservation tests
+#####
+
+@testset "Total water conservation with ImpenetrableBoundaryCondition [$(FT)]" for FT in (Float32, Float64)
+    # With ImpenetrableBoundaryCondition, rain collects at bottom but doesn't leave,
+    # so total water should be conserved.
+    
+    Oceananigans.defaults.FloatType = FT
+    model, _ = setup_test_model(FT; precipitation_boundary_condition=ImpenetrableBoundaryCondition())
```
glwagner: we should rely on the default float type for `setup_test_model`. Just because we have to rely on it elsewhere, so it is pointless to pass around `FT` explicitly. It just clutters the code.

PR #345, opened 2025-12-24
Add bibtex entries + fix make docstring citations REPL-friendly

navidcy:

PR #384, opened 2026-01-06
Add TetensFormula saturation vapor pressure

### LINKED ISSUE #370: Use TabulatedFunction to accelerate evaluation of `saturation_vapor_pressure`
`saturation_vapor_pressure` is expensive, with an `exp` and `pow`:

https://github.com/NumericalEarth/Breeze.jl/blob/cd1c7752fa8ce562281747d0072465fe585847a4/src/Thermodynamics/vapor_saturation.jl#L59

This can be accelerated by tabulating its values and then linearly interpolating between them. Oceananigans recently added a `TabulatedFunction` which can do this

```julia
help?> Oceananigans.Utils.TabulatedFunction
  │ Warning
  │
  │  The following bindings may be internal; they may change or be removed in future versions:
  │
  │    •  Oceananigans.Utils

  TabulatedFunction{F, T, FT}

  A wrapper around a unary callable func(x) that precomputes values in a lookup table for fast linear interpolation. This avoids expensive
  computations (like log, exp, sqrt, atan, etc.) during time-critical operations like GPU kernel execution.

  Fields
  ≡≡≡≡≡≡

    •  func: The original callable being tabulated (for reference/fallback)
    •  table: Precomputed values (AbstractVector)
    •  x_min: Minimum x value in table
    •  x_max: Maximum x value in table
    •  inverse_Δx: 1 / Δx for fast index computation

  Example
  ≡≡≡≡≡≡≡

  using Oceananigans.Utils: TabulatedFunction

  # Tabulate a simple function for fast evaluation
  f = TabulatedFunction(sin; range=(0, 2π))

  # output
  TabulatedFunction with 100 points over [0.0, 6.283185307179586] of sin

  # Evaluate like a regular function
  f(π/2)

  # output
  0.9996224305511583

  ────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────

  TabulatedFunction(func, [arch=CPU()], [FT=Float64]; range, points=100)

  Construct a TabulatedFunction by precomputing points values of func over range for fast linear interpolation.

  Arguments
  ≡≡≡≡≡≡≡≡≡

    •  func: Any callable that takes a single numeric argument
    •  arch: Architecture for the lookup table (CPU() or GPU()). Default: CPU()
    •  FT: Float type for table values. Default: Float64

  Keyword Arguments
  ≡≡≡≡≡≡≡≡≡≡≡≡≡≡≡≡≡

    •  range: Tuple of (minimum, maximum) x values (required)
    •  points: Number of points in the lookup table. Default: 100

  Example
  ≡≡≡≡≡≡≡

  using Oceananigans.Utils: TabulatedFunction

  # Tabulate a trigonometric function
  f = TabulatedFunction(sin; range=(0, 2π), points=1000)

  # Evaluate at π/4
  f(π/4)

  # output
  0.7071052539107768

  The tabulated function can be called like the original:

  f(1.5)  # Returns interpolated value

  Values outside range are clamped to the nearest table boundary.
```

to implement this I think we need to add a `saturation_vapor_pressure` property to `ThermodynamicConstants` and then define:

```julia
saturation_vapor_pressure(T, constants::ThermoConstantsWithTabulation, ::PlanarLiquidSurface) =
    constants.saturation_vapor_pressure.liquid(T)
```

unfortunately each surface needed will need its own table. So it is probably best to do this as a "materialization" step in the model constructor, since at that time we know which surfaces will be needed (liquid, ice, and mixed-phase surfaces).

glwagner:
This PR adds a property `saturation_vapor_pressure` to `ThermodynamicsConstants`. The default will be `ClausiusClapeyron`. We also implement a `TetensFormula` options, which is a simplified form for the saturation vapor pressure used by some models. We can also utilize this property to implement TabulatedFunction saturated vapor pressure and resolve #370 (but in another PR)

with @kaiyuan-cheng 

TODO:
- is there a better name for the property than `saturation_vapor_pressure`? Perhaps `saturation_formulation`?

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -43,7 +43,7 @@ Breeze interfaces with ClimaOcean for coupled atmosphere-ocean simulations.
    - Use `$(TYPEDSIGNATURES)` for automatic typed signature documentation (preferred over `$(SIGNATURES)`)
    - Never write explicit function signatures in docstrings; always use `$(TYPEDSIGNATURES)`
    - Add examples in docstrings when helpful
-   - **Citations in docstrings**: Use inline citations with `[Key](@citet)` syntax (renders as "Author (Year)" with link).
+   - **Citations in docstrings**: Use inline citations with `[Author1 and Author2 (year)](@cite Key)` or `[Author1 et al. (Year)](@cite Key)` syntax.
      Avoid separate "References" sections with bare `[Key](@cite)` - these just show citation keys in the REPL
      without context, which is not helpful. Instead, weave citations naturally into the prose, e.g.:
      "Tetens' formula [Tetens1930](@citet) is an empirical formula..."


src/Thermodynamics/tetens_formula_saturation_vapor_pressure.jl
```diff
@@ -0,0 +1,116 @@
...
+The Tetens formula approximates saturation vapor pressure as:
+
+```math
+pᵛ⁺(T) = pᵛ⁺₀ \\exp \\left( a \\frac{T - T₀}{T - b} \\right)
+```
+
+where ``pᵛ⁺₀`` is the reference saturation vapor pressure, ``a`` is an empirical coefficient,
+``T₀`` is the reference temperature, and ``b`` is a temperature offset.
+
+Default parameter values are derived from equation 2.11 in [Klemp1978](@cite):
```
giordano: You need to add `Klemp1978` to the bibliography, too. Also, it'd be better to use a more REPL-friendly style for docstring, see #345

navidcy:
```suggestion
Default parameter values are derived from equation 2.11 in [Klemp and Wilhelmson (1978)](@cite Klemp1978):
```

navidcy (approved):
Should we add a plot with this approximation in the Docs/Thermodynamics section so that we see how it compares to the Clausius-Clapeyron?

glwagner:
> Should we add a plot with this approximation in the Docs/Thermodynamics section so that we see how it compares to the Clausius-Clapeyron?

I like that, then we can also expand on it to show the accuracy of the TabulatedFunction versions

src/Thermodynamics/tetens_formula_saturation_vapor_pressure.jl
glwagner: is this comment resolved?

glwagner: @navidcy

navidcy: Yes!

docs/src/breeze.bib
```diff
@@ -270,3 +270,43 @@ @article{SeifertBeheng2006
...
+}
+
+@book{MonteithUnsworth2008,
+  author = {Monteith, John L. and Unsworth, Mike H.},
+  title = {Principles of Environmental Physics},
+  publisher = {Academic Press},
+  edition = {3rd},
+  year = {2008},
+  isbn = {978-0-12-505103-3}
+}
```
giordano:
```suggestion
@book{MonteithUnsworth2014,
  author = {Monteith, John L. and Unsworth, Mike H.},
  title = {Principles of Environmental Physics},
  publisher = {Academic Press},
  edition = {4th},
  year = {2014},
  isbn = {978-0-12-386910-4},
  doi = {10.1016/C2010-0-66393-0},
}
```

navidcy:
I think now all looks OK!

src/Thermodynamics/tetens_formula.jl
```diff
@@ -0,0 +1,154 @@
...
+**Ice** (T < 0°C):
+- `ice_coefficient`: 21.875
+- `ice_temperature_offset`: 7.65 K (corresponding to 265.5 K offset from 0°C)
+
+# References
+
+* Tetens, O. (1930). Über einige meteorologische Begriffe. Zeitschrift für Geophysik 6, 297–309.
+* Monteith, J. L. and Unsworth, M. H. (2014). Principles of Environmental Physics. 4th Edition (Academic Press).
+* Murray, F. W. (1967). On the computation of saturation vapor pressure. Journal of Applied Meteorology 6, 203–204.
+* Wikipedia: Tetens equation; [https://en.wikipedia.org/wiki/Tetens_equation](https://en.wikipedia.org/wiki/Tetens_equation)
```
giordano:
```suggestion
* Wikipedia: Tetens equation; <https://en.wikipedia.org/wiki/Tetens_equation>
```
doesn't require duplicating the URL

AGENTS.md
```diff
@@ -43,6 +43,10 @@ Breeze interfaces with ClimaOcean for coupled atmosphere-ocean simulations.
    - Use `$(TYPEDSIGNATURES)` for automatic typed signature documentation (preferred over `$(SIGNATURES)`)
    - Never write explicit function signatures in docstrings; always use `$(TYPEDSIGNATURES)`
    - Add examples in docstrings when helpful
+   - **Citations in docstrings**: Use inline citations with `[Key](@citet)` syntax (renders as "Author (Year)" with link).
```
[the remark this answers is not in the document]
navidcy:
```suggestion
   - **Citations in docstrings**: Use inline citations with `[Author1 and Author2 (year)](@cite Key)` or `[Author1 et al. (Year)](@cite Key)` syntax.
```

kaiyuan-cheng (approved):
Looks good. Thanks!

PR #388, opened 2026-01-06
Add bibtex entry and fixes some docs refs

navidcy:


src/AtmosphereModels/atmosphere_model.jl
```diff
@@ -1,3 +1,4 @@
+using ..AnelasticEquations: AnelasticEquations
```
giordano:
```
ERROR: LoadError: UndefVarError: `AnelasticEquations` not defined in `Breeze`
```
Probably a bootstrap problem, `AnelasticEquations` is not yet defined at this point. Not sure this is worth it just to slightly simplify a docstring.

glwagner (approved):

PR #396, opened 2026-01-08
Revisit additional tests for set!

glwagner:
Revamp of #163


glwagner:
These tests appear to be unnecessary since they don't change code coverage and pass with the source as-is, and cause the tests to be more expensive. So I am closing this PR.

PR #398, opened 2026-01-09
Add a supercell case for model validation

kaiyuan-cheng:
This PR adds a splitting supercell case following the [DCMIP2016 protocol](https://doi.org/10.5194/gmd-12-879-2019). The case provides a standardized, well-documented benchmark for validating dynamical core behavior in a deep, moist convective regime.

### DIFF
--- a/examples/splitting_supercell.jl
+++ b/examples/splitting_supercell.jl
@@ -294,6 +294,7 @@ slice_outputs = (
 
 slices_filename = "splitting_supercell_slices.jld2"
 simulation.output_writers[:slices] = JLD2Writer(model, slice_outputs; filename=slices_filename,
+                                                including = [:grid],
                                                 schedule = TimeInterval(2minutes),
                                                 overwrite_existing = true)
 


giordano:
It's not pending, it's a github UI glitch: it's a "required" check, but this PR only touches documentation and we don't run the tests jobs in that case to save resources, so they will never run anyway. The github UI shows the check as "pending" because they don't seem to have a concept of a job which is "required to pass when it runs, but could be skipped", for them "required" is _**always**_ required to even run.

giordano:
If this is good to go (it is from my point of view, I don't know if @glwagner has more comments), we can bypass the required check
image

kaiyuan-cheng:
> If this is good to go (it is from my point of view, I don't know if @glwagner has more comments), we can bypass the required check image

Let’s wait to see if @glwagner has any additional comments.

docs/make.jl
```diff
@@ -38,6 +38,7 @@ examples = [
     Example("Single column radiation", "single_column_radiation", true),
     Example("Stationary parcel model", "stationary_parcel_model", true),
     Example("Acoustic wave in shear layer", "acoustic_wave", true),
+    Example("Splitting supercell", "splitting_supercell", true),
```
glwagner:
```suggestion
    Example("Splitting supercell", "splitting_supercell", false),
```

because this seems like a fairly expensive example, let's run it only when we merge to `main` and not on every commit

examples/splitting_supercell.jl
```diff
@@ -0,0 +1,372 @@
...
+# Relative humidity profile (decreases with height, 25% above tropopause):
+
+ℋ_background(z) = (1 - 3/4 * (z / zₜᵣ)^(5/4)) * (z <= zₜᵣ) + 1/4 * (z > zₜᵣ)
+
+# Zonal wind profile with linear shear below ``z_s`` and smooth transition (Equations 15-16):
+
+function u_background(z)
+    u_lower = uₛ * (z / zₛ) - uᶜ
+    u_transition = (-4/5 + 3 * (z / zₛ) - 5/4 * (z / zₛ)^2) * uₛ - uᶜ
+    u_upper = uₛ - uᶜ
```
glwagner: recommend superscripts not subscripts for "s". But also what does "s" mean? "shear"?

examples/splitting_supercell.jl
```diff
@@ -0,0 +1,372 @@
...
+θₜᵣ = 343       # K - tropopause potential temperature
+zₜᵣ = 12000     # m - tropopause height
+Tₜᵣ = 213       # K - tropopause temperature
+nothing #hide
+
+# Wind shear parameters control the low-level environmental wind profile:
+
+zₛ = 5kilometers  # m - shear layer height
+uₛ = 30           # m/s - maximum shear wind speed
+uᶜ = 15           # m/s - storm motion (Galilean translation speed)
```
glwagner: probably better to use superscripts throughout. the subscript _tr looks weird to me:
Screenshot 2026-01-19 at 7 52 04 PM

examples/splitting_supercell.jl
glwagner: Here's another suggestion:

^t: troposphere
^p: tropopause
^u: upper

use this notation uniformly throughout

examples/splitting_supercell.jl
```diff
@@ -0,0 +1,372 @@
...
+    @info msg
+
+    return nothing
+end
+
+add_callback!(simulation, progress, IterationInterval(100))
+
+# Collect maximum vertical velocity time series during simulation:
+
+max_w_timeseries = []
```
```suggestion
max_w_ts = []
```

only beccause below, "ts" is used to represent "timeseries"

glwagner:
unfortunately the preview has expired (and there are conflicts) --- once these are fixed, i would love to see the preview

glwagner (approved):
I made only minor notational comments. Once these are addressed I'm happy to see this merged.

PR #394, opened 2026-01-08
Implement a KinematicModel based on PrescribedVelocityFields

glwagner:
This "kinematic driver" can be used to test microphysics schemes.

PR #403, opened 2026-01-14
Bump to 0.3.0

glwagner:
It hasn't been bumped in a while, now we have the kinematic driver example, maybe others?


giordano:
#356 was a breaking change, and I think also #394 is (it changed the API of the interface), we may have had other breaking changes too.

PR #405, opened 2026-01-14
Refactor microphysics interface and implement ParcelDynamics

### LINKED ISSUE #180: Add `ParcelModel` mode
Like `SingleColumnModel` mode for grid with `topology = (Flat, Flat, Bounded)`, it will also be nice to have a "parcel model mode" for grids with `topology = (Flat, Flat, Flat)`. This would be a 0D simulation of only microphysics. Somehow in this model we will want to be able to change the environmental conditions somehow, which will require a bit of thinking about what the inputs and outputs of the model are.

glwagner:
This PR refactors the internal interface for implementing microphysics schemes. The thrust of this effort is to make the implementation of microphysics schemes as "gridless" as possible. That is to say, we want to reduce the number of functions that must be defined which depend on `i, j, k, grid`.

This initiative is helped by the fact that we already have gridless thermodynamic states --- LiquidIcePotentialTemperatureState and StaticEnergyState. We continue this concept and ask microphysics schemes to implement their own auxiliary state struct. Together with the prognostic state (implied by the prognostic_names, so we don't need special type for it), we are able to achieve a gridless abstraction for the microphysical and thermodynamic state, together.

A gridless microphysics interface is necessary for supporting a paradigm wherein microphysics schemes can be tested / validated / prototyped with a "parcel model", and then immediately used in a 3D simulation. The parcel model is Lagrangian in nature, and therefore the microphysical state in a parcel evolves independently of any grid.

This "dream interface" is not yet complete, mostly because of the difficulty dealing with boundary conditions on gridded microphysical velocities. I think this points towards the need for a well-designed interface for specifying the microphysical velocities, too. But completing this task should probably wait for a future PR.

This PR also implements ParcelDynamics, which produces a ParcelModel version of AtmosphereModel. I am unsure whether this is a good idea, or whether parcel dynamics should be incorporated into a Lagrangian particles feature. I am leaning towards the latter, but would like to keep around the work that's been done so far on this PR as a stopgap, if others accept this.

Here are some reference repos for parcel and/or particle models:

- https://github.com/open-atmos/PySDM
- https://github.com/darothen/pyrcel

Resolves #180

### DIFF
--- a/test/dcmip2016_kessler.jl
+++ b/test/dcmip2016_kessler.jl
@@ -321,7 +321,7 @@ end
             reference_latent_heat = 2834000.0,
             heat_capacity = cᵖ))
 
-    microphysics = DCMIP2016KesslerMicrophysics{FT}()
+    microphysics = DCMIP2016KesslerMicrophysics(FT)
 
     # Convert to mass fractions
     rᵗ_init = rᵛ_init .+ rᶜˡ_init .+ rʳ_init


src/ParcelDynamics/parcel_dynamics.jl
```diff
@@ -0,0 +1,399 @@
...
+    u::Any
+
+    "Meridional velocity profile: v(z) [m/s]"
+    v::Any
+
+    "Vertical velocity profile: w(z) [m/s]"
+    w::Any
+
+    "Current parcel state"
+    state::Any
```
giordano: Should all these fields really be `Any`?  Not `FT`?

AGENTS.md
```diff
@@ -455,6 +464,41 @@ Before running a long simulation:
...
+
+```bash
+# Remove trailing whitespace from all .jl, .md, .sh files
+find /path/to/Breeze -type f \( -name "*.jl" -o -name "*.md" -o -name "*.sh" \) \
+  -exec sed -i '' 's/[[:space:]]*$//' {} \;
+
+# Ensure files end with exactly one newline (macOS)
+find /path/to/Breeze -type f \( -name "*.jl" -o -name "*.md" -o -name "*.sh" \) \
+  -exec sh -c 'printf "%s\n" "$(cat "$1")" > "$1"' _ {} \;
+```
```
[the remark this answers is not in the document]
giordano: Yay!

README.md
```diff
@@ -36,7 +36,11 @@
...
+Breeze's AtmosphereModel currently supports anelastic dynamics, compressible dynamics, a kinematic driver, and a parcel model, all with strong-stability-preserving (SSP) RK3 time stepping.
+The anelastic and compressible dynamics can be used with a liquid-ice potential temperature formulation or static energy formulation and a variety of advection schemes including high-order and bounds-preserving WENO.
+Breeze provides abstractions for Coriolis forces, simple bulk drags, heat and moisture fluxes, and closures for large-eddy simulation.
+A main focus of Breeze development is to facilitate the prototyping and implementation of microphysics schemes through streamlined internal interfaces and automatic kinematic driver and parcel model capabilities.
```
kaiyuan-cheng:
```suggestion
A main focus of Breeze development is to facilitate the prototyping and implementation of physics schemes through streamlined internal interfaces and automatic kinematic driver and parcel model capabilities.
```

kaiyuan-cheng: I think this approach would work for other types of schemes too.

giordano:
Example finally built successful: https://numericalearth.github.io/BreezeDocumentation/previews/PR405/literated/stationary_parcel_model/

giordano:
Yes, definitely wrong link 😅

examples/rising_parcels.jl
```diff
@@ -0,0 +1,262 @@
...
+cloudy_T = [temperature(s.𝒰, cloudy_constants) for s in cloudy_snapshots]
+cloudy_qᵛ = [s.𝒰.moisture_mass_fractions.vapor for s in cloudy_snapshots]
+cloudy_qᶜˡ = [s.μ.ρqᶜˡ / s.ρ for s in cloudy_snapshots]
+cloudy_qʳ = [s.μ.ρqʳ / s.ρ for s in cloudy_snapshots]
+cloudy_nᶜˡ = [s.μ.ρnᶜˡ / s.ρ for s in cloudy_snapshots]
+cloudy_nʳ = [s.μ.ρnʳ / s.ρ for s in cloudy_snapshots]
+cloudy_S = [supersaturation(temperature(s.𝒰, cloudy_constants), s.ρ,
+                            s.𝒰.moisture_mass_fractions, cloudy_constants,
+                            PlanarLiquidSurface()) for s in cloudy_snapshots]
+nothing
```
giordano:
```suggestion
nothing #hide
```

examples/rising_parcels.jl
```diff
@@ -0,0 +1,262 @@
...
+@info "Dry parcel reached" model.dynamics.state.z
+
+# Extract time series from snapshots
+constants = model.thermodynamic_constants
+dry_t = [s.t for s in dry_snapshots]
+dry_z = [s.z for s in dry_snapshots]
+dry_T = [temperature(s.𝒰, constants) for s in dry_snapshots]
+dry_S = [supersaturation(temperature(s.𝒰, constants), s.ρ, s.𝒰.moisture_mass_fractions,
+                         constants, PlanarLiquidSurface()) for s in dry_snapshots]
+nothing
```
giordano:
```suggestion
nothing #hide
```

kaiyuan-cheng:
The parcel is now rising correctly. Great job! However, the result of the two-moment microphysics looks a bit strange:

At the beginning of the simulation, there is a non-zero number concentration for cloud water, even though cloud water has not formed yet. Given the decreasing trend at the beginning, I think the cloud water number concentration should correspond to the aerosol number concentration.

image

glwagner:
the stationary parcel example also uses 2M, so that needs to be changed as well? can you take a look @kaiyuan-cheng ?

glwagner:
cc @haakon-e

kaiyuan-cheng (approved):

src/AtmosphereModels/formulation_interface.jl
```diff
@@ -97,7 +97,7 @@ Diagnose the thermodynamic state at grid point `(i, j, k)` from the given `formu
 `dynamics`, and pre-computed moisture mass fractions `q`.
 
 Note: This function does NOT compute moisture fractions internally to avoid circular dependencies.
-The caller is responsible for computing `q = compute_moisture_fractions(...)` before calling.
+The caller is responsible for computing `q = moisture_fractions(...)` before calling.
```
giordano: Is this `moisture_fractions` or `grid_moisture_fractions`?  Looking at https://github.com/NumericalEarth/Breeze.jl/blob/e6f113c85cf139e06084b7be66cd51a91bc3e917/src/StaticEnergyFormulations/static_energy_tendency.jl#L61-L62 and https://github.com/NumericalEarth/Breeze.jl/blob/e6f113c85cf139e06084b7be66cd51a91bc3e917/src/AtmosphereModels/dynamics_kernel_functions.jl#L138-L140 https://github.com/NumericalEarth/Breeze.jl/blob/e6f113c85cf139e06084b7be66cd51a91bc3e917/src/AtmosphereModels/update_atmosphere_model_state.jl#L242-L245 etc I think it's the latter?
```suggestion
The caller is responsible for computing [`q = grid_moisture_fractions(...)`](@ref grid_moisture_fractions) before passing `q` to this function.
```

src/ParcelModels/parcel_dynamics.jl
```diff
@@ -0,0 +1,880 @@
...
+Base.summary(::ParcelDynamics) = "ParcelDynamics"
+
+function Base.show(io::IO, d::ParcelDynamics)
+    print(io, "ParcelDynamics\n")
+    state_str = d.state isa ParcelState ? d.state : "uninitialized"
+    print(io, "├── state: ", state_str, '\n')
+    print(io, "├── timestepper: ", isnothing(d.timestepper) ? "uninitialized" : "ParcelTimestepper (SSP RK3)", '\n')
+    print(io, "├── density: ", isnothing(d.density) ? "unset" : summary(d.density), '\n')
+    print(io, "├── pressure: ", isnothing(d.pressure) ? "unset" : summary(d.pressure), '\n')
+    print(io, "├── surface_pressure: ", d.surface_pressure, '\n')
```
giordano:
```suggestion
    println(io, "ParcelDynamics")
    state_str = d.state isa ParcelState ? d.state : "uninitialized"
    println(io, "├── state: ", state_str)
    println(io, "├── timestepper: ", isnothing(d.timestepper) ? "uninitialized" : "ParcelTimestepper (SSP RK3)")
    println(io, "├── density: ", isnothing(d.density) ? "unset" : summary(d.density))
    println(io, "├── pressure: ", isnothing(d.pressure) ? "unset" : summary(d.pressure))
    println(io, "├── surface_pressure: ", d.surface_pressure)
```

test/parcel_dynamics.jl
```diff
@@ -0,0 +1,651 @@
...
+
+    # In an isentropic atmosphere, parcel temperature should match environment
+    # Allow 1 K tolerance for numerical errors
+    @test abs(T_parcel - T_environment) < 1.0
+
+    # Specific static energy should be EXACTLY conserved (specific quantity evolution)
+    @test e_final == e_initial
+
+    # Parcel should have risen to expected height
+    @test z_final ≈ 1200.0 atol=1.0
```
giordano: Ok, I can tolerate absolute tolerance here where the reference number is a literal, so at least one can easily tell what's the relative error 🥲

glwagner: hhahaaa

test/parcel_dynamics.jl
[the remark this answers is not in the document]
giordano: Ok, should we just skip this when running on GPU then, to save resources? I've been thinking we could do the same also for the doctests, they don't actually run on GPU.

giordano:
I'm doing some more investigation in #421, there are genuine NaNs, but tracing them down may take a bit.

giordano:
> Arguably we actually do want this kind of thing to fail 😂 since obviously we do not intend to use F64 microphysics parameters with a code that is otherwise F32.

I agree on this, but I think it'd still be worth understanding _why_ we're getting NaN

glwagner:
damn you Julia min

PR #395, opened 2026-01-08
Implement PredictedParticlePropertyMicrophysics

glwagner:
This PR takes a stab at implementing the most recent version of the "Predicted Particle Property" (P3) Microphysics in Breeze as contained in https://github.com/P3-microphysics/P3-microphysics and described by https://agupubs.onlinelibrary.wiley.com/doi/full/10.1029/2024MS004404.

Some references for P3 (please let me know if there are others I should add):

* [Morrison and Milbrandt 2015](https://journals.ametsoc.org/view/journals/atsc/72/1/jas-d-14-0065.1.xml)
* [Morrison et al 2015](https://journals.ametsoc.org/view/journals/atsc/72/1/jas-d-14-0066.1.xml)
* [Milbrandt and Morrison 2016](https://journals.ametsoc.org/view/journals/atsc/73/3/jas-d-15-0204.1.xml)
* [Cholette et al 2019](https://journals.ametsoc.org/view/journals/atsc/76/2/jas-d-18-0278.1.xml)
* [Paukert et al 2019](https://agupubs.onlinelibrary.wiley.com/doi/full/10.1029/2018MS001512)
* [Jouan et al 2020](https://journals.ametsoc.org/view/journals/wefo/35/6/waf-d-20-0111.1.xml)
* [Milbrandt et al 2021](https://journals.ametsoc.org/view/journals/atsc/78/2/jas-d-20-0084.1.xml)
* [Cholette et al 2023](https://agupubs.onlinelibrary.wiley.com/doi/full/10.1029/2022MS003328)
* [Morrison, Milbrandt, Cholette 2025](https://agupubs.onlinelibrary.wiley.com/doi/full/10.1029/2024MS004644)
* [Milbrandt, Morrison, Cholette 2025](https://agupubs.onlinelibrary.wiley.com/doi/full/10.1029/2024MS004404)

github repo: https://github.com/P3-microphysics/P3-microphysics, which includes interfaces for [CM1](https://github.com/NCAR/CM1)

PR #133, opened 2025-11-10
[CI] Add job for suggesting formatting with Runic

giordano:
This is liberally <s>stolen from</s> inspired by https://github.com/EnzymeAD/Enzyme.jl/blob/986bbf45b7d56d9761d2868e6c4e9b3d3c546cdd/.github/workflows/Format.yml.  It's only for posting a comment to suggest changes, it doesn't fail a check if formatting isn't perfect.

Ref: #92.

PR #406, opened 2026-01-14
Cleanup whitespaces in Breeze codebase

giordano:
This also adds a simple check for simple whitespace issues which is used in the Julia repo.  In PRs, this annotates the offending lines in the "files changed" tab, see an example at https://github.com/JuliaLang/julia/pull/55281#issue-2433901534.

A [previous attempt](https://github.com/NumericalEarth/Breeze.jl/pull/133) at adding a formatting linter didn't go very far, but this is very lightweight and only about extra, or missing, whitespaces at the end of lines and similar issues, and it's even used in the Julia repo where there's traditionally some resistance to enforce heavy handed code formatters.  Note: #395 would introduce 400+ more whitespace issues (which is my main motivation for this 🙃), but I can fix them all in one go with a simple elisp script if this is accepted.

### DIFF
--- a/.github/workflows/Whitespace.yml
+++ b/.github/workflows/Whitespace.yml
@@ -1,0 +1,36 @@
+name: Whitespace
+
+permissions: {}
+
+on:
+  push:
+    branches:
+      - 'main'
+  pull_request:
+
+jobs:
+  whitespace:
+    name: Check whitespace
+    runs-on: ubuntu-latest
+    timeout-minutes: 2
+    steps:
+      - name: Checkout the NumericalEarth/Breeze.jl repository
+        uses: actions/checkout@v6
+        with:
+          persist-credentials: false
+      - name: Checkout the JuliaLang/julia repository
+        uses: actions/checkout@v6
+        with:
+          persist-credentials: false
+          repository: 'JuliaLang/julia'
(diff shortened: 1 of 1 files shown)


examples/rico.jl
```diff
@@ -148,7 +148,7 @@ set!(∂t_ρqᵗ_large_scale, ρᵣ * ∂t_ρqᵗ_large_scale)
 # This is the key simplification that allows us to avoid interactive radiation.
 
 ∂t_ρθ_large_scale = Field{Nothing, Nothing, Center}(grid)
-∂t_θ_large_scale = - 2.5 / day # K / day
+∂t_θ_large_scale = - 2.5 / day # K / day
```
giordano: This replaced
```julia-repl
julia> ' '
' ': Unicode U+00A0 (category Zs: Separator, space)
```
with
```julia-repl
julia> ' '
' ': ASCII/Unicode U+0020 (category Zs: Separator, space)
```
No idea how a [non-breaking space](https://en.wikipedia.org/wiki/Non-breaking_space) ended up there in the first place.

.github/workflows/Whitespace.yml
navidcy: I like this!

PR #409, opened 2026-01-16
Fix xlims and xticks on lineplots in acoustic wave example

navidcy:


examples/acoustic_wave.jl
```diff
@@ -154,8 +154,8 @@ fig = Figure(size = (900, 600), fontsize = 12)
 axρ = Axis(fig[1, 2]; aspect = 5, ylabel = "z (m)")
 axw = Axis(fig[2, 2]; aspect = 5, ylabel = "z (m)")
 axu = Axis(fig[3, 2]; aspect = 5, xlabel = "x (m)", ylabel = "z (m)")
-axR = Axis(fig[1, 1]; xlabel = "⟨ρ⟩ (kg/m³)")
-axW = Axis(fig[2, 1]; xlabel = "⟨w²⟩ (m²/s²)")
+axR = Axis(fig[1, 1]; xlabel = "⟨ρ⟩ (kg/m³)", limits = ((1.155, 1.18), nothing))
+axW = Axis(fig[2, 1]; xlabel = "⟨w²⟩ (m²/s²)", limits = ((-0.001, 0.055), nothing))
```
glwagner:
```suggestion
axR = Axis(fig[1, 1]; xlabel = "⟨ρ⟩ (kg/m³)", limits = ((1.16, 1.18), nothing))
axW = Axis(fig[2, 1]; xlabel = "⟨w²⟩ (m²/s²)", limits = ((-0.001, 0.06), nothing))
```

glwagner: I think overly specific plotting is poor practice because it is not transferrable; eg if the example changes, we have to repeat effort to 4 significant digits again.

giordano:
I'm not terribly familiar with Makie API, but I'd assume `limits` affects the range of the plot, not the frequency of the ticks.

navidcy:
> That changed the second plot, but not the first one, right? Ticks for the first plot look to me the same as https://numericalearth.github.io/BreezeDocumentation/dev/literated/acoustic_wave/#Visualization

Now they are the same ticks as in main because I reverted the very specific limits.

glwagner:
probably making the plots a little wider (or the whole plot bigger, keeping fontisze the same) will help with the axes

PR #410, opened 2026-01-16
Convert animation to use countourf in inertia gravity wave example

navidcy:


giordano (approved):
I like it's the same style as the figure above (and the contour plot is also less pixelated than the heatmap)

PR #412, opened 2026-01-17
Resolve issues identified by JETLS

giordano:
I gave [`JETLS`](https://github.com/aviatesk/JETLS.jl) a go.  It didn't find anything terribly interesting, mainly unused arguments, which I prefixed with `_` as convention, and a couple of unused local variables, which I simply deleted.

### DIFF
--- a/src/BoundaryConditions/BoundaryConditions.jl
+++ b/src/BoundaryConditions/BoundaryConditions.jl
@@ -659,7 +659,7 @@ regularize_atmosphere_boundary_condition(bc, side, loc, grid, surface_pressure, 
 function regularize_atmosphere_boundary_condition(bc::BoundaryCondition{<:Flux, <:BulkDragFunction{Nothing}},
                                                   side, loc, grid, surface_pressure, constants)
     df = bc.condition
-    LX, LY, _LZ = loc
+    LX, LY, _ = loc
 
     # Determine direction from location: Face in x means x-momentum, Face in y means y-momentum
     if LX isa Face


src/AtmosphereModels/Diagnostics/potential_temperatures.jl
```diff
@@ -487,21 +487,17 @@ function (d::MoistPotentialTemperatureKernelFunction)(i, j, k, grid)
     # Plain potential temperature (used as a base for several others)
     θ = T / Πᵐ
 
-    if d.flavor isa AbstractPlainFlavor
-        θ★ = θ
-
+    θ★ = if d.flavor isa AbstractPlainFlavor
```
giordano: This was probably the only vaguely interesting case: the variable `θ★` was defined only inside (nested) if conditionals, and inference couldn't be entirely sure it'd be defined at the end.  This resolves the issue, making inference work better (and it's also clearer that the point of the whole block is to define `θ★`)

src/Advection.jl
```diff
@@ -35,8 +35,8 @@ using ..AtmosphereModels: AtmosphereModels, div_ρUc
 end
 
 # Fallback for nothing advection
-@inline AtmosphereModels.div_ρUc(i, j, k, grid, ::Nothing, ρ, U, c) = zero(grid)
-@inline AtmosphereModels.div_ρUc(i, j, k, grid, advection::BoundsPreservingWENO, ρ, U, ::ZeroField) = zero(grid)
+@inline AtmosphereModels.div_ρUc(_i, _j, _k, grid, ::Nothing, _ρ, _U, _c) = zero(grid)
+@inline AtmosphereModels.div_ρUc(_i, _j, _k, grid, _advection::BoundsPreservingWENO, _ρ, _U, ::ZeroField) = zero(grid)
```
glwagner: so `_` means "unused"?

glwagner:
```suggestion
@inline AtmosphereModels.div_ρUc(_i, _j, _k, grid, ::BoundsPreservingWENO, _ρ, _U, ::ZeroField) = zero(grid)
```

glwagner:
I do think it is useful to know what variables are not used in a function, but I think adding `_` everywhere isn't worth that benefit because it harms readability and interpretability. I think there's a balance to be maintained here, so we can discuss.

glwagner:
I think we should discuss this. I think the downsides are genuine. But I understand the upsides.

The fact is that we have a substantial number of interface functions with unused arguments. I don't see an alternative design that would work -- take the convention `i, j, k, grid` which has proved very powerful (but requires unused arguments in a huge number of functions).

The underscore is also an annoying character. It's hard to type and can actually injure people if there are too many that they have to type! We can work around that, but just saying...

@navidcy approved this PR so he must see the benefit over downsides but curious if he wants to chime in.

navidcy:
Well I am a bit on the fence: it's clutter but I also found it useful to know with first glance the unused variables.
It'd be great if they could be color coded without the `_`...

giordano:
> > Well I am a bit on the fence: it's clutter but I also found it useful to know with first glance the unused variables. It'd be great if they could be color coded without the `_`...
> 
> heh, that is an epic idea to add color annotation to characters

That's already the case for [editors connected to JETLS](https://aviatesk.github.io/JETLS.jl/release/#Editor-setup).  This is Emacs, note the greyed out arguments, those are unused:

image

and this is VS Code:

image

Also here the unused arguments are greyed out, and additionally have wiggles underneath, if you hover over them with the mouse it explains what's the issue and suggests actions (e.g. removing them, or adding prefix).

This happens unless you just [disable the unused arguments check entirely](https://github.com/NumericalEarth/Breeze.jl/pull/412#issuecomment-3765248342), as I have done now.

src/BoundaryConditions/BoundaryConditions.jl
```diff
@@ -659,7 +659,7 @@ regularize_atmosphere_boundary_condition(bc, side, loc, grid, surface_pressure,
 function regularize_atmosphere_boundary_condition(bc::BoundaryCondition{<:Flux, <:BulkDragFunction{Nothing}},
                                                   side, loc, grid, surface_pressure, constants)
     df = bc.condition
-    LX, LY, LZ = loc
+    LX, LY, _LZ = loc
```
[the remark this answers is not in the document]
navidcy: I'm happy either way!

navidcy:
@giordano if you merge this tag 0.3.1 as well

PR #418, opened 2026-01-19
Clean up EnergyFluxBoundaryCondition implementation

glwagner:
Following up to #411 which I merged a little prematurely.

### DIFF
--- a/src/BoundaryConditions/BoundaryConditions.jl
+++ b/src/BoundaryConditions/BoundaryConditions.jl
@@ -164,19 +164,15 @@ end
 
 convert_energy_to_theta_bcs(bcs, f::Symbol, c) = convert_energy_to_theta_bcs(bcs, Val(f), c)
 
-# Pass through non-FieldBoundaryConditions
-regularize_atmosphere_field_bcs(fbcs, loc, grid, dynamics, microphysics, surface_pressure, constants) = fbcs
-
 # Regularize FieldBoundaryConditions by walking through each boundary
-function regularize_atmosphere_field_bcs(fbcs::FieldBoundaryConditions, loc, grid, dynamics, microphysics,
-                                         surface_pressure, constants)
-    west     = regularize_atmosphere_boundary_condition(fbcs.west, West(), loc, grid, dynamics, microphysics, surface_pressure, constants)
-    east     = regularize_atmosphere_boundary_condition(fbcs.east, East(), loc, grid, dynamics, microphysics, surface_pressure, constants)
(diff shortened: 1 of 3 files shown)


src/BoundaryConditions/thermodynamic_variable_bcs.jl
```diff
@@ -0,0 +1,320 @@
...
+    EnergyFluxBoundaryCondition(flux)
+
+Create a boundary condition that wraps an energy flux and converts it to a potential
+temperature flux for use with `LiquidIcePotentialTemperatureFormulation`.
+
+The energy flux is divided by the local mixture heat capacity `cᵖᵐ` to obtain the
+potential temperature flux: `Jᶿ = 𝒬 / cᵖᵐ`.
+"""
+function EnergyFluxBoundaryCondition(flux)
+    ef = EnergyFluxBoundaryConditionFunction(flux, nothing, nothing, nothing, nothing)
```
giordano: In 60d8c40307b47e43ef2ed8d81243339b82112202 I added an extra `nothing` argument here and in the tests, to satisfy the signature, let me know if I got it wrong!

glwagner: hehh

I will take a look but I think this is what we've taken to calling a "stub", which must be "materialized" later with `dynamics_density(dynamics)`

validation/p3/kin1d_reference.nc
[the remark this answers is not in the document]
glwagner: yes, sorry. bad commit

test/forcing_and_boundary_conditions.jl
```diff
@@ -405,13 +388,13 @@ end
         using Breeze.BoundaryConditions: EnergyFluxBoundaryConditionFunction
 
         # Test summary for number condition
-        ef_number = EnergyFluxBoundaryConditionFunction(FT(500), nothing, nothing)
+        ef_number = EnergyFluxBoundaryConditionFunction(FT(500), nothing, nothing, nothing, nothing)
```
navidcy: Do we need the FT here?
```suggestion
        ef_number = EnergyFluxBoundaryConditionFunction(500, nothing, nothing, nothing, nothing)
```

glwagner:
> Looks good!
> 
> Comment: Oceananigans also exports BulkDragFunction, BulkDrag, BulkDragBoundaryCondition
> 
> Do we extend these here? Otherwise it creates ambiguity issues?

This is possible, but I would vote for a new PR because it requires solving a design issue; Breeze uses rho u while Oceananigans uses u.

PR #420, opened 2026-01-20
CCN activation for 2M microphysics

### LINKED ISSUE #416: 2M microphysics lacks aerosol activation for cloud water number concentration
### Problem: 
In PR #405 and `examples/stationary_parcel_model.jl`, an inconsistency can occur between cloud water mass and number concentration during initialization; that is, there can be a non-zero cloud water number concentration even though cloud water has not yet formed. Upon closer inspection, the 2M warm‑phase microphysics state only contains cloud/rain mass and number; there is no aerosol/CCN state or activation source term. As a result, droplet number can only change via collection/breakup, and activation is missing.

### Expected: 
2M microphysics should include aerosol activation (source to ρnᶜˡ/ρqᶜˡ) and, ideally, a corresponding CCN/aerosol sink.

### Possible fix: 
Coupling the 2M microphysics with aerosol activation (`AerosolActivation.jl`). An example can be found in `parcel/ParcelModel.jl` of [CloudMicrophysics](https://github.com/CliMA/CloudMicrophysics.jl.git). nᶜˡ/qᶜˡ coupling (activation of droplets and corresponding mass source):

https://github.com/CliMA/CloudMicrophysics.jl/blob/38d5833944bf24167491356e8273bb217038ea01/parcel/ParcelModel.jl#L117-L123

kaiyuan-cheng:
close #416

### DIFF
--- a/src/AtmosphereModels/atmosphere_model.jl
+++ b/src/AtmosphereModels/atmosphere_model.jl
@@ -87,11 +87,13 @@ AtmosphereModel{CPU, RectilinearGrid}(time = 0 seconds, iteration = 0)
 ├── grid: 8×8×8 RectilinearGrid{Float64, Periodic, Periodic, Bounded} on CPU with 3×3×3 halo
 ├── dynamics: AnelasticDynamics(p₀=101325.0, θ₀=288.0)
 ├── formulation: LiquidIcePotentialTemperatureFormulation
+├── thermodynamic_constants: ThermodynamicConstants{Float64}
 ├── timestepper: SSPRungeKutta3
 ├── advection scheme:
 │   ├── momentum: Centered(order=2)
 │   ├── ρθ: Centered(order=2)
 │   └── ρqᵗ: Centered(order=2)
+├── forcing: @NamedTuple{ρu::Returns{Float64}, ρv::Returns{Float64}, ρw::Returns{Float64}, ρθ::Returns{Float64}, ρqᵗ::Returns{Float64}, ρe::Returns{Float64}}
 ├── tracers: ()
 ├── coriolis: Nothing
 └── microphysics: Nothing
@@ -260,15 +262,30 @@ function Base.summary(model::AtmosphereModel)
                   "(time = ", prettytime(model.clock.time), ", iteration = ", model.clock.iteration, ")")
 end
 
+is_default_atmosphere_model_forcing(::Returns) = true
+is_default_atmosphere_model_forcing(_) = false
+
+function atmosphere_model_forcing_summary(model::AtmosphereModel)
+    forcing = model.forcing
+    names = Tuple(name for name in keys(forcing) if !is_default_atmosphere_model_forcing(forcing[name]))
+
+    isempty(names) && return summary(model.forcing)
+
+    summary_tuple = Tuple(string(name, "=>", nameof(typeof(forcing[name]))) for name in names)
+    return join(summary_tuple, ", ")
+end
+
 function Base.show(io::IO, model::AtmosphereModel)
     TS = nameof(typeof(model.timestepper))
     Mic = nameof(typeof(model.microphysics))
     tracernames = prettykeys(model.tracers)
+    forcing_summary = atmosphere_model_forcing_summary(model)
 
     print(io, summary(model), "\n",
               "├── grid: ", summary(model.grid), "\n",
               "├── dynamics: ", summary(model.dynamics), "\n",
               "├── formulation: ", summary(model.formulation), "\n",
+              "├── thermodynamic_constants: ", summary(model.thermodynamic_constants), "\n",
               "├── timestepper: ", TS, "\n")
 
     if model.advection !== nothing
@@ -281,7 +298,8 @@ function Base.show(io::IO, model::AtmosphereModel)
         print(io, "│   └── " * string(name) * ": " * summary(model.advection[name]), "\n")
     end
 
-    print(io, "├── tracers: ", tracernames, "\n",
+    print(io, "├── forcing: ", forcing_summary, "\n",
+              "├── tracers: ", tracernames, "\n",
               "├── coriolis: ", summary(model.coriolis), "\n",
               "└── microphysics: ", Mic)
 end
--- a/test/atmosphere_model_construction.jl
+++ b/test/atmosphere_model_construction.jl
@@ -30,6 +30,22 @@ end
     model = AtmosphereModel(grid)
     @test model.grid === grid
 
+    @testset "show includes forcing and thermodynamic constants" begin
+        shown_model = sprint(show, model)
+        @test occursin("thermodynamic_constants: ThermodynamicConstants{$FT}", shown_model)
+        @test occursin("forcing: @NamedTuple{", shown_model)
+        @test occursin("ρu::Returns{$FT}", shown_model)
+        @test occursin("ρe::Returns{$FT}", shown_model)
+
+        uᵍ(z) = -10
+        vᵍ(z) = 0
+        coriolis = FPlane(f=FT(1e-4))
+        forced_model = AtmosphereModel(grid; coriolis, forcing=geostrophic_forcings(uᵍ, vᵍ))
+        shown_forced_model = sprint(show, forced_model)
+
+        @test occursin("forcing: ρu=>GeostrophicForcing, ρv=>GeostrophicForcing", shown_forced_model)
+    end
+
     @testset "Basic tests for set!" begin
         set!(model, time=1)
         @test model.clock.time == 1


giordano:
Uhm, there are many seemingly unrelated commits in this PR

ext/BreezeCloudMicrophysicsExt/cloud_microphysics_translations.jl
```diff
@@ -294,12 +313,62 @@ Maximum supersaturation (dimensionless, e.g., 0.01 = 1% supersaturation)
...
+    @inbounds for j in 1:fieldcount(T)
+        κ_mean += mode.vol_mix_ratio[j] * mode.kappa[j]
+    end
+    return κ_mean
+end
+
+@inline _mean_hygroscopicity(ap, mode::Mode_κ{T, FT}) where {T <: Real, FT} = mode.vol_mix_ratio * mode.kappa
 
-    return S_max
+@inline function _mean_hygroscopicity(ap, mode::Mode_B{T, FT}) where {T <: Tuple, FT}
```
glwagner: why prefix this with `_`?

src/AtmosphereModels/microphysics_interface.jl
```diff
@@ -135,27 +137,29 @@ while precipitation (rain, snow) still comes from `μ`.
...
 # Returns
 An `AbstractMicrophysicalState` subtype containing the local specific microphysical variables.
 
 See also [`microphysical_tendency`](@ref), [`AbstractMicrophysicalState`](@ref).
 """
-@inline microphysical_state(::Nothing, ρ, μ, 𝒰) = NothingMicrophysicalState(typeof(ρ))
-@inline microphysical_state(::Nothing, ρ, ::Nothing, 𝒰) = NothingMicrophysicalState(typeof(ρ))
-@inline microphysical_state(microphysics, ρ, ::Nothing, 𝒰) = NothingMicrophysicalState(typeof(ρ))
-@inline microphysical_state(microphysics, ρ, ::NamedTuple{(), Tuple{}}, 𝒰) = NothingMicrophysicalState(typeof(ρ))
+@inline microphysical_state(::Nothing, ρ, μ, 𝒰, w=0, Δt=1) = NothingMicrophysicalState(typeof(ρ))
```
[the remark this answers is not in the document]
glwagner: Also I am not sure that it is possible right now to include `Δt` as an argument. We can make this an objective though. It imposes some restrictions on how we implement time-stepping and can influence the scaling we get in a distributed context (I can explain more if desired). I would advocate for doing this in a new PR.

ext/BreezeCloudMicrophysicsExt/cloud_microphysics_translations.jl
```diff
@@ -211,3 +220,175 @@ Named tuple `(; evap_rate_0, evap_rate_1)` where:
...
+"""
+@inline function max_supersaturation_breeze(
+    aerosol_activation,
+    aps::AirProperties{FT},
+    T::FT,
+    p::FT,
+    w::FT,
+    qᵗ::FT,
+    qˡ::FT,
+    qⁱ::FT,
```
glwagner: can you pass `MoistureMassFractions` directly here?

ext/BreezeCloudMicrophysicsExt/cloud_microphysics_translations.jl
```diff
@@ -211,3 +220,175 @@ Named tuple `(; evap_rate_0, evap_rate_1)` where:
...
+
+        # Critical supersaturation (Eq. 9 in ARG 2000)
+        Sᶜʳⁱᵗ = 2 / sqrt(κ̄) * (A / 3 / mode.r_dry)^(3/2)
+
+        # Fitting parameters (fᵥ and gᵥ are ventilation-related)
+        fᵥ = ap.f1 * exp(ap.f2 * log(mode.stdev)^2)
+        gᵥ = ap.g1 + ap.g2 * log(mode.stdev)
+
+        # η parameter
+        η = (α * w / G)^(3/2) / (2π * ρʷ * γ * mode.N)
```
glwagner:
```suggestion
        η = sqrt(α * w / G)^3 / (2π * ρʷ * γ * mode.N)
```

generally it is good to avoid `pow` on Float if it can be avoided. Note that `a^3` is expanded to `a*a*a` (we call this "literal pow" because "3" is a "literal" number), so `^3` is not a true pow.

ext/BreezeCloudMicrophysicsExt/zero_moment_microphysics.jl
```diff
@@ -28,7 +28,7 @@ AtmosphereModels.materialize_microphysical_fields(bμp::ZMCM, grid, bcs) = mater
     return adjust_thermodynamic_state(𝒰₁, bμp.cloud_formation, constants)
 end
 
-@inline function AtmosphereModels.grid_microphysical_tendency(i, j, k, grid, bμp::ZMCM, ::Val{:ρqᵗ}, ρ, μ, 𝒰, constants)
+@inline function AtmosphereModels.grid_microphysical_tendency(i, j, k, grid, bμp::ZMCM, ::Val{:ρqᵗ}, ρ, μ, 𝒰, constants, velocities, Δt=1)
```
glwagner: I think using a default `Δt=1` is inviting bugs. We shouldn't have any defaults in the interface functions like this, the interface functions should have only one possible calling signature.

src/ParcelModels/parcel_dynamics.jl
```diff
@@ -494,8 +494,8 @@ for [`AtmosphereModel`](@ref) and consolidates all state-dependent computations:
 # Keyword Arguments
 - `compute_tendencies`: If `true` (default), compute tendencies for prognostic variables.
 """
-function TimeSteppers.update_state!(model::ParcelModel, callbacks=[]; compute_tendencies=true)
-    compute_tendencies && compute_parcel_tendencies!(model)
+function TimeSteppers.update_state!(model::ParcelModel, callbacks=[]; compute_tendencies=true, Δt=1.0)
```
glwagner:
```suggestion
function TimeSteppers.update_state!(model::ParcelModel, callbacks=[]; compute_tendencies=true, Δt)
```

ext/BreezeCloudMicrophysicsExt/cloud_microphysics_translations.jl
giordano: It's very stable, the output is always Float64 😁 but yeah, needs to be fixed to retain precision of other variable.

glwagner:
> > > > @kaiyuan-cheng can you explain why you need the time-step within all these functions? I am concerned about it and want to understand precisely where we are using the time-step and for what purpose. I believe there are numerical reasons to invoke a dependency of tendency on time-step, but not physical reasons.
> > > 
> > > 
> > > The aerosol activation scheme computes the number of activated CCN, not a tendency. image I think we need a Δt to convert it to a tendency.
> > 
> > 
> > Doesn't this mean that it is not a prognostic variable? It is a diagnostic or auxiliary variable.
> 
> Yes, it is a diagnostic variable. I’m curious how you typically handle this kind of situation. If we want the model to have a specific cloud water concentration diagnosed by the above formula at a given time (e.g., at the end of a timestep), how should we implement it?

You mean you want do have a scheme where you have knowledge of $N^{n+1}$ in order to compute a tendency at time level $n$?

kaiyuan-cheng:
> > > You mean you want do have a scheme where you have knowledge of N n + 1 in order to compute a tendency at time level n ?
> > 
> > 
> > Yes.
> 
> This leads to an implicit scheme that in general requires a nonlinear solver.
> 
> The only case that it does not require a nonlinear solver is if we are using a particular time-stepping scheme sometimes called "ForwardBackward", where the momentum variables are integrated in a staggered fashion wrt to tracers.
> 
> I don't think we want to write a scheme that cannot be used unless we have forward backward time-stepping. It's not worth it.
> 
> Why is this numerical method required? It is fishy to me that we can only implement some scheme with one numeriucal method. The scheme and numerical method should be independent, otherwise I don't trust the scheme.

Now I understand your concerns. You are right. What I have here is basically the backward Euler method. I have this formulation based on https://github.com/CliMA/CloudMicrophysics.jl/blob/4d6d3f7ec17095aaefd87839de58bffd4a0b3d3a/parcel/ParcelTendencies.jl#L40

glwagner:
I just realized I was totally confused what $u$ is.

docs/src/developer/microphysics/example.md
```diff
@@ -74,9 +74,9 @@ struct ExplicitMicrophysicsState{FT} <: AbstractMicrophysicalState{FT}
     qⁱ :: FT
 end
 
-function microphysical_state(::ExplicitMicrophysics, ρ, μ::NamedTuple, 𝒰, w, Δt)
+function microphysical_state(::ExplicitMicrophysics, ρ, μ::NamedTuple, 𝒰, w)
```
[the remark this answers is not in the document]
glwagner: For turbulence closures there is an interface of sorts, but it is less sophisticated than what we are building for microphysics. And it doesn't include the ability to specify a scheme "pointwise" (because we don't have the analogue of a parcel model that needs a turbulence closure). The momentum flux interface is

https://github.com/NumericalEarth/Breeze.jl/blob/8d43b2219d5849024f2934429a46cbd8e2356802/src/AtmosphereModels/dynamics_kernel_functions.jl#L61

which simply takes in all `fields(model) = model_fields`. This has the velocities, so things like shear can be computed.

ext/BreezeCloudMicrophysicsExt/cloud_microphysics_translations.jl
```diff
@@ -211,3 +220,249 @@ Named tuple `(; evap_rate_0, evap_rate_1)` where:
...
+- `nᵃ`: Aerosol number per unit mass (1/kg)
+- `velocities`: NamedTuple of velocity components `(; u, v, w)` [m/s].
+  The vertical velocity `w` is used for aerosol activation.
+"""
+struct WarmPhaseTwoMomentState{FT, V} <: AbstractMicrophysicalState{FT}
+    qᶜˡ :: FT         # cloud liquid mixing ratio
+    nᶜˡ :: FT         # cloud liquid number per unit mass
+    qʳ  :: FT         # rain mixing ratio
+    nʳ  :: FT         # rain number per unit mass
+    nᵃ  :: FT         # aerosol number per unit mass
```
glwagner: add n^a to notation.md (or make TODO about it)

ext/BreezeCloudMicrophysicsExt/cloud_microphysics_translations.jl
```diff
@@ -211,3 +220,249 @@ Named tuple `(; evap_rate_0, evap_rate_1)` where:
...
+
+    # Compute critical supersaturation and contribution from each mode
+    Σ_inv_Sᵐᵃˣ² = zero(FT)
+    for mode in ad.modes
+
+        # Mean hygroscopicity for mode (volume-weighted κ)
+        κ̄ = mean_hygroscopicity(ap, mode)
+
+        # Critical supersaturation (Eq. 9 in ARG 2000)
+        Sᶜʳⁱᵗ = 2 / sqrt(κ̄) * (A / 3 / mode.r_dry)^(3/2)
```
glwagner:
```suggestion
        Sᶜʳⁱᵗ = 2 / sqrt(κ̄) * sqrt(A / (3 * mode.r_dry))^3
```

ext/BreezeCloudMicrophysicsExt/two_moment_microphysics.jl
```diff
@@ -19,47 +19,113 @@
...
+    # Mode_κ(r_dry, stdev, N, vol_mix_ratio, mass_mix_ratio, molar_mass, kappa)
+    r_dry = 0.05e-6           # 50 nm dry radius
+    stdev = 2.0               # geometric standard deviation
+    N = 100e6                 # 100 cm⁻³
+    vol_mix_ratio = (1.0,)    # single component
+    mass_mix_ratio = (1.0,)
+    molar_mass = (0.132,)     # ammonium sulfate ~132 g/mol
+    kappa = (0.5,)            # hygroscopicity
+
+    mode = CMAM.Mode_κ(r_dry, stdev, N, vol_mix_ratio, mass_mix_ratio, molar_mass, kappa)
```
glwagner:
```suggestion
    mode = CMAM.Mode_κ(r_dry, stdev, nᵃ, vol_mix_ratio, mass_mix_ratio, molar_mass, kappa)
```

ext/BreezeCloudMicrophysicsExt/two_moment_microphysics.jl
```diff
@@ -19,47 +19,113 @@
...
+Returns the sum of `N` (number concentration in 1/m³) across all aerosol modes.
 """
-    TwoMomentCategories{W, AP, LV, RV}
+function initial_aerosol_number_from_activation(aerosol_activation::AerosolActivation)
+    ad = aerosol_activation.aerosol_distribution
+    Nᵃ = zero(eltype(ad.modes[1].N))
+    for mode in ad.modes
+        Nᵃ += mode.N
+    end
+    return Nᵃ
```
glwagner:
```suggestion
    nᵃ = zero(eltype(ad.modes[1].N))
    for mode in ad.modes
        nᵃ += mode.N
    end
    return nᵃ
```

PR #423, opened 2026-01-22
Try Float64 for supercell example from ParcelDynamics PR

glwagner:


giordano:
It works 👀 (it's also a lot slower, but that's probably expected on this GPU)

PR #424, opened 2026-01-22
Add a precipitating microphysics case to the cloudy thermal bubble example

jtbuch:
This PR:

* Extends the cloudy thermal bubble benchmark for a precipitating case by replacing saturation adjustment with `OneMomentCloudMicrophysics`, enabling autoconversion and accretion processes

* Follows CM1 benchmark configuration for a precipitating moist thermal bubble(`iinit=4`, `isnd=4`) with saturated, neutrally-stable sounding (`θᵉ = 320 K`, `qᵗ = 0.020`)

Comparing outputs of the precipitating `cloudy_thermal_bubble.jl` example with that of CM1 (see plot in the comment below) indicates that precipitation does not properly exit the domain in Breeze.jl, causing rain to accumulate indefinitely rather than falling out through the bottom boundary. Enabling open boundary conditions might fix this issue.

### DIFF
--- a/.github/workflows/Whitespace.yml
+++ b/.github/workflows/Whitespace.yml
@@ -6,12 +6,18 @@ on:
   push:
     branches:
       - 'main'
+    paths: &paths
+      - '.github/workflows/Whitespace.yml'
+      - '**.jl'
+      - '**.md'
+      - '**.sh'
   pull_request:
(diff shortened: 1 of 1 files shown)


glwagner:
beautiful! main comment is to fix the aspect ratio of those figures

PR #426, opened 2026-01-23
Add a benchmarking subpackage

glwagner:
This PR adds a benchmarking subpackage to Breeze which will contain "canonical" cases for benchmarking. I started by implementing a convective boundary layer case described in the FastEddy paper:

https://agupubs.onlinelibrary.wiley.com/doi/pdfdirect/10.1029/2020MS002100#page=12.84

I will report back with some GPU benchmarks, and then multiGPU. We can put the results on the rEADME.

I want to discuss how to add this to CI. We can just run the benchmark cases at super low resolution to make sure they run. But another option is to actually generate some benchmarks automatically, so that we can track the evolution of performance. @giordano

### DIFF
--- a/.github/workflows/Benchmarks.yml
+++ b/.github/workflows/Benchmarks.yml
@@ -97,6 +97,19 @@ jobs:
           repositories: |
             Breeze.jl
             BreezeBenchmarks
+      - name: Push results to BreezeBenchmarks repo
+        if: ${{ github.event_name == 'push' }}
+        run: |
+          git clone https://numterra-bot:${{ steps.generate_token.outputs.token }}@github.com/NumericalEarth/BreezeBenchmarks.git
+          cd BreezeBenchmarks
+          DIRNAME="benchmarks/by_branch/${{ github.head_ref || github.ref_name }}/${{ github.sha }}"
+          mkdir -p "${DIRNAME}"
+          cp -v ../benchmark_results.json "${DIRNAME}/."
+          git config --global user.name "numterra-bot[bot]"
+          git config --global user.email "242463128+numterra-bot[bot]@users.noreply.github.com"
+          git add .
+          git commit -m'Save benchmark results for commit ${{ github.ref }}'
+          git push -u origin main
       - uses: benchmark-action/github-action-benchmark@v1
         with:
           name: Breeze.jl Benchmarks
@@ -112,16 +125,3 @@ jobs:
           auto-push: ${{ github.event_name == 'push' }}
           max-items-in-chart: 50
           comment-on-alert: true
-      - name: Push results to BreezeBenchmarks repo
-        if: ${{ github.event_name == 'push' }}
-        run: |
-          git clone https://numterra-bot:${{ steps.generate_token.outputs.token }}@github.com/NumericalEarth/BreezeBenchmarks.git
-          cd BreezeBenchmarks
-          DIRNAME="benchmarks/by_branch/${{ github.head_ref || github.ref_name }}/${{ github.sha }}"
-          mkdir -p "${DIRNAME}"
-          cp -v ../benchmark_results.json "${DIRNAME}/."
-          git config --global user.name "numterra-bot[bot]"
-          git config --global user.email "242463128+numterra-bot[bot]@users.noreply.github.com"
-          git add .
-          git commit -m'Save benchmark results for commit ${{ github.ref }}'
-          git push -u origin main


benchmarking/run_benchmarks.jl
```diff
@@ -0,0 +1,473 @@
...
+        println(io)
+        println(io, "| Benchmark | Float | Grid | Time/Step (ms) | Steps/s | Points/s | Timestamp |")
+        println(io, "|-----------|-------|------|----------------|---------|----------|-----------|")
+
+        for entry in entries
+            grid = entry["grid_size"]
+            grid_str = "$(grid[1])×$(grid[2])×$(grid[3])"
+            timestamp = entry["metadata"]["timestamp"]
+            # Extract just the date portion for brevity
+            date_str = split(timestamp, "T")[1]
```
giordano: Showing only the date is a bit too short for a "timestamp", no?

benchmarking/src/BreezeBenchmarks.jl
```diff
@@ -0,0 +1,43 @@
...
+
+using Dates
+using JLD2
+using Printf
+using Statistics
+
+using Oceananigans
+using Oceananigans.Units
+using Oceananigans.TimeSteppers: time_step!
+using Oceananigans.OutputWriters: JLD2OutputWriter, IterationInterval, TimeInterval, write_output!
```
giordano: `JLD2OutputWriter`, used in the `src/utils.jl` file, doesn't exist

benchmarking/run_benchmarks.jl
```diff
@@ -0,0 +1,473 @@
...
+                entry["time_per_step_seconds"] * 1000,
+                entry["steps_per_second"],
+                entry["grid_points_per_second"],
+                date_str)
+        end
+    end
+end
+
+# Convert struct to Dict, handling nested structs and special types
+struct_to_dict(x) = x
```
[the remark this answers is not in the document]
giordano: When there's a method https://github.com/NumericalEarth/Breeze.jl/blob/749da652ccc5a13f2c17d3cfa8d69733785af4c7/benchmarking/run_benchmarks.jl#L463, this method can never be possibly hit.

benchmarking/run_benchmarks.jl
```diff
@@ -0,0 +1,471 @@
...
+# Convert struct to Dict, handling nested structs and special types
+struct_to_dict(x::Tuple) = collect(x)
+struct_to_dict(x::DateTime) = string(x)
+struct_to_dict(x::String) = x
+function struct_to_dict(x::T) where T
+    isstructtype(T) || return x
+    Dict(string(f) => struct_to_dict(getfield(x, f)) for f in fieldnames(T))
+end
+
+result_to_dict(r) = merge(struct_to_dict(r), Dict("steps_per_second" => 1.0 / r.time_per_step_seconds))
```
giordano: Now I need to go, but all this hacky system should be replaced by [`JSON.lower`](https://juliaio.github.io/JSON.jl/stable/writing/#Custom-Type-Serialization-with-lower) (this is mainly a memo to myself)

giordano: BTW, this is also entirely useless, `JSON.jl` already does the right thing out-of-the-box:
```julia-repl
julia> print(JSON.json(BenchmarkMetadata("CPU"); pretty=true))
{
  "julia_version": "1.12.4",
  "oceananigans_version": "0.104.2",
  "breeze_version": "0.3.1",
  "architecture": "String",
  "gpu_name": null,
  "cuda_version": null,
  "cpu_model": "AMD Ryzen Threadripper PRO 7995WX 96-Cores (znver4)",
  "num_threads": 1,
  "hostname": "hydra",
  "timestamp": "2026-02-02T19:53:41.788"
}
```

benchmarking/README.md
```diff
@@ -0,0 +1,366 @@
...
+├── README.md                         # This file
+├── run_benchmarks.jl                 # Command-line benchmark script
+├── run_gpu_benchmarks.jl             # Comprehensive GPU benchmark suite
+├── plot_benchmarks.jl                # Visualization utilities
+├── benchmark_results.json            # Accumulated results (JSON)
+├── benchmark_results.md              # Auto-generated report
+└── src/
+    ├── BreezeBenchmarks.jl           # Module exports and utilities
+    └── convective_boundary_layer.jl  # CBL benchmark case
+```
```
giordano: Can we just remove this?  It feels oddly overspecific, unnecessary, and easy to get bitrot.  BTW, there's no `run_gpu_benchmarks.jl` nor `plot_benchmarks.jl`.

giordano:
Now I need to leave, in the meantime you can see what the benchmark charts looks like at https://numericalearth.github.io/BreezeBenchmarks/.  For an alternative view, see https://enzymead.github.io/Reactant.jl/benchmarks/.  We could group the lines in a single chart, or group by configuration, etc.  Let me know if you want to benchmark some other configurations.

Still to do for me:

* [ ] fix https://github.com/NumericalEarth/Breeze.jl/pull/426#discussion_r2755660579
* [ ] push the full json file to the `main` branch of https://github.com/NumericalEarth/BreezeBenchmarks

giordano:
Alright, this should be ready for prime time.

There's some refactoring that I had in mind (e.g. moving as much code as possible from the `run_benchmarks.jl` script to the package, with the end goal of making it a [Pkg App](https://pkgdocs.julialang.org/dev/apps)), but those can be done in follow up PRs when time, the basic functionalities we want are here now.

.github/workflows/Benchmarks.yml
```diff
@@ -0,0 +1,127 @@
...
+        with:
+          name: Breeze.jl Benchmarks
+          tool: "customBiggerIsBetter"
+          output-file-path: "github-action-benchmark.json"
+          benchmark-data-dir-path: "."
+          gh-repository: "github.com/NumericalEarth/BreezeBenchmarks"
+          github-token: ${{ github.event.pull_request.head.repo.fork && github.token || steps.generate_token.outputs.token }}
+          comment-always: false
+          summary-always: true
+          alert-threshold: "110%"
```
giordano: During my experiments I never saw performance fluctuating more than 2/3%, in principle we could tighten the alert threshold to 5%, but this can be revised later.

Side note, I really think that this threshold is ill defined: https://github.com/benchmark-action/github-action-benchmark/blob/2e6f4e33d6695578ee68981a04100aa77e0ac037/README.md#minimal-setup

> By default, this action marks the result as performance regression when it is worse than the previous exceeding 200% threshold. For example, if the previous benchmark result was 100 iter/ns and this time it is 230 iter/ns, it means 230% worse than the previous and an alert will happen. The threshold can be changed by `alert-threshold` input.

No!  If previous value was 100 and new one is 230, that's a 130% worsening, not 230%!  If the value stayed at 100 it wouldn't be 100% worse, that doesn't make sense!

Issue #436, opened 2026-01-28
VirtualPotentialTemperature diagnostic looks ...unexpected? for BOMEX


mmr0:
This could well be user error, but I'm trying to plot the virtual potential temperature flux $w'\theta_v'$ for BOMEX to compare to the Siebesma et al 2003 paper (fig 4d)

<img width="536" height="654" alt="Image" src="https://github.com/user-attachments/assets/d389b130-0f93-42cc-b5ac-d185bff06461" />

I tried using the VirtualPotentialTemperature diagnostic (code here https://github.com/NumericalEarth/Breeze.jl/pull/435) but it looks funky (solid line). If I use θᵛapprox = θ * (1 + 0.61 * qᵛ - qˡ)(dashed line) it looks comparable to the Siebesma plot:

<img width="1800" height="2400" alt="Image" src="https://github.com/user-attachments/assets/c49af0b3-a56c-47f0-92d4-60c5b6ec33a3" />

Any ideas? Am I misusing `VirtualPotentialTemperature`?

glwagner:
Ok and finally, on why including the $q^l$ matters. That is interesting. It could possibly be explained by the fact that their is a _very high correlation_ between $w$ and $q^l$. So you may not notice at all if you neglect $q^l$ from the virtual potential temperature, but you will notice if you compute $w' \theta^{v}{'}$.

giordano:
Was this fully fixed by #438 or is there anything else to do?

PR #438, opened 2026-01-28
Fix definition of virtual potential temperature

glwagner:
As discussed on #436 -- thank you @mmr0 !

Here is an explanation repeated from there:

The virtual temperature is defined as

$$ p = \rho R^d T^v $$

So, given the ideal gas law for the mixture $p = \rho R^m T$, it's easy to see that

$$ T^v = \frac{R^m}{R^d} T $$

Then we note that $R^m = q^d R^d + q^v R^v$, and since $q^d = 1 - q^t = 1 - q^v - q^l - q^i$, we have

$$ R^m = R^d \left ( 1 + \delta^v q^v - q^l - q^i \right ) $$

where $\delta^v = \frac{R^v}{R^d} - 1$. So finally

$$ T^v = T \left ( q^d + \frac{R^v}{R^d} \right ) = T \left ( 1 + \delta q^v - q^l - q^i \right ) $$

The virtual potential temperature is defined analogously.


mmr0 (approved):
Looks good! Love the documentation also. 

As a sense check it is now indistinguishable from `θ * (1 + 0.61 * qᵛ - qˡ)` (for BOMEX)

Image

PR #455, opened 2026-02-07
Use kernels for ParcelModel

### LINKED ISSUE #454: Use kernels not nested loops for ParcelModel setting
Looks like an LLM got a little overzealous with loops and it needs to be cleaned up:

https://github.com/NumericalEarth/Breeze.jl/blob/78b2717c7436906138bd9f7998d31c06533a6f8d/src/ParcelModels/parcel_dynamics.jl#L371-L397

Note, the parcel model is almost always run on CPU probably (except if we are doing ensemble-based calibration?) nevertheless, its good to always use kernels for readability and least-suprise.

kaiyuan-cheng:
close #454


glwagner (approved):
negative diff --- ⭐ for you

PR #458, opened 2026-02-07
`sedimentation_speed` interface + effective sedimentation for total water

glwagner:
This PR implements a "sedimentation interface" for microphysics schemes to specify the sedimentation speed of hydrometeors. The intent is to produce a system that automatically computes the effective sedimentation speed of aggregate quantities: total liquid fraction $q^l$, total ice fraction $q^i$ (both diagnostic quantities, possibly not stored), and finally total water fraction $q^t$ (which is required to consistently advect total moisture density as well as density).

### DIFF
--- a/docs/src/appendix/notation.md
+++ b/docs/src/appendix/notation.md
@@ -58,6 +58,7 @@ The following table also uses a few conventions that suffuse the source code and
 | ``\mathbb{W}^{ci}``                 | `𝕎ᶜⁱ`  |                                     | Terminal velocity of cloud ice (scalar, positive downward)                     |
 | ``\mathbb{W}^r``                    | `𝕎ʳ`   |                                     | Terminal velocity of rain (scalar, positive downward)                          |
 | ``\mathbb{W}^s``                    | `𝕎ˢ`   |                                     | Terminal velocity of snow (scalar, positive downward)                          |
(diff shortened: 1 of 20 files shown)


glwagner:
looking into it

glwagner:
Still digesting this, but here is a summary of one reason for the instability:

> Root cause: The fall-speed-interface branch introduced a bulk sedimentation velocity for ρqᵗ (total water), computed as wᵗ = -Σ(wᵢ qᵢ)/qᵗ.  This is physically correct (derived by summing individual species equations), but numerically unstable because WENO reconstruction of ρ wᵗ qᵗ produces a different discrete flux than WENO reconstruction of ρ wᵢ qᵢ for the individual species. This discrete inconsistency causes qᵗ < qʳ over time, leading to negative vapor, negative temperature, and the DomainError in saturation_vapor_pressure.

In other words. In order to consistently include this advection term in multiple equations, it has to be computed in exactly the same way. We can't use WENO to compute the moisture density flux and rain flux separately, because the advection fluxes are numerically different.

This is serious difficulty in prognosing total water. I'd like to think about it a bit more. cc @kaiyuan-cheng 

This difficulty also must be addressed if we wanted to include the hydrometeor fluxes in the density equation, or the energy equation (when using static or total energy).

glwagner:
superceded by #614

Issue #433, opened 2026-01-26
Reactant compilation failure as a function of grid size


dkytezab:
On the CPU with Julia 1.11.8. If I try and run the first test on the REPL in `test/differentiation.jl'`, given as
```julia
    grid = RectilinearGrid(ReactantState(); size=(11, 11), extent=(1000, 1000),
                           halo=(3, 3), topology=(Periodic, Periodic, Flat))
    model = AtmosphereModel(grid; dynamics=CompressibleDynamics())
    dmodel = Enzyme.make_zero(model)

    θ_init = CenterField(grid)
    dθ_init = CenterField(grid)
    set!(θ_init, (x, y) -> 0.01 * x + 0.01 * y)
    set!(dθ_init, 0.0)

    function loss(model, θ_init, Δt, nsteps)
        set!(model, θ=θ_init, ρ=1.0)
        @trace mincut=true checkpointing=false track_numbers=false for i in 1:nsteps
            time_step!(model, Δt)
        end
        return mean(interior(model.temperature).^2)
    end

    function grad_loss(model, dmodel, θ_init, dθ_init, Δt, nsteps)
        parent(dθ_init) .= 0
        _, loss_value = Enzyme.autodiff(
            Enzyme.set_strong_zero(Enzyme.ReverseWithPrimal),
            loss, Enzyme.Active,
            Enzyme.Duplicated(model, dmodel),
            Enzyme.Duplicated(θ_init, dθ_init),
            Enzyme.Const(Δt),
            Enzyme.Const(nsteps))
        return dθ_init, loss_value
    end

    Δt = 0.01
    nsteps = 4

    compiled = Reactant.@compile raise_first=true raise=true sync=true grad_loss(
            model, dmodel, θ_init, dθ_init, Δt, nsteps)
```

with a grid of size `(11, 11)` or larger, I get the following error
```julia
julia> include("test/differentiation.jl")
[...]

PR #461, opened 2026-02-09
Change kernel launching to `:xyz` to support further Reactant integration

dkytezab:
See https://github.com/NumericalEarth/Breeze.jl/issues/433. Launching `compute_velocities!` with `:xyz` rather than `KernelParameters` works around this bug.

Note that Reactant integration tests are blocked via some upstream issues.

### DIFF
--- a/src/AtmosphereModels/update_atmosphere_model_state.jl
+++ b/src/AtmosphereModels/update_atmosphere_model_state.jl
@@ -79,6 +79,7 @@ function compute_velocities!(model::AtmosphereModel)
 
     #TODO: Better support OffsetStaticSize in KernalAbstractions
     # For now, just use :xyz instead of KernelParameters
+    # See: https://github.com/NumericalEarth/Breeze.jl/issues/433
 
     # TX, TY, TZ = topology(grid)
     # Nx, Ny, Nz = size(grid)


src/AtmosphereModels/update_atmosphere_model_state.jl
```diff
@@ -77,23 +77,26 @@ function compute_velocities!(model::AtmosphereModel)
     grid = model.grid
     arch = grid.architecture
 
-    TX, TY, TZ = topology(grid)
-    Nx, Ny, Nz = size(grid)
-    Hx, Hy, Hz = halo_size(grid)
+    #TODO: Better support OffsetStaticSize in KernalAbstractions
+    # For now, just use :xyz instead of KernelParameters
```
giordano: Maybe we should add a reference to https://github.com/NumericalEarth/Breeze.jl/issues/433, to provide context to the readers?

glwagner: also note that we may have found a bug in `diagnostic_indices`.

giordano: Is there an issue/PR about it?

glwagner: there is only this PR

PR #429, opened 2026-01-23
Enable Reactant compilation of + tracing through Breeze timestepping

dkytezab:
See #427. 

- Enables Reactant tracing of SSPRK3 timestepping via the addition of `BreezeReactantExt` that enables Reactant-friendly `Clock` mutation.
- Tests Breeze integration of Reactant + Enzyme with a MWE in `test/differentiation.jl`.
- Adds Enzyme/Reactant to test environment.

PR #464, opened 2026-02-10
[CI] Disable GPU support for Reactant until we actually start using it

giordano:
After #429 we're currently wasting time (and storage) downloading large GPU artifact for not actually using it at all.


giordano:
We definitely don't use Reactant for benchmarks nor docs at the moment, and the Reactant test is disabled in Julia v1.12, and the GPU job is only in v1.12.  So there's no case where Reacant is used with a GPU at the moment.

glwagner:
Ah. If we want to test Reactant we need a GPU job for 1.11. Hmm...

PR #463, opened 2026-02-10
Allows construction of `AnelasticDynamics` and compilation of `time_step!` with Reactant

dkytezab:
See https://github.com/NumericalEarth/Breeze.jl/issues/462.

PR #470, opened 2026-02-13
[CI] Bump minimum Julia version to v1.11.9

giordano:
@dkytezab is this what you suggested in https://github.com/NumericalEarth/Breeze.jl/pull/463#issuecomment-3894144609?


dkytezab:
Yup! Though I wonder if we want to run non-Reactant tests on Julia min. I'm not sure. I don't feel strongly about that to be honest.

.github/workflows/CI.yml
```diff
@@ -41,7 +41,7 @@ jobs:
       fail-fast: false
       matrix:
         version:
-          - "min"
+          - "1.11"
           - "1.12.4"
```
navidcy:
```suggestion
          - "1.12.5"
?

glwagner (approved):
I am comfortable with saying that Breeze supports 1.11.9--1.12.4

PR #474, opened 2026-02-13
Fix quality of `BreezeReactantExt` extension

giordano:


glwagner (approved):

giordano:
At least now the error
```
  Module `BreezeReactantExt.TimeSteppers` was found to be unanalyzable. Include this module in the `allow_unanalyzable` keyword argument to allow it to be unanalyzable.
```
is happening consistently everywhere, instead of randomly depending on the scheduling of the jobs 😃  Ok, I'm going to skip that.

@ericphanson maybe you have a clue of why this is happening?  Only In Julia v1.10-v1.11 though, works in v1.12.

PR #482, opened 2026-02-17
Add dynamics and microphysics benchmark suites

glwagner:
Generalize the benchmark script to support compressible dynamics (fully explicit and split-explicit with 12 fixed acoustic substeps). Add dynamics and microphysics matrix jobs to the CI benchmark workflow, both at 512x512x256 resolution. Update the publish step to merge results from all benchmark artifact directories.


glwagner:
wut

PR #488, opened 2026-02-18
RICO with 2M microphysics

### LINKED ISSUE #473: Use 2M microphysics with CCN activation in RICO example?
I think it would be nice to use RICO to have a dynamics-coupled validation case. We can just switch over the example to use the 2M microphysics. What do you think @kaiyuan-cheng ?

kaiyuan-cheng:
close #473


kaiyuan-cheng:
This is not needed anymore.

PR #489, opened 2026-02-18
Refactor moisture handling with scheme-dependent prognostic naming

glwagner:
## Summary

- Replace hardcoded `:ρqᵗ` prognostic key with scheme-dependent names: `:ρqᵛ` (non-equilibrium/Nothing) and `:ρqᵉ` (equilibrium moisture density --- often vapor, cloud liquid, and cloud ice)
- Rename `model.vapor_density` → `model.moisture_density`; remove `model.specific_moisture` field (replaced by `specific_prognostic_moisture(model)` function accessor)
- Non-equilibrium microphysics schemes now store true vapor `qᵛ` as the prognostic variable instead of total moisture `qᵗ`
- Add `grid_moisture_fractions` (reads vapor directly from fields for grid models) vs `moisture_fractions` (subtracts condensate from total moisture for parcel models)

### New public API
- `moisture_prognostic_name(microphysics)` → `:ρqᵛ` or `:ρqᵉ`
- `moisture_specific_name(microphysics)` → `:qᵛ` or `:qᵉ`
- `specific_prognostic_moisture(model)` → specific moisture field from microphysical fields

### Breaking changes
- `model.vapor_density` → `model.moisture_density`
- `model.specific_moisture` removed (use `specific_prognostic_moisture(model)`)
- `vapor_mass_fraction` removed (use `specific_humidity`)
- Boundary condition keys: `:ρqᵗ` → `:ρqᵛ` (or `:ρqᵉ` for SA)

## Test plan
- [x] Full test suite passes (1521/1521)
- [x] Run BOMEX example (SA path with `:ρqᵉ`)
- [x] Run cloudy_thermal_bubble example (non-eq path with `:ρqᵛ`)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #490, opened 2026-02-18
Make rain evaporation in 2M microphysics GPU-safe

kaiyuan-cheng:
The rain_evaporation_2m function uses an if/else block with mutable local variables (evap_rate_0, evap_rate_1 initialized before the if and reassigned inside it). This pattern is GPU-unsafe.


ext/BreezeCloudMicrophysicsExt/cloud_microphysics_translations.jl
```diff
@@ -179,47 +179,53 @@ Named tuple `(; evap_rate_0, evap_rate_1)` where:
...
+    (; xr_mean) = pdf_rain_parameters(pdf_r, qʳ_safe, ρ, Nʳ_safe)
+    xr_mean_safe = max(xr_mean, eps(FT))
+    Dʳ = cbrt(6 * xr_mean_safe / (π * ρᴸ))
 
-        a_vent_1 = av * Γ(FT(2)) / cbrt(FT(6))
-        b_vent_1 = bv * Γ(5 // 2 + 3 // 2 * β) / 6^(β / 2 + 1 // 2)
+    # Ventilation factors for number and mass tendencies
+    t_star = cbrt(6 * x_star / xr_mean_safe)
+    a_vent_0 = av * Γ_incl(FT(-1), t_star) / FT(6)^(-2 // 3)
+    b_vent_0 = bv * Γ_incl(-1 // 2 + 3 // 2 * β, t_star) / FT(6)^(β / 2 - 1 // 2)
```
glwagner: better notation for these?

ext/BreezeCloudMicrophysicsExt/cloud_microphysics_translations.jl
```diff
@@ -179,47 +179,53 @@ Named tuple `(; evap_rate_0, evap_rate_1)` where:
...
+    evap_rate_0 = min(zero(FT), 2π * G * 𝒮 * Nʳ_safe * Dʳ * Fv0 / xr_mean_safe)
+    evap_rate_1 = min(zero(FT), 2π * G * 𝒮 * Nʳ_safe * Dʳ * Fv1 / ρ)
+
+    # Handle edge cases where xr_mean approaches zero
+    evap_rate_0 = ifelse(xr_mean / x_star < eps(FT), zero(FT), evap_rate_0)
+    evap_rate_1 = ifelse(qʳ < eps(FT), zero(FT), evap_rate_1)
+
+    # Zero out when no evaporation should occur
+    evap_rate_0 = ifelse(evaporating, evap_rate_0, zero(FT))
+    evap_rate_1 = ifelse(evaporating, evap_rate_1, zero(FT))
```
glwagner: do we have a notation for evaporation rates?

glwagner (approved):
seems good to go to me, if the notation can be improved, that'll be great

glwagner:
> I think we need a general method to keep tracers positive definite, rather than clamping them individually in each scheme.

Who is clamping tracers? We should not do that. I don't think it works either.

I'm investigating some possibilities on #507 but the first one is to limit the tendencies.

kaiyuan-cheng:
> @kaiyuan-cheng can we merge this?

Sure. The merge button isn’t available for me, and I’m not sure if I’m able to do a manual merge. Please go ahead.

glwagner:
hmm yeah we should give you the ability to override

glwagner:
well, now we can continue on #507 . Doing a bit of work on it, I think we may need #489 first though.

PR #491, opened 2026-02-19
Remove SSPRungeKutta3 time stepping function

glwagner:
Removed SSPRungeKutta3 time stepping implementation from Timesteppers.jl.


glwagner:
It looks to be superceded. @dkytezab reopen if thats not true.

PR #496, opened 2026-02-20
[benchmarks] Actually use microphysics

### LINKED ISSUE #495: Microphysics not actually used in benchmarks
I'm staring at the benchmarking code https://github.com/NumericalEarth/Breeze.jl/blob/9354383fb192bf792f7c391ef4a28c785fbcffb9/benchmarking/run_benchmarks.jl#L314-L342 and, unless I'm missing something obvious, `microphysics` isn't actually used.  My hunch is that all the benchmarks in [this group](https://numericalearth.github.io/BreezeBenchmarks/#benchmark-cbl--dynamics--anelastic--grid--512x512x256--f32-) are just running the same model all over again, although it's not clear to me why `1M_MixedNonEquilibrium` and `1M_MixedEquilibrium` are systematically lower than the other two, which just swaps their timings across different runs.

giordano:
Fix #495.

### DIFF
--- a/.github/workflows/Benchmarks.yml
+++ b/.github/workflows/Benchmarks.yml
@@ -167,7 +167,7 @@ jobs:
             Breeze.jl
             BreezeBenchmarks
       - name: Push results to BreezeBenchmarks repo
-        # if: ${{ github.event_name == 'push' }}
+        if: ${{ github.event_name == 'push' }}
         run: |
           # Clone BreezeBenchmarks
           git clone https://numterra-bot:${{ steps.generate_token.outputs.token }}@github.com/NumericalEarth/BreezeBenchmarks.git


benchmarking/src/convective_boundary_layer.jl
```diff
@@ -52,7 +55,9 @@ function convective_boundary_layer(arch = CPU();
                                    Nx = 64, Ny = 64, Nz = 64,
                                    dynamics = nothing,
                                    advection = WENO(order=5),
-                                   closure = SmagorinskyLilly())
+                                   closure = SmagorinskyLilly(),
+                                   microphysics = nothing,
```
giordano: The actual default is still to be decided, we'll evaluate it.

PR #497, opened 2026-02-20
[benchmarks] Add comparison with WENO9 advection scheme

giordano:


giordano:
Question: do we want to run both WENO5 and WENO9, to compare them, or just switch everything to WENO9?

giordano:
What do you think of reducing the number of time steps, to contain the duration of the benchmarks? That doesn't seem to affect the results significantly. If this is good, I'll clean up the PR (need to fix the directory where results are pushed) and this will be ready.

PR #507, opened 2026-02-20
Squall line example + microphysics numerics improvements

glwagner:
## Summary

- Add 2D idealized squall line example based on the Rotunno-Klemp-Weisman framework
- Warn in `AtmosphereModel` constructor when microphysical tracers lack bounds-preserving advection
- Thread the simulation `clock` through the `microphysical_tendency` call chain, preparing for tendency limiters that will use `clock.last_Δt` to prevent source terms from driving variables negative

## Context

The squall line is a stress test for microphysics numerics — rain evaporating into the dry rear inflow jet creates large evaporation tendencies. Testing with `TwoMomentCloudMicrophysics` revealed that:
1. Bounded WENO alone doesn't prevent negativity (source terms overshoot within RK3 substeps)
2. Small timesteps don't help (blowup is physics-driven, occurs at ~45 min sim time regardless of Δt)
3. Microphysical tendency limiters are needed

## Test plan
- [x] atmosphere_model_construction (192/192)
- [x] dynamics (80/80)
- [x] dcmip2016_kessler (12/12)
- [x] cloud_microphysics_1M (62/62)
- [x] cloud_microphysics_2M (70/70)
- [x] parcel_dynamics (93/93)


giordano:
Unless this PR is intended to be only 657c75d3396e0c60ccf4e2315e6446a301576211 (in which case the PR title and description would need to be updated), because that change alone passes the tests, so this would be good to go without much hassle.

glwagner:
> Unless this PR is intended to be only [657c75d](https://github.com/NumericalEarth/Breeze.jl/commit/657c75d3396e0c60ccf4e2315e6446a301576211) (in which case the PR title and description would need to be updated), because that change alone passes the tests, so this would be good to go without much hassle.

Sorry about the misleading names. The purpose of this PR is to explore whether more aggressive limiting of the microphysical tendencies will produce better numerical properties. However, in order to investigate this issue, it also made sense to introduce a new example which provides a larger stress test for the microphysics numerics.

glwagner:
> @glwagner Where is the idealized squall line example? I couldn't find it in this PR.

looking for it! i will also resolve the conflicts

kaiyuan-cheng:
@glwagner I am thinking that we should split the squall line example out of this PR and get it merged first. That way, we'll have a clean version to experiment with ways to stablize microphyiscs-dynamics coupling.

glwagner:
> @glwagner I am thinking that we should split the squall line example out of this PR and get it merged first. That way, we'll have a clean version to experiment with ways to stablize microphyiscs-dynamics coupling.

fine with me. although, i don't completely understand why that would help here?

glwagner:
ah ok I understand. I actually don't think its a good idea to merge that to main personally (even if we would plan on improving it). I'm also not sure why we would want to merge it (eg what part of that algorithm do we actually want to keep?)

PR #515, opened 2026-02-23
Add buoyancy-driven updraft to ParcelDynamics

kaiyuan-cheng:
### Summary

- Add BuoyancyDrivenUpdraft option for ParcelDynamics where the parcel has a prognostic vertical velocity driven by buoyancy: dw/dt = B, dz/dt = w

- Buoyancy B = -g(ρ_parcel - ρ_env)/ρ_env uses the density-based formulation from Thermodynamics.density

- Default remains SpecifiedUpdraft (backward compatible)


kaiyuan-cheng:
> should we use this in the example?

With prognostic vertical velocity, the thermodynamic profile in the rising-parcel case continually accelerates the updraft, producing unrealistically strong velocities in the upper domain, which causes the microphysics scheme to fail. I've adjusted the domain and configuration to prevent the run from crashing.

PR #516, opened 2026-02-23
[CI] Add a job with `--check-bounds=auto`

giordano:


giordano:
I'd prefer postponing that if possible, because the GPU jobs use the the Docker images at https://github.com/NumericalEarth/breeze-docker-images, so would need first to set up the new image.

Also, the new job is failing because for some reason everything is being recompiled, will need to figure out why, but unluckily to have the time today 😞

PR #521, opened 2026-02-26
[docs] Temporarily ignore Oceananigans links

giordano:
Publishing of the Oceananigans docs is currently broken (should hopefully be fixed soon), in the meantime we ignore links to Oceananigans docs in linkchecker.

PR #522, opened 2026-02-26
[docs] Check again links to Oceananigans docs

giordano:
Ref: https://github.com/NumericalEarth/Breeze.jl/pull/521#issuecomment-3967313633.


giordano:
I'm starting to think the link to the docstring of `Fields.set!` was a hallucination, it doesn't look like it ever existed (which maybe should?).

PR #523, opened 2026-02-26
[Reactant] Test raising of `CompressibleDynamics`

dkytezab:
* Re-factors existing tests for raising + compiling `CompressibleDynamics` into a test with Centered advection and a test with WENO
* Uses the grid topologies that are known to work when raising both the forward and the backward pass.

### DIFF
--- a/.github/workflows/Benchmarks.yml
+++ b/.github/workflows/Benchmarks.yml
@@ -187,7 +187,7 @@ jobs:
           name: Breeze.jl Benchmarks
           tool: "customBiggerIsBetter"
           output-file-path: "github-action-benchmark.json"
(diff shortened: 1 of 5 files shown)


test/reactant_centered_compilation.jl
```diff
@@ -0,0 +1,130 @@
...
+
+#####
+##### Tests
+#####
+
+@testset "Reactant CompressibleDynamics — Centered" begin
+    Δt_val = 0.02
+
+    for (label, topo, nd) in topologies
+        @info "Centered advection: $label"
```
giordano: I don't think these `@info` messages are useful with ParallelTestRunner as they aren't printed live
```suggestion
```

dkytezab:
@glwagner can we merge this?

dkytezab:
sounds good, I can create an issue to restore `test_reactant_weno` to the normal test set once serial running is implemented

PR #532, opened 2026-02-28
Upgrade Julia version in Documentation workflow

navidcy:
Updated Julia version in the documentation workflow from 1.12.4 to 1.12.5.


giordano:
The Docker image comes from https://github.com/NumericalEarth/breeze-docker-images, more specifically the Julia version is set in  https://github.com/NumericalEarth/breeze-docker-images/blob/e9bdf4094fb5ae8ba5723cb38228c010aa66cd3b/.github/workflows/DockerPublish.yml#L41-L55

PR #547, opened 2026-03-06
Fix water conservation in 1M and 2M microphysics and numerical stability in 2M microphysics

kaiyuan-cheng:
## Problem

Water vapor was being excessively depleted in simulations using the non-equilibrium 1M and 2M microphysics schemes (WPNE1M, MPNE1M, WPNE2M). These schemes implement `microphysical_tendency` for cloud liquid, cloud ice, and rain, but were missing the corresponding `Val{:ρqᵛ}` method. As a result, vapor received no net tendency from microphysics — condensation and deposition removed vapor without any balancing update, leading to unrealistically negative vapor in long simulations (e.g. squall line).

Additionally, the advection operator is not positive-definite, so individual moisture species and number concentrations could become negative between time steps, causing unphysical states. Also, the 2M microphysics itself seems to have stability issues.

## Fix

**Vapor tendency methods** (one_moment_microphysics.jl, two_moment_microphysics.jl):
- Add `microphysical_tendency(bμp, Val{:ρqᵛ}, ...)` for WPNE1M, MPNE1M, and WPNE2M
- Vapor tendency = −(condensation + deposition + aerosol activation) − rain evaporation,
  ensuring mass conservation across all phase changes
- Limit vapor removal to available vapor via a numerical relaxation floor (−qᵛ / τⁿᵘᵐ)

**ARG 2000 numerical guards** (cloud_microphysics_translations.jl):
- Guard intermediate quantities with `max(0, ...)` and `max(eps(FT), ...)` to prevent
  NaN from fractional powers of negative numbers or division by zero
- Guard Sᵐᵃˣ and Sᶜʳⁱᵗ before log/sqrt in the activated fraction calculation

**Local total water conservation**:
Refactored tendency computation in both 1M and 2M schemes from per-variable methods into a single bundle function (`wpne1m_tendencies` and `wpne2m_tendencies`) that computes all process rates once and returns all tendencies as a named tuple. This guarantees discrete conservation: the same rate value appears with opposite signs in every tendency that references it, so ρqᵛ + ρqᶜˡ + ρqʳ = 0 by construction. Each per-variable `microphysical_tendency` dispatcher now delegates to the bundle and extracts its component.

**Negative moisture correction via vertical borrowing**:
`negative_moisture_correction.jl` — a column-wise kernel that runs at the start of update_state!, before tendency computation:
  - Same-level borrowing: fixes negative hydrometeors by transferring mass along the chain rain ← cloud ← vapor within each grid cell
  - Vertical borrowing: remaining negative vapor is redistributed vertically (top→bottom sweep, then one upward step), preserving column-integrated moisture
  - Number consistency: zeros orphaned number concentrations (positive nᶜˡ/nʳ with zero mass) and clamps negative number
  - Microphysics schemes extend `correction_moisture_fields`, `correction_number_mass_pairs`, and `correction_number_fields` to define their borrowing chains

**Coupled sink limiting for 2M microphysics**:
For the 2M scheme, added per-reservoir tendency limiting that computes a scaling factor α for each reservoir (vapor, cloud, rain, aerosol) to prevent any reservoir from being depleted faster than q/τⁿᵘᵐ per time step. Mass and number sinks within the same reservoir share the most restrictive (minimum) α, preventing microphysics from depleting mass faster than number or vice versa.

τⁿᵘᵐ is now a configurable field on TwoMomentCategories (default: 10 s) rather than a module-level constant.

**Terminal velocity fix**: 
Enforce minimum number density floor Nᶜˡ_min = ρqᶜˡ/xc_max (and similarly for rain) to prevent unphysically large mean droplet mass when advection creates qᶜˡ > 0, nᶜˡ ≈ 0.


glwagner (commented):
thank you!

in the longer term, I would like to have a design that _guarantees_ discretely that the sources of one tracer are sinks of another. Let's brainstorm how to implement that.

glwagner:
@kaiyuan-cheng tell me if this makes sense:

* we always want to clamp negative number concentrations and zero orphaned number concentrations --- these should be baked into 2M schemes?
* "same level borrowing" should be a scheme. I might call this "SpeciesBorrowing" though to clarify that we are transferring water from one species to another. This is relatively cheap and can be done in a 3D kernel.
* "vertical borrowing" requires a loop over `Nz`

With this logic we need to have two schemes for fixing negatives --- we can allow a tuple of "fixers" so that multiple fixes can be applied at once. In terms of a default, I am not sure. I think for LES, vertical borrowing is suspect (The vertical direction is not as special in LES as it would be in a coarser model). Therefore vertical borrowing may not be universal to all configurations. Species borrowing is perhaps universal.

src/AtmosphereModels/negative_moisture_correction.jl
```diff
@@ -1,25 +1,91 @@
...
+"""
+$(TYPEDEF)
+
+Redistribute remaining negative vapor vertically within each column after
+species borrowing: a top-to-bottom sweep pushes deficits downward, then one
+bottom-to-top step borrows from the level above if the bottom is still negative.
+
+Column-integrated moisture is conserved (Δz-weighted).
+"""
+struct VerticalBorrowing end
```
glwagner: should this also subtype `AbstractNegativeMoistureCorrection`?

glwagner (approved):
looking good!

glwagner (commented):
small q

giordano:
Always forgot to ask: based on the numbers in https://numericalearth.github.io/BreezeBenchmarks/#benchmark-cbl--dynamics--anelastic--grid--512x512x256--float32-, this PR seems to have made the 1M_MixedNonEquilibrium benchmark [systematically slower](https://github.com/NumericalEarth/Breeze.jl/pull/547#pullrequestreview-3947751666), I presume that's expected because this PR made some changes to address accuracy issues?

glwagner:
I think it is because this PR basically requires microphysics calculations to be repeated for all species. Is that right @kaiyuan-cheng ? This is helpful for numerical stability, but might be possible to achieve the same outcome with a cheaper method with less redundancy.

glwagner:
It's pretty severe too, a few PRs in a row all decreased performance by almost 20%:

image

giordano:
I think the other big regression was https://github.com/NumericalEarth/Breeze.jl/pull/587#pullrequestreview-4044449916, perhaps unavoidable.

glwagner:
> #587 (review)

right, that is a necessary correctness improvement. 

Probably what we should try to understand is why there is slow down on this PR. On #587 the slowdown makes sense because we add more calculations. Although I am somewhat surprised because I thought that microphysics was not a big cost per time-step, especially considering these use the anelastic solver which has a quite expensive pressure solve.

PR #553, opened 2026-03-11
Fix NaN in saturation adjustment

### LINKED ISSUE #551: [docs] NaNs in tropical cyclone world example
In [v0.4.0](https://numericalearth.github.io/BreezeDocumentation/v0.4.0/literated/tropical_cyclone_world/#Run) the tropical cyclone world is ending due to the presence of NaNs.  This wasn't the case in [v0.3.3](https://numericalearth.github.io/BreezeDocumentation/v0.3.3/literated/tropical_cyclone_world/#Run).  Greg suggested to enable the NaN checker to prevent this kind of regression in the future.  In the meantime I'm correctly setting the RNG seed for the GPU code in #550, to make the examples more reproducible.  Then, we should figure out when this regressed (I may be able to do a bisection later this week) and then enable the NaN checker once this problem is fully resolved.

kaiyuan-cheng:
Summary

  - Fix three bugs in the saturation adjustment secant iteration that could produce NaN, especially in Float32:
    a. Second initial guess used all-vapor q₁ instead of the condensate-adjusted 𝒰₁.moisture_mass_fractions, making ΔT ≈ 0 always
    b. No guard against stagnation: when r₂ ≈ r₁, ΔTΔr = Inf → NaN propagation. Added GPU-safe ifelse(isfinite(...)) masking
    c. Default maxiter = Inf allowed NaN to propagate indefinitely; changed to maxiter = 10
  - Fix GPU branching to use ifelse instead of short-circuit patterns (kernel compatibility)
  - Add regression test for NaN robustness with a highly supersaturated Float32 state

  Test plan

  - Existing saturation adjustment tests pass
  - New Saturation adjustment NaN robustness test passes for Float32 and Float64
  - GPU tests pass (no branching in kernels)

close #551

### DIFF
--- a/src/Microphysics/saturation_adjustment.jl
+++ b/src/Microphysics/saturation_adjustment.jl
@@ -212,12 +212,10 @@ Return the saturation-adjusted thermodynamic state using a secant iteration.
     iter = 0
 
     while abs(r₂) > δ && iter < microphysics.maxiter
-        # Compute slope; guard against stagnation (r₂ = r₁ → ΔTΔr = Inf → NaN).
+        # Compute slope; guard against stagnation (r₂ = r₁ → division by zero).
         ΔTΔr = (T₂ - T₁) / (r₂ - r₁)
-
-        # Mask the components; mathematically bypass instead of breaking
         valid_step = isfinite(ΔTΔr)
-        ΔTΔr *= valid_step
+        ΔTΔr = ifelse(valid_step, ΔTΔr, zero(FT))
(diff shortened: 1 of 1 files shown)


giordano:
The performance issue is a glitch: we've had a few times occasional performance drops for the 128x128x128 grid.  The [rerun](https://github.com/NumericalEarth/Breeze.jl/actions/runs/22972481423?pr=553) was good.

PR #562, opened 2026-03-14
Add FieldTimeSeriesRelaxation

ewquon:
This implements a Newtonian relaxation forcing term, to be used for grid/analysis nudging. The example script depends on ERA5 data retrieval developed in https://github.com/NumericalEarth/NumericalEarth.jl/pull/93. 

In the single-column example, nudging is applied above 1.5km to a domain with slip-wall boundary and uniform initial fields. 

6-hour timescale:
single_column_nudging_tau6h

1-hour timescale (more responsive):
single_column_nudging_tau1h


src/Forcings/relaxation_forcing.jl
```diff
@@ -0,0 +1,240 @@
+using ..AtmosphereModels: AtmosphereModels
+using Oceananigans: Average, Field, set!, compute!
+using Oceananigans.BoundaryConditions: fill_halo_regions!
+using Oceananigans.Fields: AbstractField, location, interior
+using Oceananigans.Grids: Center, λnodes, φnodes, znode
+using Oceananigans.OutputReaders: Time
```
giordano: Based on the [error message](https://github.com/NumericalEarth/Breeze.jl/actions/runs/23078107692/job/67042203343?pr=562#step:10:597):
```suggestion
using Oceananigans.Units: Time
```

examples/single_column_nudging.jl
```diff
@@ -0,0 +1,153 @@
...
+
+# We downloaded a subset of the ERA5, defined by bounding_box. Because this data grid is not
+# coincident with the simulation grid, we need to perform a mapping to the reference lat, lon.
+
+temp = Field(Metadatum(:temperature; dataset, date=first(dates), bounding_box))
+Δλ = temp.grid.Δλᶜᵃᵃ
+Δφ = temp.grid.Δφᵃᶜᵃ
+
+xn = [ref_loc.longitude - Δλ/2, ref_loc.longitude + Δλ/2]
+yn = [ref_loc.latitude  - Δφ/2, ref_loc.latitude  + Δφ/2]
```
navidcy: we can use `xnodes` and `ynodes` as well?
or just

```Julia
xn, yn, zn = Oceananigans.nodes(temp.grid, Center, Center, Face, with_halos=false)
```
?

examples/single_column_nudging.jl
```diff
@@ -0,0 +1,153 @@
...
+#
+# This configuration is suitable for driving a column LES with large-scale
+# reanalysis data (e.g. ERA5), where the mesoscale environment is prescribed
+# and only the sub-grid turbulence is explicitly simulated.
+
+using Breeze
+using Oceananigans
+using Oceananigans.Units
+using Oceananigans.TurbulenceClosures
+using Oceananigans.Fields: fractional_x_index, fractional_y_index
```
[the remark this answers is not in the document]
ewquon: Ah, that was a carryover of some other analysis / sanity checking I was doing. I was doing something like the following—not sure if there's a better way?
```
icol = round(Int, fractional_x_index(ref_longitude, (Center(), Center(), Center()), my_field_timeseries.grid))
jcol = round(Int, fractional_y_index(ref_latitude, (Center(), Center(), Center()), my_field_timeseries.grid))
my_profile = view(my_field_timeseries, icol, jcol, :)
```

glwagner:
> > I suggest we call this `HorizontalAverageRelaxation`. In a future PR we can consider merging with `Oceananigans.Forcing.Relaxation`.
> 
> @glwagner This naming only captures half the functionality, for an idealized simulation configuration. The more general use case of this forcing is to nudge a larger (mesoscale) domain with open boundaries towards a (possibly coarser) background field to ensure that the solution remains physically consistent with the larger-scale flow as it evolves in time and space.
> 
> How about `GridNudging` which is consistent with meteorological community usage?

That works for me. However, I do not think we should implement the more general 3D nudging in this feature. Instead, I think we should merge this feature with `Oceananigans.Forcing.Relaxation` first. After we merge the two, we can implement the 3D nudging to a 3D FieldTimeSeries in Oceananigans. It will also be useful for open boundary conditions for ocean models.

What do you think @ewquon ? The point of proposing `HorizontalAverageRelaxation` was simply to propose a temporary name which would be soon deleted when this is merged into Oceananigans.

src/Forcings/relaxation_forcing.jl
```diff
@@ -0,0 +1,240 @@
...
+
+#####
+##### RelaxationForcing struct
+#####
+
+struct RelaxationForcing{R, TF, Cl, D, F, RC, TS, ZB}
+    reference        :: R    # FieldTimeSeries of the specific variable ϕᵣ
+    target           :: TF   # Field of ϕᵣ interpolated to current time, horizontally averaged to a single column in profile mode; Nothing pre-materialization
+    clock            :: Cl   # Model Clock; Nothing pre-materialization
+    density          :: D    # Reference density ρᵣ(z); Nothing pre-materialization
```
glwagner: is this always reference density, even for `CompressibleDynamics`?

examples/single_column_nudging.jl
```diff
@@ -0,0 +1,215 @@
...
+
+simulation = Simulation(model; Δt = 60seconds, stop_time = T_ref.times[end])
+
+# Progress reporting
+function progress(sim)
+    u_max = maximum(abs, sim.model.velocities.u)
+    θ_mean = mean(sim.model.temperature)
+    @printf "t = %s, max|u| = %.2f m/s, mean T = %.2f K\n" prettytime(sim) u_max θ_mean
+end
+simulation.callbacks[:progress] = Callback(progress, IterationInterval(60))
```
glwagner:
```suggestion
    @printf "t = %s, max|u| = %.2f m/s, mean T = %.2f K\n" prettytime(sim) u_max θ_mean
    return nothing
end

simulation.callbacks[:progress] = Callback(progress, IterationInterval(60))
```

src/Forcings/fts_relaxation.jl
```diff
@@ -0,0 +1,242 @@
...
+F_{ρ ϕ} = -\\frac{ρ_r \\left(\\bar{ϕ} - ϕ_r\\right)}{τ} \\, \\mathbf{1}_{z \\geq z_b}
+```
+
+where ``\\bar{ϕ}`` is either the local field value (3D mode) or its horizontal
+average (profile mode), ``ϕ_r`` is the time-interpolated reference value,
+``ρ_r`` is the reference density, ``τ`` is the relaxation time scale, and
+``z_b`` is the cutoff height below which no nudging is applied.
+
+The `reference` [`FieldTimeSeries`](@ref) must contain the specific variable
+(e.g. ``u``, ``v``, ``θ``, ``q^t``), not the density-weighted form. The user is
```
giordano: Can't use `@ref` with symbols defined in other packages
````suggestion
toward a reference `FieldTimeSeries`:

```math
F_{ρ ϕ} = -\\frac{ρ_r \\left(\\bar{ϕ} - ϕ_r\\right)}{τ} \\, \\mathbf{1}_{z \\geq z_b}
```

where ``\\bar{ϕ}`` is either the local field value (3D mode) or its horizontal
average (profile mode), ``ϕ_r`` is the time-interpolated reference value,
``ρ_r`` is the reference density, ``τ`` is the relaxation time scale, and
``z_b`` is the cutoff height below which no nudging is applied.

The `reference` `FieldTimeSeries` must contain the specific variable
````

Maybe one day we'll explore [`DocumenterInterLinks.jl`](https://github.com/JuliaDocs/DocumenterInterLinks.jl), but we aren't using anything like that at the moment.

***Edit***: I think GitHub messed up my suggestion, this was meant to be ``[`FieldTimeSeries`](@ref)`` -> `` `FieldTimeSeries` ``.

examples/single_column_nudging.jl
glwagner, line 1: I think this example should eventually be moved to NumericalEarth (and added to the docs build if we keep it). We can keep this for a future PR though.

ewquon:
This feature can be replicated by Oceananigans' `Relaxation` forcing, following https://github.com/CliMA/Oceananigans.jl/pull/5575 and https://github.com/NumericalEarth/Breeze.jl/pull/708.

glwagner:
and also: https://github.com/CliMA/Oceananigans.jl/pull/5620

I will post to here to prove it once that lands

PR #520, opened 2026-02-26
Add LatitudeLongitudeGrid support with CompressibleDynamics

glwagner:
Preliminary support for using `LatitudeLongitudeGrid` with `CompressibleDynamics`.

- Import and export LatitudeLongitudeGrid, SphericalCoriolis, HydrostaticSphericalCoriolis from Oceananigans
- Add baroclinic wave example on near-global LatitudeLongitudeGrid with ExplicitTimeStepping, thermal-wind-balanced jet, hydrostatic density initialization, and sphere visualization
- Add latitude_longitude_grid test file (23 tests covering model construction, adaptive substeps, balanced state stability, SSP-RK3, WS-RK3, and explicit compressible on LatitudeLongitudeGrid)
- Add Jablonowski & Williamson 2006 citation to breeze.bib
- Add example entry to docs/make.jl (build_always=false)

cc @jtbuch 

Closes #518 

Here's an animation of what the baroclinic wave example looks like so far. Clearly needs a bit of work.

https://github.com/user-attachments/assets/cf264649-87fa-47a3-bfd5-ebc6da03b4a8

PR #566, opened 2026-03-17
Update `BreezeReactantExt` to enable `ReactantState` on `LatitudeLongitudeGrid`

dkytezab:
- Fixes two problems related to `set!` called on concrete arrays. See https://github.com/NumericalEarth/Breeze.jl/pull/520.


giordano:
The new tests seem to be very expensive, might be better to move them to separate files for better load balancing. Also, maybe we want to move all the Reactant tests to a `reactant/` subdirectory? If we need some common init code for all these tests it can be put in a, say, `reactant/setup.jl` file which all these test files will include (but remember to always delete it from the testsuite in `runtests.jl`)

giordano:
Reactant tests at the moment are run only in a single job because they're already very expensive

giordano:
@dkytezab the rebase went wrong 🥲

dkytezab:
claude got a little uppity./.. but i think now this is ok? cleaning this up atm

PR #573, opened 2026-03-18
Finite difference validation tests

dkytezab:
Adds finite-difference gradient verification to both reactant compilation test files. Closes https://github.com/NumericalEarth/Breeze.jl/issues/571.

### DIFF
--- a/test/reactant_centered_compilation.jl
+++ b/test/reactant_centered_compilation.jl
@@ -87,65 +87,63 @@ end
 @testset "Reactant CompressibleDynamics — Centered" begin
     Δt = 0.02
 
-    for (label, topo, nd) in topologies
-        @testset "$label" begin
-            grid = make_grid(topo, nd)
+    @testset "$label" for (label, topo, nd) in topologies
+        grid = make_grid(topo, nd)
 
-            # ── Build ──
-            @testset "Build" begin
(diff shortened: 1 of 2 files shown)


test/reactant_centered_compilation.jl
```diff
@@ -87,45 +87,69 @@ end
...
+
+                for ε in [1e-4, 1e-6]
+                    for (ic, jc, kc) in test_cells
+                        θ_fd = CenterField(grid_fd); set!(θ_fd, (args...) -> 300.0)
+                        @allowscalar interior(θ_fd, ic, jc, kc)[] += ε
+                        J₊ = loss(make_fd_model(), θ_fd, Δt, Ns)
+                        fd = (J₊ - J₀) / ε
+                        ad = ad_grad[ic, jc, kc]
+                        rel = abs(ad - fd) / (max(abs(ad), abs(fd)) + eps())
+                        @test rel < 0.01
```
giordano: This should be pretty much the same as
```suggestion
                        @test ad ≈ fd rtol=0.01
```
Why did you use 1% tolerance?  Did you get failures with tighter values?  I guess the problem is that we can't fully trust finite difference anyway.

giordano (approved):
🚀

PR #575, opened 2026-03-20
Allow Ocenanaigans v0.106

giordano:
Based on the [release notes of v0.106](https://github.com/CliMA/Oceananigans.jl/releases/tag/v0.106.0) we shouldn't be affect by breaking changes.  I guess we'll see how CI goes.


giordano:
Uhm, nothing is actually using Oceananigans v0.106 at the moment, because other packages aren't compatible with it yet.

navidcy:
Like which package(s)?

glwagner (approved):

PR #565, opened 2026-03-17
Add AD component to acoustic wave example

dkytezab:
- Adds Enzyme, Reactant to deps for docs
- Adds an AD visualization to the Breeze documentation
- Computes derivative of density perturbation at point after an acoustic wave has propagated

image

PR #578, opened 2026-03-24
Add verifiable AD example

dkytezab:
Reaction-diffusion with prescribed advection, for which we can compute an analytical gradient for the continuous problem. We verify that, with increased grid resolution, the AD truncation error decays suitably


dkytezab:
Turns out that flag was deprecated.... oops!

examples/advection_diffusion_ad.jl
```diff
@@ -0,0 +1,309 @@
...
+# is undefined; instead we normalise by ``J/U_0`` (the natural scale for
+# a velocity-derivative of the objective) to obtain a dimensionless
+# quantity comparable to the other relative errors.
+
+Δx_vec = [r.Δx for r in results]
+safe(v) = max.(v, eps(Float64))
+
+fig = Figure(size = (820, 520), fontsize = 14)
+ax  = Axis(fig[1, 1];
+    xscale = log2, yscale = log2,
```
giordano: Why using log2 instead of log10?  Maybe it's just me, but I find log10 a more "natural" scale

examples/advection_diffusion_ad.jl
```diff
@@ -0,0 +1,309 @@
+# # Advection–Diffusion of a Gaussian: AD Verification
+#
+# This example verifies reverse-mode automatic differentiation (AD)
```
giordano:
```suggestion
# This example verifies reverse-mode [automatic differentiation](https://en.wikipedia.org/wiki/Automatic_differentiation) (AD)
```

examples/advection_diffusion_ad.jl
```diff
@@ -0,0 +1,309 @@
...
+#   \exp\!\Bigl(-\frac{(x - U_0 t)^2 + y^2}{2\,s(t)}\Bigr).
+# ```
+
+A   = 2.0       # amplitude
+σ₀  = 20.0      # initial width               [m]
+κ   = 40.0      # isotropic diffusivity        [m²/s]
+U₀  = 10.0      # advection velocity           [m/s]
+L   = 200.0     # domain side length           [m]
+t_f = 1.0       # integration time             [s]
+CFL  = 0.5       # CFL number  (Δt = CFL × Δx / max_speed)
```
giordano:
```suggestion
CFL  = 0.5      # CFL number  (Δt = CFL × Δx / max_speed)
```

giordano (commented):
I must say I kinda liked the example in #565 more because the physical interpretation was more intuitive and the discussion about sensitivity was an interesting application, on the other hand having a reference derivative to compare with was a lot harder.  I'm not complaining about this example, I guess we can't easily have both a nice and intuitive example + simple analytical derivatives (and for the purpose of showcasing effectiveness of the autodiff approach, verifiable derivatives are probably more useful)

glwagner:
The example is AMAZING but I am wondering if we can make it slightly more exciting by plotting the sensitivity. Or something. Also illustrating what exactly the example does might be neat. I wonder if the example should actually first verify the gradient, and then do a second simulation/AD calculation that is more interesting. Also lets use "Autodifferentiation" in the title rather than AD

@dkytezab I can work on these additions if you like

dkytezab:
I like that idea; maybe we should tweak reference velocity field to be varying in both x and y so it's not such a far departure... Also why not include both AD examples but separately? 

We could try plugging Breeze into an optimization loop as well.

glwagner:
> I like that idea; maybe we should tweak reference velocity field to be varying in both x and y so it's not such a far departure... Also why not include both AD examples but separately?
> 
> We could try plugging Breeze into an optimization loop as well.

what if we start with a gaussian spot diffusing, show gradient for this, and then for round 2 add a thermal bubble which will introduce some cool nonlinearity into it

dkytezab:
After a discussion with @glwagner, I believe that we should move this analytical example to the tests and instead simply  use https://github.com/NumericalEarth/Breeze.jl/pull/565 to demo AD

PR #588, opened 2026-03-27
Switch splitting supercell to mixed-phase 1M microphysics

glwagner:
## Summary
- Replace DCMIP2016 Kessler (warm-rain only) with CloudMicrophysics.jl `OneMomentCloudMicrophysics` using mixed-phase saturation adjustment, adding prognostic snow and diagnosed cloud ice
- Update visualization to 4-panel output (w, cloud liquid, rain, snow)
- Set `build_always=true` in `docs/make.jl` so the supercell example builds on every commit

## Test plan
- [ ] CI docs build passes (supercell example now runs every commit)
- [ ] Verify the simulation completes without NaN on GPU


glwagner:
not needed anymore

PR #589, opened 2026-03-27
Bump all environments to Oceananigans 0.106.3

dkytezab:


giordano:
Why is this necessary? Note that Oceananigans v0.106.3 is already _allowed_, just not strictly _required_ (i.e. other previous versions are allowed, too)

PR #590, opened 2026-03-30
Add DynamicSmagorinsky to RICO

glwagner:
@ewquon


glwagner (commented):
couple comments

glwagner:
the investigation is stale so I'll close this, but easy to re-start if needed

PR #591, opened 2026-03-30
Add sedimentation latent heat tendency ρe and ρθ for 1M and 2M

### LINKED ISSUE #364: Microphysics sedimentation does not change total water or liquid-ice potential temperature

kaiyuan-cheng:
This PR addresses an issue where sedimentation was unable to modulate the liquid-ice potential temperature. We are encountering the same challenge previously discussed regarding the advection of total water and moist density: no consistent way to advect moisture-coupled thermodynamic quantities alongside falling hydrometeors.

To at least address the missing influence of sedimentation on the liquid-ice potential temperature, we implemented a 1st-order upwind scheme to compute the energy flux due to sedimentation. We recognize that this approach is not fully consistent with the mass flux.

close #364 and supersede #383

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -111,8 +111,8 @@ jobs:
               # than the single one it'd use by default.
               TEST_ARGS+=(--jobs=2)
               # ...but also set a more conservative limit to memory usage before
-              # reciclying a worker, to avoid out-of-memory issues.
-              JULIA_TEST_MAXRSS_MB=2500
+              # reciclying a worker, to avoid trashing memory or out-of-memory errors.
+              JULIA_TEST_MAXRSS_MB=1800
               echo "JULIA_TEST_MAXRSS_MB=${JULIA_TEST_MAXRSS_MB}" | tee -a "${GITHUB_ENV}"
           fi
           echo "runtest_test_args=${TEST_ARGS[@]}" | tee -a "${GITHUB_ENV}"


kaiyuan-cheng:
@glwagner Here is what I think. Start from the diagnostic definition of θ_li at constant T:

`θˡⁱ = (T - (ℒˡᵣ qˡ + ℒⁱᵣ qⁱ) / cᵖᵐ) / Π`

Differentiate it by applying the chain rule, and setting dT = 0 (assuming sedimentation moves liquid water mass without exchanging sensible heat with the air), we get 

`∂θˡⁱ/∂t = -(ℒˡᵣ/(cᵖᵐ Π)) × ∂qˡ/∂t`

glwagner (commented):
comment

kaiyuan-cheng:
> ```
> │ WARNING: Method definition sedimentation_mass_flux_z(Any, Any, Any, Any, Any, Any, Any, Any, Any) in module BreezeCloudMicrophysicsExt at /Users/runner/work/Breeze.jl/Breeze.jl/ext/BreezeCloudMicrophysicsExt/one_moment_microphysics.jl:855 overwritten at /Users/runner/work/Breeze.jl/Breeze.jl/ext/BreezeCloudMicrophysicsExt/two_moment_microphysics.jl:586.
> │ ERROR: Method overwriting is not permitted during Module precompilation. Use `__precompile__(false)` to opt-out of precompilation.
> ```

Good catch! Thanks!

PR #602, opened 2026-04-01
[CI] Further reduce memory limit for macOS CI jobs

giordano:
Let's see if this helps, together with #601


giordano:
With this change the two macOS jobs are the quickest to finish 😆

PR #603, opened 2026-04-02
Add error message when required keyword arguments to BulkDrag are missing

fergu:
This PR adds an error message when a `BulkDrag` boundary condition is configured with `coefficient=PolynomialCoefficient()`, but the required `surface_temperature` kwarg is omitted.

This fixes a minor issue I stumbled across where if `surface_temperature` is omitted, a rather non-specific error stating that division between `nothing` and `Float64` types is not defined will be raised during the first simulation step. This PR just adds an explicit error message if this required argument is missing.


glwagner (approved):
thank you!

PR #604, opened 2026-04-02
See if energy is conserved in moist thermal bubble with 1M microphysics

glwagner:
cc @kaiyuan-cheng


kaiyuan-cheng:
Total energy is conserved!

glwagner:
we aren't using this anymore, so closing

PR #612, opened 2026-04-03
Don't promote in CloudMicrophysicsExt via `4π` 

dkytezab:
A subtlety: π is not f64 but a constant times π is.


glwagner:
indeed one of the ways that Irrational is poorly specified (Int * Irrational = ... F64?!)

Issue #534, opened 2026-03-02
Implement SurfaceLayerDiffusivity (Sullivan et al. 1994 mean-field surface viscosity)


glwagner:
## Background

Standard LES SGS closures produce mean wind and temperature profiles that deviate from Monin-Obukhov similarity theory (MOST) in the surface layer. The error grows with increasing shear and persists even with horizontal grid refinement. [Sullivan, McWilliams, and Moeng (1994, *Boundary-Layer Meteorology*, 71, 247–276)](https://www2.mmm.ucar.edu/people/sullivan/talks/papers/sgs.pdf) proposed a two-part SGS eddy-viscosity model that decomposes the SGS stress as

$$\tau_{ij} = -2\nu_t \gamma S_{ij} - 2\nu_T \langle S_{ij} \rangle$$

where $\langle \cdot \rangle$ denotes horizontal averaging, $\nu_t$ is the standard (fluctuating/isotropic) eddy viscosity, $\nu_T(z)$ is a mean-field eddy viscosity derived from MOST, and $\gamma(z)$ is an isotropy factor that damps the fluctuating closure near the surface.

The mean-field term $\nu_T \langle S_{ij} \rangle$ is only a function of $z$ — it acts as a vertical diffusion of the horizontally-averaged profile. The isotropy factor

$$\gamma = \frac{S'}{S' + \langle S \rangle}$$

where $S' = \sqrt{2\langle(S_{ij} - \langle S_{ij}\rangle)(S_{ij} - \langle S_{ij}\rangle)\rangle}$ is the rms fluctuating strain and $\langle S \rangle = \sqrt{2\langle S_{ij}\rangle \langle S_{ij}\rangle}$ is the mean strain magnitude, transitions from 0 near the surface (mean shear dominates, base closure suppressed) to 1 in the interior (base closure recovered). Above a configurable matching height (~ $0.5 z_i$), set $\gamma = 1$ and $\nu_T = 0$.

Sullivan et al. found that the momentum correction alone improves both velocity and temperature profiles without an explicit scalar correction, though a scalar analogue is straightforward.

## Mean-field viscosity $\nu_T(z)$

At the first $w$-grid point $z_1$ (grid index $k=2$), MOST matching gives (Eq. 23, simplified for ILES where $\langle \nu_t \gamma \rangle = 0$):

[...]

PR #615, opened 2026-04-07
Add neutral atmospheric boundary layer example

ewquon:
This is based on the shear-driven ABL case from Moeng & Sullivan's 1994 JAS paper -- and subsequently used to develop the NCAR LES SGS scheme (see also https://github.com/NumericalEarth/Breeze.jl/issues/534). This example uses a prescribed surface shear stress (through ustar). Currently, I'm using this to assess flowfield behaviors in the surface layer (e.g., due to the overshoot problem, cf. Brasseur & Wei 2010 Phys. Fluids). Runtime is ~15 min on a T4 GPU. 

nabl_mean_profiles

nabl_velocity_variances

nabl_fluxes


examples/neutral_atmospheric_boundary_layer.jl
```diff
@@ -0,0 +1,425 @@
...
+using Oceananigans: Oceananigans
+using Oceananigans.Units
+using CUDA
+using Printf
+using Random
+using CairoMakie
+
+Random.seed!(42)
+if CUDA.functional()
+    CUDA.seed!(42)
```
glwagner:
```suggestion
Random.seed!(1994)
if CUDA.functional()
    CUDA.seed!(1994)
```

docs/src/breeze.bib
```diff
@@ -584,6 +584,51 @@ @article{Sullivan2014
...
+  volume = {51},
+  number = {7},
+  pages = {999--1022},
+  year = {1994},
+  doi = {10.1175/1520-0469(1994)051<0999:ACOSAB>2.0.CO;2}
+}
+
+@article{Berg2020,
+  author = {Berg, Jacob and Patton, Edward G. and Sullivan, Peter P.},
+  title = {Large-{{Eddy Simulation}} of {{Conditionally Neutral Boundary Layers}}: {{A Mesh Resolution Sensitivity Study}}},
```
navidcy:
```suggestion
  title = {{Large-eddy simulation of conditionally neutral boundary layers: A mesh resolution sensitivity study}},
```

docs/src/breeze.bib
```diff
@@ -584,6 +584,51 @@ @article{Sullivan2014
...
+  volume = {153},
+  number = {1},
+  pages = {43--62},
+  year = {2014},
+  doi = {10.1007/s10546-014-9937-z}
+}
+
+@article{Sauer2020,
+  author = {Sauer, Jeremy A. and {Mu{\~n}oz-Esparza}, Domingo},
+  title = {The {{FastEddy}}\textregistered{} {{Resident-GPU Accelerated Large-Eddy Simulation Framework}}: {{Model Formulation}}, {{Dynamical-Core Validation}} and {{Performance Benchmarks}}},
```
navidcy:
```suggestion
  title = {The {FastEddy}\textregistered{} resident-{GPU} accelerated large-eddy simulation framework: {Model} formulation, dynamical-core validation and performance benchmarks},
```

docs/src/breeze.bib
```diff
@@ -584,6 +584,51 @@ @article{Sullivan2014
...
+  journal = {Journal of Advances in Modeling Earth Systems},
+  volume = {12},
+  number = {11},
+  pages = {1--30},
+  year = {2020},
+  doi = {10.1029/2020MS002100}
+}
+
+
```
navidcy:
```suggestion
```

examples/neutral_atmospheric_boundary_layer.jl
```diff
@@ -0,0 +1,429 @@
...
+z = (0, 1000)
+
+grid = RectilinearGrid(arch; x, y, z,
+                       size = (Nx, Ny, Nz), halo = (3, 3, 3),
+                       topology = (Periodic, Periodic, Bounded))
+
+# ## Reference state and formulation
+
+p₀ = 1e5
+θ₀ = 300
```
navidcy:
```suggestion
p₀ = 1e5   # Pa
θ₀ = 300   # K
```

examples/neutral_atmospheric_boundary_layer.jl
```diff
@@ -0,0 +1,431 @@
...
+sponge_width = 200  # m
+sponge_mask = GaussianMask{:z}(center = last(z), width = sponge_width)
+
+ρw_sponge = Relaxation(rate = sponge_rate, mask = sponge_mask) # relax to 0 by default
+
+# relax to initial temperature profile
+ρᵣ = reference_state.density
+ρθˡⁱᵣ = Field{Nothing, Nothing, Center}(grid)
+set!(ρθˡⁱᵣ, z -> θᵣ(z))
+set!(ρθˡⁱᵣ, ρᵣ * ρθˡⁱᵣ)
```
navidcy: Is this cleaner? I think it's a bit cleaner because we don't set the ρθ field as θ and then set again to multiply itself with reference density, even if it means defining an extra field for θ.

```suggestion

θᵣ = Field{Nothing, Nothing, Center}(grid)
set!(θᵣ, z -> θᵣ(z))

ρθˡⁱᵣ = Field{Nothing, Nothing, Center}(grid)
set!(ρθˡⁱᵣ, ρᵣ *  θᵣ)

```

navidcy: Of course the above won't work because it uses θᵣ both for the function and the field..

glwagner: I think we use this pattern in the rico and bomex cases

navidcy: oh gotcha; then let's do it here as well then!

examples/neutral_atmospheric_boundary_layer.jl
```diff
@@ -0,0 +1,435 @@
...
+                                 surface_pressure = p₀,
+                                 potential_temperature = θ₀)
+
+dynamics = AnelasticDynamics(reference_state)
+
+# Capping inversion for "S" simulation, as in the paper by [Moeng1994](@citet).
+Δz  = first(zspacings(grid))
+zᵢ₁ = 468        # m
+zᵢ₂ = zᵢ₁ + 6Δz  # m
+Δθi = 8 / 6Δz    # K
```
navidcy: how does Δθi has units K if it's divided. by Δz?

Also later, `θᵣ(z) = ... + Δθi * (zᵢ₂ - zᵢ₁) + ...` which implies that `Δθi` has units K/m?

If Δθi is a temperature gradient then let's denote it with Γ?

examples/neutral_atmospheric_boundary_layer.jl
```diff
@@ -0,0 +1,435 @@
...
+Γᵗᵒᵖ = 0.003     # K/m, = dθ/dz
+θᵣ(z) = z < zᵢ₁ ? θ₀ :
+        z < zᵢ₂ ? θ₀ + Δθi * (z - zᵢ₁) :
+        θ₀ + Δθi * (zᵢ₂ - zᵢ₁) + Γᵗᵒᵖ * (z - zᵢ₂)
+
```
navidcy, line 71:
```suggestion
Γᵢ  = 8 / 6Δz    # K/m
Γᵗᵒᵖ = 0.003     # K/m
θᵣ(z) = z < zᵢ₁ ? θ₀ :
        z < zᵢ₂ ? θ₀ + Γᵢ * (z - zᵢ₁) :
        θ₀ + Γᵢ * (zᵢ₂ - zᵢ₁) + Γᵗᵒᵖ * (z - zᵢ₂)
```

navidcy (approved):
I pushed a few commits. Let's see the docs preview. But I believe it should look good now!

examples/neutral_atmospheric_boundary_layer.jl
```diff
@@ -0,0 +1,426 @@
...
+u, v, w = model.velocities
+θ = liquid_ice_potential_temperature(model)
+νₑ = model.closure_fields.νₑ
+
+## For keeping track of the computational expense
+wall_clock = Ref(time_ns())
+
+function progress(sim)
+    wmax = maximum(abs, sim.model.velocities.w)
+    elapsed = 1e-9 * (time_ns() - wall_clock[])
```
giordano: Why is `wall_clock` a `Ref`?  It's only used in the `progress` function to read its value
```suggestion
## For keeping track of the computational expense
wall_clock = time_ns()

function progress(sim)
    wmax = maximum(abs, sim.model.velocities.w)
    elapsed = 1e-9 * (time_ns() - wall_clock)
```
would work as well, and be simpler

docs/make.jl
```diff
@@ -35,6 +35,7 @@ examples = [
     Example("Precipitating shallow cumulus (RICO)", "rico", false),
     Example("Convection over prescribed sea surface temperature (SST)", "prescribed_sea_surface_temperature", true),
     Example("Inertia gravity wave: many time steppers", "inertia_gravity_wave", true),
+    Example("Neutral atmospheric boundary layer", "neutral_atmospheric_boundary_layer", true),
```
ewquon: @navidcy Should this be set to false since it's a long-running example?

giordano: Yeah, I think it'd be a good idea.  In that case we'd need to apply the label "build all the examples" to this PR before making the change, so that they're all built (and we can check whether concurrent build is still ok).  ***Edit***: I already applied the label anyway.

ewquon:
Ninth-order WENO result (https://github.com/NumericalEarth/Breeze.jl/issues/534#issuecomment-4238532716) shows the most resolved turbulent shear stress; the 6th-order centered scheme originally used actually shows slightly worse resolution in the surface layer, since dispersive errors can lead to greater dissipation as @glwagner pointed out.

examples/neutral_atmospheric_boundary_layer.jl
```diff
@@ -186,16 +186,22 @@ end
...
 
-# Output averaged products for calculating variances
-outputs = merge(model.velocities, model.tracers, (; θ, νₑ,
-                                                    uu = u*u, vv = v*v, ww = w*w,
-                                                    uw = u*w, vw = v*w, θw = θ*w))
+# Output averaged products for calculating statistics
+avg_outputs_varlist = (;
+    θ, νₑ,
+    uu = u*u, vv = v*v, ww = w*w,
+    uw = u*w, vw = v*w, θw = θ*w,       # second-order moments for fluxes
```
navidcy: I saw that there is some post-process interpolation to bring, e.g., w or w^2 on Centers.
If we don't really need the w, w^2 on Faces, we can interpolate them via the output writer? Would that be useful?

docs/make.jl
navidcy: Yeah, I just had set it as `true` to see it in the docs preview.

examples/neutral_atmospheric_boundary_layer.jl
```diff
@@ -0,0 +1,435 @@
...
+if CUDA.functional()
+    CUDA.seed!(1994)
+end
+
+# ## Domain and grid
+#
+# For faster time to solution, the numerical precision may be reduced to Float32.
+
+arch = GPU()
+Oceananigans.defaults.FloatType = Float64;
```
navidcy: Since Float64 is the default, no need to explicitly declare `Oceananigans.defaults.FloatType = Float64`, right?

```suggestion
# For faster time to solution, the numerical precision may be reduced to Float32; we can change the default float type by redefining `Oceananigans.defaults.FloatType`.

arch = GPU()
```

examples/neutral_atmospheric_boundary_layer.jl
navidcy: E.g., we can do 


```Julia
avg_outputs = NamedTuple(name => Average(@at((Center, Center, Center), outputs[name]), dims=(1, 2))
        for name in keys(outputs))
```

and then all outputs are at the cell's centers.

ewquon:
Looks like the docs won't build with Float64, this case is too slow...

examples/neutral_atmospheric_boundary_layer.jl
```diff
@@ -0,0 +1,434 @@
...
+                 LineElement(color=:black, linestyle=:dot)]
+axislegend(ax3a, style_entries, ["total", "resolved", "SGS"],
+           position=:lt, framevisible=false)
+
+linkyaxes!(ax3a, ax3b, ax3c)
+hideydecorations!(ax3b, grid=false)
+hideydecorations!(ax3c, grid=false)
+
+Legend(fig3[1, 4], ax3a, "Time", framevisible=false)
+fig3
```
navidcy, line 442:
```suggestion
# Let's load the saved output.

u_ts  = FieldTimeSeries(avg_filename, "u")
v_ts  = FieldTimeSeries(avg_filename, "v")
w_ts  = FieldTimeSeries(avg_filename, "w")
θ_ts  = FieldTimeSeries(avg_filename, "θ")
uu_ts = FieldTimeSeries(avg_filename, "uu")
vv_ts = FieldTimeSeries(avg_filename, "vv")
ww_ts = FieldTimeSeries(avg_filename, "ww")
uw_ts = FieldTimeSeries(avg_filename, "uw")
vw_ts = FieldTimeSeries(avg_filename, "vw")
θw_ts = FieldTimeSeries(avg_filename, "θw")
νₑ_ts = FieldTimeSeries(avg_filename, "νₑ")

grid = u_ts.grid
times = u_ts.times
Nt = length(times)

# Compute diagnostics at each saved time. First, we create a few empty timeseries to save the computed diagnostics.

loc = (nothing, nothing, Center())

WS_mean_ts = FieldTimeSeries(loc, grid, times)
WD_mean_ts = FieldTimeSeries(loc, grid, times)
 uu_var_ts = FieldTimeSeries(loc, grid, times)
 vv_var_ts = FieldTimeSeries(loc, grid, times)
 ww_var_ts = FieldTimeSeries(loc, grid, times)
 uw_res_ts = FieldTimeSeries(loc, grid, times)
 vw_res_ts = FieldTimeSeries(loc, grid, times)
 θw_res_ts = FieldTimeSeries(loc, grid, times)
 uw_sgs_ts = FieldTimeSeries(loc, grid, times)
 vw_sgs_ts = FieldTimeSeries(loc, grid, times)
 θw_sgs_ts = FieldTimeSeries(loc, grid, times)

# and then we loop over all saved fields and compute what we want.

for n in 1:Nt
    u_n = u_ts[n]
    v_n = v_ts[n]
    w_n = Field(@at loc w_ts[n]) # interpolate to cell centers
    θ_n = θ_ts[n]
    νₑ_n = νₑ_ts[n]
    uu_n = uu_ts[n]
    vv_n = vv_ts[n]
    ww_n = Field(@at loc ww_ts[n]) # interpolate to cell centers
    uw_n = uw_ts[n]
    vw_n = vw_ts[n]
    θw_n = θw_ts[n]

    WS_mean_ts[n] .= sqrt(u_n^2 + v_n^2)
    WD_mean_ts[n].data .= @. mod(270 - atand(v_n.data, u_n.data), 360)
    uu_var_ts[n] .= (uu_n - u_n^2) / u★^2
    vv_var_ts[n] .= (vv_n - v_n^2) / u★^2
    ww_var_ts[n] .= (ww_n - w_n^2) / u★^2
    uw_res_ts[n] .= ρᵣ * (uw_n - u_n * w_n) / (ρ₀ * u★^2)
    vw_res_ts[n] .= ρᵣ * (vw_n - v_n * w_n) / (ρ₀ * u★^2)
    θw_res_ts[n] .= θw_n - θ_n * w_n

    ∂z_u = @at loc ∂z(u_n)
    ∂z_v = @at loc ∂z(v_n)
    ∂z_θ = @at loc ∂z(θ_n)

    uw_sgs_ts[n] .= -ρᵣ * νₑ_n * ∂z_u / (ρ₀ * u★^2)
    vw_sgs_ts[n] .= -ρᵣ * νₑ_n * ∂z_v / (ρ₀ * u★^2)
    θw_sgs_ts[n] .= -νₑ_n * ∂z_θ / closure.Pr
end

# Define a colormap for each time.

cmap = cgrad(:viridis)
colors = [cmap[(n-1)/max(Nt-1, 1)] for n in 1:Nt]
labels = [n == 1 ? "0–1 hr" : "$(n-1)–$n hr" for n in 1:Nt];

# Finally, we are ready to plot.

plot_interval = Int(1hour / avg_output_interval)

# First we plot the mean profiles (wind speed, wind direction, potential temperature).

fig1 = Figure(size=(800, 400), fontsize=14)

ax1a = Axis(fig1[1, 1], xlabel="√(U² + V²) (m/s)", ylabel="z (m)",title="Horizontal wind speed")
ax1b = Axis(fig1[1, 2], xlabel="WD (° from N)", ylabel="z (m)",title="Wind direction")
ax1c = Axis(fig1[1, 3], xlabel="θ (K)", ylabel="z (m)",title="Potential temperature")

for n in 1:plot_interval:Nt
    lines!(ax1a, WS_mean_ts[n], color=colors[n], label=labels[n])
    lines!(ax1b, WD_mean_ts[n], color=colors[n])
    lines!(ax1c,  θ_mean_ts[n], color=colors[n])
end

linkyaxes!(ax1a, ax1b, ax1c)
hideydecorations!(ax1b, grid=false)
hideydecorations!(ax1c, grid=false)

Legend(fig1[1, 4], ax1a, "Time", framevisible=false)
fig1

# Next, we plot the velocity variances normalized by ``u_★^2``

fig2 = Figure(size=(800, 400), fontsize=14)

ax2a = Axis(fig2[1, 1], xlabel="⟨u′u′⟩ / u_★²", ylabel="z (m)", title="u variance")
ax2b = Axis(fig2[1, 2], xlabel="⟨v′v′⟩ / u_★²", ylabel="z (m)", title="v variance")
ax2c = Axis(fig2[1, 3], xlabel="⟨w′w′⟩ / u_★²", ylabel="z (m)", title="w variance")

for n in 1:plot_interval:Nt
    lines!(ax2a, uu_var_ts[n], color=colors[n], label=labels[n])
    lines!(ax2b, vv_var_ts[n], color=colors[n])
    lines!(ax2c, ww_var_ts[n], color=colors[n])
end

linkyaxes!(ax2a, ax2b, ax2c)
hideydecorations!(ax2b, grid=false)
hideydecorations!(ax2c, grid=false)

Legend(fig2[1, 4], ax2a, "Time", framevisible=false)
fig2

# Last, we plot the resolved and the SGS fluxes.

fig3 = Figure(size=(1200, 450), fontsize=14)

ax3a = Axis(fig3[1, 1], xlabel="τˣ / ρ₀u_★²", ylabel="z (m)", title="x-momentum flux")
ax3b = Axis(fig3[1, 2], xlabel="τʸ / ρ₀u_★²", ylabel="z (m)", title="y-momentum flux")
ax3c = Axis(fig3[1, 3], xlabel="Jᶿ (K m/s)", ylabel="z (m)", title="Potential temperature flux")

for n in 1:plot_interval:Nt
    lines!(ax3a, uw_res_ts[n] + uw_sgs_ts[n], color=colors[n], label=labels[n])
    lines!(ax3a, uw_res_ts[n], color=colors[n], linestyle=:dash)
    lines!(ax3a, uw_sgs_ts[n], color=colors[n], linestyle=:dot)

    lines!(ax3b, vw_res_ts[n] + vw_sgs_ts[n], color=colors[n])
    lines!(ax3b, vw_res_ts[n], color=colors[n], linestyle=:dash)
    lines!(ax3b, vw_sgs_ts[n], color=colors[n], linestyle=:dot)

    lines!(ax3c, θw_res_ts[n] + θw_sgs_ts[n], color=colors[n])
    lines!(ax3c, θw_res_ts[n], color=colors[n], linestyle=:dash)
    lines!(ax3c, θw_sgs_ts[n], color=colors[n], linestyle=:dot)
end

for ax in (ax3a, ax3b, ax3c)
    vlines!(ax, 0, color=:grey, linewidth=0.5)
end

## Legends: line style (inside panel a) and time (right)
style_entries = [LineElement(color=:black, linestyle=:solid),
                 LineElement(color=:black, linestyle=:dash),
                 LineElement(color=:black, linestyle=:dot)]
axislegend(ax3a, style_entries, ["total", "resolved", "SGS"],
           position=:lt, framevisible=false)

linkyaxes!(ax3a, ax3b, ax3c)
hideydecorations!(ax3b, grid=false)
hideydecorations!(ax3c, grid=false)

Legend(fig3[1, 4], ax3a, "Time", framevisible=false)
fig3
```

navidcy: I made all the post-processing and figures using Oceananigans abstract operations (eg for derivatives and multiplications), avoiding using Arrays. Then for the plots the OceananigansMakieExt deals with the vertical coordinates automatically.

ewquon: Thanks!

examples/neutral_atmospheric_boundary_layer.jl
ewquon: I think the only time we might want to have `w` on a staggered grid is if we're looking at profiles of `w` variance and we want accurate maxima. For fluxes, a decision about interpolating the horizontal velocity component or scalar versus interpolating `w` must be made. Where it matters most is near the surface, where the vertical velocity vanishes, so one could argue that `w` is the smoother quantity and the one that should be interpolated.

For this example, if we can simplify the script, I'm okay with doing more work behind the scenes.

ewquon:
I could change the example to Float32 following https://github.com/CliMA/Oceananigans.jl/pull/5491, that will save time building the docs.

giordano:
Yeah, that might make things simpler.  With #630 I at least tried to reduce the concurrent GPU jobs, but I don't think that's going to be sufficient with this example.  And the L4 doesn't have a great fp64 performance :cry:

giordano:
With the Oceananigans PR and using single precision, all examples are built within a finite time: https://numericalearth.github.io/BreezeDocumentation/previews/PR615/literated/neutral_atmospheric_boundary_layer/. But we'll need to wait for that PR to make it into a release, and then also revert https://github.com/NumericalEarth/Breeze.jl/pull/615/commits/6c27e7714ff99faf9c3e0b5d240acd546219dd86

examples/neutral_atmospheric_boundary_layer.jl
```diff
@@ -0,0 +1,408 @@
...
+ρθ_forcing = ρθ_sponge
+
+forcing = (; ρu=ρu_forcing, ρv=ρv_forcing, ρw=ρw_forcing, ρθ=ρθ_forcing)
+nothing #hide
+
+# ## Model setup
+
+advection = WENO(order=9)           # WENO(order=5), Centered(order=6) are dissipative
+
+closure = SmagorinskyLilly(C=0.18)  # Sullivan et al. (1994)
```
glwagner: I'd like to discuss this. This may (will?) have the effect of spurring users to try to set C=0.18 for model coefficient in their research scripts. Is that the impact we intend to have here? We should be intentional about it.

I think fiddling with closures on a case-by-case basis is something of a fool's errand unless it serves a very specific purpose, which I am not sure we are after in this example.

examples/neutral_atmospheric_boundary_layer.jl
```diff
@@ -0,0 +1,408 @@
...
+
+add_callback!(simulation, progress, IterationInterval(1000))
+
+# Output averaged products for calculating statistics
+
+avg_outputs_varlist = (;
+    θ, νₑ,
+    uu = u*u, vv = v*v, ww = w*w,
+    uw = u*w, vw = v*w, θw = θ*w,       # second-order moments for fluxes
+    uuw = u*u*w, vvw=v*v*w, www=w*w*w,  # third-order moments for turbulent transport term
```
glwagner:
```suggestion
    uuw = u*u*w, vvw=v*v*w, www=w^3,  # third-order moments for turbulent transport term
```

examples/neutral_atmospheric_boundary_layer.jl
```diff
@@ -0,0 +1,408 @@
...
+    return nothing
+end
+
+add_callback!(simulation, progress, IterationInterval(1000))
+
+# Output averaged products for calculating statistics
+
+avg_outputs_varlist = (;
+    θ, νₑ,
+    uu = u*u, vv = v*v, ww = w*w,
```
glwagner:
```suggestion
    uu = u^2, vv = v^2, ww = w^2,
```

examples/neutral_atmospheric_boundary_layer.jl
```diff
@@ -0,0 +1,419 @@
...
+
+# Define a colormap for each time.
+
+cmap = cgrad(:viridis)
+colors = [cmap[(n-1)/max(Nt-1, 1)] for n in 1:Nt]
+
+# Note that the AveragedTimeInterval schedule outputs the initial snapshot
+# followed by averaged snapshots at the requested output interval.
+smart_label(n) = prettytime(n * avg_output_interval)
+labels = [n == 1 ? "initial condition" : smart_label(n-1) for n in 1:Nt]
```
giordano, line 353: Just to avoid printing this array
```suggestion
labels = [n == 1 ? "initial condition" : smart_label(n-1) for n in 1:Nt]
nothing #hide
```

examples/neutral_atmospheric_boundary_layer.jl
ewquon: I'm not tied to any particular closure constant and see how this could be perceived as "recommended practice."

examples/neutral_atmospheric_boundary_layer.jl
```diff
@@ -0,0 +1,427 @@
...
+    θ, νₑ,
+    uu = u^2, vv = v^2, ww = w^2,
+    uw = u*w, vw = v*w, θw = θ*w,           # second-order moments for fluxes
+    uuw = u^2*w, vvw = v^2*w, www = w^3,    # third-order moments to calculate turbulent transport
+    νₑ³ = νₑ^3,                             # SGS dissipation — note: |S̄|² = νₑ² / (Cₛ Δ)⁴) with Smagorinsky model
+)
+outputs = merge(model.velocities, model.tracers, avg_outputs_varlist)
+
+# After computing the output quantity and prior to calculating the slab average,
+# staggered quantities are averaged to cell centers
```
navidcy:
```suggestion
# staggered quantities are interpolated to cell centers.
```

examples/neutral_atmospheric_boundary_layer.jl
```diff
@@ -0,0 +1,427 @@
...
+
+    uw_sgs_ts[n] .= -ρᵣ * νₑ_n * ∂z_u / (ρ₀ * u★^2)
+    vw_sgs_ts[n] .= -ρᵣ * νₑ_n * ∂z_v / (ρ₀ * u★^2)
+    θw_sgs_ts[n] .= -νₑ_n * ∂z_θ / closure.Pr
+end
+
+# Define a colormap for each time.
+
+cmap = cgrad(:viridis)
+colors = [cmap[(n-1)/max(Nt-1, 1)] for n in 1:Nt]
```
giordano, line 347: Hide more stuff
```suggestion
colors = [cmap[(n-1)/max(Nt-1, 1)] for n in 1:Nt]
nothing #hide
```

giordano:
Uhm, the potential temperature flux
image
is _very_ different from https://github.com/NumericalEarth/Breeze.jl/pull/615#issue-4219731348

ewquon:
Problem is that the halos aren't filled so `∂z` gives bogus values on the bottom boundary.

navidcy:
> Problem is that the halos aren't filled so `∂z` gives bogus values on the bottom boundary.

Oh, that's an easy fix. We just need to save the halos!

navidcy:
> Problem is that the halos aren't filled so `∂z` gives bogus values on the bottom boundary.

Hm... Why are halos not saved? The default option for JLD2Writer is that `with_halos = true`.

https://github.com/CliMA/Oceananigans.jl/blob/10c0f41db2245e3c753fc51248eb9d81b653505b/src/OutputWriters/jld2_writer.jl#L168

navidcy:
Perhaps it's not the halos, but rather the boundary conditions on θ that are different in the outputs and the actual model?

navidcy:
Or perhaps we can compute the derivatives on the fly and output them. I'll try that.

ewquon:
The halo values are all 0. Unless it's a FieldTimeSeries loading issue.

navidcy:
> The halo values are all 0. Unless it's a FieldTimeSeries loading issue.

I think it's because we take the Average...? So for some reason the output writer doesn't apply the operation in the halo region and save the Average halo?

I'm trying (with https://github.com/NumericalEarth/Breeze.jl/pull/615/commits/8286c794341db3446b0497cfafb060dde59444e5) to take the derivative before the Average; let's see.

navidcy:
Now docs get a bunch of `newton_div`-related errors. Is this related to https://github.com/CliMA/Oceananigans.jl/pull/5140?

cc @glwagner

giordano:
> Is this related to https://github.com/CliMA/Oceananigans.jl/pull/5140?

Most likely yes

navidcy:
Was it a breaking change?
I don't know how to fix it. Actually `newton_div` is never called in Breeze so perhaps this is an Oceananigans bug that isn't captured in the Oceananigans CI?

ewquon:
It's caught by the unit tests

giordano:
Yeah, it's failing in https://github.com/CliMA/Oceananigans.jl/pull/5491, which is what we're using here 😅

ewquon:
Hmm, I tried running RICO separately and it failed for the same reason, this is the fix that Claude came up with https://github.com/NumericalEarth/Breeze.jl/pull/631

ewquon:
Now waiting on https://github.com/JuliaRegistries/General/pull/153245

examples/neutral_atmospheric_boundary_layer.jl
```diff
@@ -0,0 +1,434 @@
...
+# Note: Higher-order moments are computed at the location of the _first_ field.
+# E.g., u * w results in a BinaryOperation at (Face, Center, Center).
+avg_outputs_varlist = (;
+    θ, νₑ,
+    uu = u^2, vv = v^2, ww = w^2,
+    uw = u*w, vw = v*w, θw = θ*w,           # second-order moments for fluxes
+    uuw = u^2*w, vvw = v^2*w, www = w^3,    # third-order moments to calculate turbulent transport
+    νₑ³ = νₑ^3,                             # SGS dissipation — note: |S̄|² = νₑ² / (Cₛ Δ)⁴) with Smagorinsky model
+)
+outputs = merge(model.velocities, model.tracers, avg_outputs_varlist)
```
giordano, line 213: Maybe we hide the value of `outputs`?  [It's very long](https://numericalearth.github.io/BreezeDocumentation/previews/PR615/literated/neutral_atmospheric_boundary_layer/#Horizontal-averaging), and I'm not sure it's particularly interesting
```suggestion
outputs = merge(model.velocities, model.tracers, avg_outputs_varlist)
nothing #hide
```

navidcy:
The $J_θ$ looks good now. But the last plot doesn't show any difference between total and resolved. Is something just zeros by accident?

giordano:
https://numericalearth.github.io/BreezeDocumentation/previews/PR615/literated/neutral_atmospheric_boundary_layer/ looking good now?

navidcy:
Nice job!

PR #622, opened 2026-04-08
Acoustic substepping cleanup: typed AcousticDampingStrategy + new default + BW example

glwagner:
## Summary

This PR refactors the compressible split-explicit acoustic substepper and its documentation. The main changes are:

- `SplitExplicitTimeDiscretization` now carries a typed acoustic-substep configuration: adaptive `acoustic_cfl`, `forward_weight`, a `damping` strategy, optional `UpperSponge`, and a `substep_distribution`.
- The default split-explicit timestepper is `:AcousticRungeKutta3`, using a Wicker-Skamarock RK3 outer loop with acoustic inner substeps.
- Reference-state handling was tightened for split-explicit compressible dynamics, including discrete hydrostatic balance and pressure seeding during model materialization.
- Baroclinic-wave and inertia-gravity-wave docs/examples were updated for the current API and generated documentation.

## Split-Explicit API

| field on `SplitExplicitTimeDiscretization` | supported values |
|---|---|
| `damping` | `ThermalDivergenceDamping()` by default, or `NoDivergenceDamping()` |
| `sponge` | `nothing` by default, or `UpperSponge(damping_rate, depth, ramp=CubicRamp())` |
| `substep_distribution` | `ProportionalSubsteps()` by default, or `MonolithicFirstStage()` |

Example:

```julia
sponge = UpperSponge(damping_rate = 0.2, depth = 5kilometers)
damping = ThermalDivergenceDamping(coefficient = 0.1)

time_discretization = SplitExplicitTimeDiscretization(; damping, sponge)
dynamics = CompressibleDynamics(time_discretization; reference_potential_temperature = θᵣ)
```

`AbstractRamp` controls the sponge profile. Built-in ramps are `CubicRamp` (default), `Sin2Ramp`, and `LinearRamp`.

## Reference-State And Initialization Changes

The compressible reference-state path now builds reference pressure, density, and Exner fields so that the discrete vertical pressure-gradient and buoyancy terms cancel for rest-state split-explicit tests. Model materialization also seeds the diagnostic pressure from the reference state or surface pressure before the first thermodynamic update.

Terrain-aware compressible dynamics continue to use terrain reference pressure and density fields for pressure-gradient corrections.

## Documentation And Examples

- `examples/baroclinic_wave.jl` now relies on the dynamics-selected `:AcousticRungeKutta3` timestepper instead of passing an explicit `timestepper` keyword.
- `examples/inertia_gravity_wave.jl` and `docs/src/compressible_dynamics.md` document the split-explicit API and current damping/sponge options.
- Missing citations, media links, and `SplitExplicitTimeDiscretization` cross-references were cleaned up.

## Tests

Relevant coverage includes:

- `test/acoustic_substepping.jl` for the split-explicit constructor/API, default timestepper, acoustic substep logic, damping, and sponge behavior.
- `test/substepper_structural.jl` for structural properties of the acoustic update.
- `test/substepper_rest_state.jl` for rest-atmosphere preservation.
- `test/latitude_longitude_grid.jl`, `test/forcing_and_boundary_conditions.jl`, and `test/geostrophic_subsidence_forcings.jl` for integration coverage around grids, forcing, and boundary-condition setup.

## Scope Note

Earlier versions of this branch experimented with additional damping and advection ideas. Those are not part of the current PR. Review should focus on the split-explicit acoustic substepper cleanup, the reference-state fixes needed by that substepper, and the associated docs/tests.


examples/baroclinic_wave.jl
```diff
@@ -0,0 +1,371 @@
...
+#
+# This is roughly **110× larger** than the acoustic-CFL-limited Δt = 2 s
+# the fully explicit solver requires for the same grid. We run for 14 days
+# to observe baroclinic wave growth — the instability becomes visible
+# around day 4 and develops explosive cyclogenesis near day 8.
+
+Δt = 225seconds
+stop_time = 14days
+
+simulation = Simulation(model; Δt, stop_time)
```
giordano, line 258:
```suggestion
simulation = Simulation(model; Δt, stop_time)
Oceananigans.Diagnostics.erroring_NaNChecker!(simulation)
```

examples/moist_baroclinic_wave.jl
```diff
@@ -0,0 +1,416 @@
...
+# within ~25 outer steps, at ``Δt = 30`` s a slow ``2 Δt`` oscillation grows,
+# and at ``Δt = 20`` s the integration is rock-stable with ``\\max |w|`` of a
+# few cm/s. This is still ~10× larger than the explicit acoustic-CFL ``Δt`` of
+# the original moist BW example. (Future work: subcycle the microphysics
+# inside one outer Δt so the dynamics CFL is the only limit.)
+
+Δt = 20.0
+stop_time = 15days
+
+simulation = Simulation(model; Δt, stop_time)
```
giordano:
```suggestion
simulation = Simulation(model; Δt, stop_time)
Oceananigans.Diagnostics.erroring_NaNChecker!(simulation)
```

examples/moist_baroclinic_wave.jl
giordano, line 1: This file isn't added to `docs/make.jl`, is that intentional?

examples/baroclinic_wave.jl
giordano: https://github.com/NumericalEarth/Breeze.jl/actions/runs/24930301334/job/73006849544?pr=622#step:7:112
````
ERROR: LoadError: LoadError: time = 90000.0, iteration = 400: NaN found in field ρ. Aborting simulation.
in expression starting at /__w/Breeze.jl/Breeze.jl/docs/src/literated/baroclinic_wave.md:1
when executing the following code block from inputfile `/__w/Breeze.jl/Breeze.jl/examples/baroclinic_wave.jl`
```julia
run!(simulation)
```
````
Yay for the NaNCehcker!

File: test/substepper_structural.jl [hunk 3]
```diff
@@ -0,0 +1,535 @@
...
+    δτ_new = 0.5  # arbitrary
+
+    b₁ = @allowscalar get_coefficient(1, 1, 1, grid, AcousticTridiagDiagonal(),
+                                      nothing, ZDirection(),
+                                      Π⁰, θ⁰, γRᵈ, g, δτ_new)
+    c₁ = @allowscalar get_coefficient(1, 1, 1, grid, AcousticTridiagUpper(),
+                                      nothing, ZDirection(),
+                                      Π⁰, θ⁰, γRᵈ, g, δτ_new)
+
+    @info @sprintf("[S1] b[1] = %.6f, c[1] = %.6e (must be 1.0 and 0.0)", b₁, c₁)
```
giordano: Are these prints really necessary?  I'm not a fan of printing during tests, if there's any useful information that'd should be tested, not printed.  Same applies to all other new tests which print stuff

glwagner: probably not. I agree, printing during testing is annoying. Learned that lesson with Oceananigans.

src/Thermodynamics/reference_states.jl
```diff
@@ -505,36 +617,130 @@ function ExnerReferenceState(grid, constants=ThermodynamicConstants(eltype(grid)
     Rᵈ = dry_air_gas_constant(constants)
     cᵖᵈ = constants.dry_air.heat_capacity
     κ = Rᵈ / cᵖᵈ
-    g = constants.gravitational_acceleration
-    Nz = size(grid, 3)
-    loc = (nothing, nothing, Center())
+    g Tᵣstants.gravitational_acceleration
+    NzTᵣze(grid, 3)
+Tᵣ
```
giordano: Something went wrong here

glwagner: argh. tried to search and replace

giordano: I reverted 511a8759ecbe695899c066adef10471f3c156613 entirely, I think many things went wrong with that change

examples/baroclinic_wave.jl
```diff
@@ -211,19 +211,19 @@ end
...
 # slow tendencies see only perturbations from the background.
 # `HydrostaticSphericalCoriolis` retains the traditional ``f = 2Ω \sin φ``
 # Coriolis terms.
 
 coriolis = HydrostaticSphericalCoriolis(rotation_rate=Ω)
 
-T₀ᵣ= 250.0
-θᵣ(z) = T₀ᵣ* exp(g * z / (cᵖᵈ * T₀_ref))
+T₀_ref = 250.0
+θ_ref(z) = T₀_ref * exp(g * z / (cᵖᵈ * T₀_ref))
```
glwagner: 😢

glwagner: @giordano are you ok if I convert these back to subscript "r" in accordance with our notation conventions?

giordano: Yes, but the previous attempt had multiple issues in multiple files, that's why I reverted it

ext/BreezeReactantExt/Timesteppers.jl
```diff
@@ -1,10 +1,11 @@
...
-using Oceananigans.TimeSteppers: TimeSteppers as OceananigansTimeSteppers, time_step!, update_state!
+using Oceananigans.TimeSteppers: TimeSteppers as OceananigansTimeSteppers, update_state!
 using Breeze.TimeSteppers: SSPRungeKutta3
 
 function OceananigansTimeSteppers.first_time_step!(model::AtmosphereModel{<:Any, <:Any, <:ReactantState, <:SSPRungeKutta3}, Δt)
     initialize!(model)
     update_state!(model)
-    time_step!(model, Δt)
+    # time_step!(model, Δt)
+    # Don't call time_step! temporarily for GB purposes
```
glwagner:
```suggestion
    time_step!(model, Δt)
```

src/CompressibleEquations/time_discretizations.jl
```diff
@@ -1,58 +1,475 @@
...
+        throw(ArgumentError("`sponge` must be `nothing` or an `UpperSponge`"))
+
     return SplitExplicitTimeDiscretization(substeps,
                                            forward_weight,
-                                           divergence_damping_coefficient,
-                                           acoustic_damping_coefficient)
+                                           damping,
+                                           sponge,
+                                           substep_distribution)
 end
```
giordano: I just realised #678 conflicts with this PR, but this code has the same problem as what that PR was trying to solve: it's hard to enforce the same float type everywhere.  For example running on Metal GPU it's hard to consistently use `Float32` everywhere.

glwagner: mm yes! We need to add an `FT` option to the constructor. Is that what you mean?

giordano: Yes, and make sure that all fields have the same floating point type.

src/CompressibleEquations/acoustic_substepping.jl
```diff
@@ -1,1203 +1,1457 @@
...
-                    substepper.virtual_potential_temperature, grid, ϰᵃᶜ, cᵖ)
-        end
 
-        launch!(arch, grid, :xyz, _update_pressure_and_average!,
-                substepper.exner_perturbation, substepper.filtered_exner_perturbation, substepper.previous_exner_perturbation,
-                u, v, w, ū,
-                grid, ϰᵈⁱ, 1 / Nτ)
-    end
+    g   = convert(eltype(grid), model.thermodynamic_constants.gravitational_acceleration)
+    FT  = eltype(grid)
```
giordano:
```suggestion
    FT  = eltype(grid)
    g   = convert(FT, model.thermodynamic_constants.gravitational_acceleration)
```

src/CompressibleEquations/acoustic_substepping.jl
```diff
@@ -1,1203 +1,1457 @@
...
 
-function _set_exner_reference!(substepper, model, ::Nothing, pˢᵗ, κ)
-    grid = model.grid
-    arch = architecture(grid)
+# ProportionalSubsteps: every stage uses Δτ = Δt/N, Nτ = round(β·N).
+@inline function stage_substep_count_and_size(::ProportionalSubsteps, β_stage, Δt, N)
+    Δτ = Δt / N
+    Nτ = max(1, round(Int, β_stage * N))
+    return Nτ, Δτ
+end
```
giordano, line 531: I think it'd be better to take also the grid as argument and do
```
FT = eltype(grid)
Δτ = FT(Δt) / N
```
because it's generally more accurate to do the conversion before the division.  Same for the other methods below.  And then you can replace all the `FT(Δτ)` -> `Δτ`

docs/src/compressible_dynamics.md
```diff
@@ -1,413 +1,448 @@
 # [Compressible dynamics](@id Compressible-section)
 
-[`CompressibleDynamics`](@ref) solves the fully compressible Euler equations with prognostic density ``ρ``.
-This formulation retains acoustic waves and is suitable for problems where full compressibility is important.
+[`CompressibleDynamics`](@ref) solves the fully compressible Euler equations with prognostic
+density ``ρ``. The formulation retains acoustic waves and is suitable for problems where full
```
ewquon:
```suggestion
total density ``ρ`` (including dry air, vapor, and condensate). The formulation retains acoustic waves and is suitable for problems where full
```

docs/src/compressible_dynamics.md
```diff
@@ -1,413 +1,448 @@
...
-    pressure gradient seen by the initial velocities is inconsistent with the pressure field,
-    generating spurious acoustic oscillations that grow over multiple time steps.
-
-!!! warning "WS-RK3: use θⁿ as the base for θ evolution"
-    The slow ``\theta`` tendency must be applied to ``\theta^n`` from ``U^0``, **not** to
-    ``\bar{\theta}_v`` from the evaluation state. Using the evaluation state's ``\theta``
-    double-counts the ``\theta`` change from earlier stages (since ``\theta(U^*) = \theta^n +
-    \beta_1 \Delta t \, \dot{\theta}^s`` already includes the stage 1 contribution), producing
-    an ``O(\Delta t)`` error per time step that causes instability at larger ``\Delta t``.
+Breeze can also fold the vertical component into the column tridiag by setting
```
ewquon:
```suggestion
This horizontal divergence damping is applied by default. Breeze can also fold the vertical component into the column tridiag by setting
```

docs/src/compressible_dynamics.md
```diff
@@ -1,413 +1,448 @@
...
 
 ```math
 \begin{aligned}
-\pi^n &= \left[ \frac{R^d \, (\rho\theta)^n}{p^{st}} \right]^{R/c_v} \\
-\pi_{\text{new}} &= \pi^n + \Delta\pi', \qquad \Delta\pi' = \pi'_{\text{final}} - \pi'_{\text{initial}} \\
-\rho\theta_{\text{new}} &= \frac{p^{st}}{R^d} \, \pi_{\text{new}}^{\,c_v / R} \\
-\theta_{\text{new}} &= \theta^n + \beta \, \Delta t \, \dot{\theta}^s \\
-\rho_{\text{new}} &= \rho\theta_{\text{new}} \, / \, \theta_{\text{new}}
+Δ(ρu)' &= - γ_h \, ∂_x D_τ , \\
+Δ(ρv)' &= - γ_h \, ∂_y D_τ .
```
ewquon:
```suggestion
Δ(ρu)' &= - γ_x \, ∂_x D_τ , \\
Δ(ρv)' &= - γ_y \, ∂_y D_τ .
```

docs/src/compressible_dynamics.md
```diff
@@ -1,413 +1,448 @@
...
+``∂(ρθ)'``. After the tridiag is solved the perturbations of ``ρ'`` and ``(ρθ)'`` are
+recovered by back-substitution.
 
-**(C) Filter and accumulate:**
+The off-centering parameter ``ω = 1/2`` is classical centered Crank–Nicolson — neutrally
+stable for the linearized inviscid system but susceptible to amplification of distributed
+floating-point noise through the non-normal substep operator (see
+[Stability analysis](@ref stability-analysis)). The default ``ω = 0.65`` adds modest
+dissipation; the dimensionless parameter ``ε = 2ω - 1 = 0.3`` quantifies the deviation from
+centered.
```
ewquon:
```suggestion
The off-centering parameter ``ω = 1/2`` is classical centered Crank–Nicolson — neutrally
stable for the linearized inviscid system but susceptible to amplification of distributed
floating-point noise through the non-normal substep operator (see
[Stability analysis](@ref stability-analysis)).
A fully implicit backward Euler scheme is obtained with ``ω = 1`` and offers the most dissipation.
The default ``ω = 0.65`` adds modest dissipation; the dimensionless parameter ``ε = 2ω - 1 = 0.3`` quantifies the deviation from centered.
```

ext/BreezeCloudMicrophysicsExt/one_moment_microphysics.jl
```diff
@@ -842,7 +842,7 @@ end
                      qᶜˡ, qʳ, ρ)
 
     # Ice → snow autoconversion
-    Sᵃᶜⁿᵛⁱˢ = conv_q_icl_to_q_sno_no_supersat(categories.snow.acnv1M, qᶜⁱ, true)
+    Sᵃᶜⁿᵛⁱˢ = conv_q_icl_to_q_sno_no_supersat(categories.snow.acnv1M, qᶜⁱ, false)
```
ewquon: What's the rationale for switching off the smooth form?

src/AtmosphereModels/dynamics_interface.jl
```diff
@@ -66,7 +66,7 @@ $(TYPEDSIGNATURES)
 Return the default timestepper symbol for the given dynamics.
 
 For anelastic dynamics, returns `:SSPRungeKutta3`.
```
ewquon:
```suggestion
For anelastic dynamics or compressible dynamics with explicit timestepping, returns `:SSPRungeKutta3`.
```

src/AtmosphereModels/dynamics_interface.jl
```diff
@@ -66,7 +66,7 @@ $(TYPEDSIGNATURES)
...
-For compressible dynamics, returns `:AcousticSSPRungeKutta3` (acoustic substepping).
+For compressible dynamics, returns `:AcousticRungeKutta3` (acoustic substepping).
```
ewquon:
```suggestion
For compressible dynamics with acoustic substepping, returns `:AcousticRungeKutta3`.
```

src/AtmosphereModels/dynamics_kernel_functions.jl
```diff
@@ -55,13 +63,24 @@ end
...
+@inline x_momentum_flux_divergence(i, j, k, grid, advection, momentum, velocities, dynamics) =
+    div_𝐯u(i, j, k, grid, advection, momentum, velocities.u) +
+    U_dot_∇u_metric(i, j, k, grid, advection, momentum, velocities)
+@inline y_momentum_flux_divergence(i, j, k, grid, advection, momentum, velocities, dynamics) =
+    div_𝐯v(i, j, k, grid, advection, momentum, velocities.v) +
+    U_dot_∇v_metric(i, j, k, grid, advection, momentum, velocities)
+@inline z_momentum_flux_divergence(i, j, k, grid, advection, momentum, velocities, dynamics) =
+    div_𝐯w(i, j, k, grid, advection, momentum, velocities.w) +
+    U_dot_∇w_metric(i, j, k, grid, advection, momentum, velocities)
+
```
ewquon: Nit: It would read more naturally if the `*_momentum_flux_divergence` definitions all came before or after the tendencies, not between `x_momentum_tendency` and `y_momentum_tendency`

src/AtmosphereModels/dynamics_interface.jl
ewquon: Consistent with what's in compressible_dynamics.jl

src/TimeSteppers/acoustic_runge_kutta_3.jl
```diff
@@ -90,52 +83,65 @@ function AcousticRungeKutta3(grid, prognostic_fields;
...
+    β₃ = FT(β[3])
 
     U⁰ = map(similar, prognostic_fields)
     U0 = typeof(U⁰)
 
-    substepper = AcousticSubstepper(grid, dynamics.time_discretization)
+    substepper = AcousticSubstepper(grid, dynamics.time_discretization;
+                                    prognostic_momentum = (ρu = prognostic_fields.ρu,
+                                                            ρv = prognostic_fields.ρv,
+                                                            ρw = prognostic_fields.ρw))
```
ewquon:
```suggestion
                                    prognostic_momentum = (ρu = prognostic_fields.ρu,
                                                           ρv = prognostic_fields.ρv,
                                                           ρw = prognostic_fields.ρw))
```

src/TimeSteppers/acoustic_runge_kutta_3.jl
```diff
@@ -163,70 +166,85 @@ end
...
-    # Adjust final time-step
     corrected_Δt = time_difference_seconds(tⁿ⁺¹, model.clock.time)
     tick_stage!(model.clock, corrected_Δt, Δt)
 
     step_closure_prognostics!(model.closure_fields, model.closure, model, Δt)
 
+    # Operator-split microphysics: applied once per time step on the post-RK state
+    # (rather than once per stage from update_state!), so that the full Δt of
+    # autoconversion / accretion / condensation / sedimentation is applied exactly once.
+    microphysics_model_update!(model.microphysics, model)
```
ewquon: If I understand correctly, whether or not the microphysical state is actually updated here with the full Δt will depend on the selected MP scheme, so this comment is potentially misleading. I would put something in the docs as well.

src/CompressibleEquations/acoustic_substepping.jl
glwagner: ok!

glwagner: or FT directly?

ext/BreezeCloudMicrophysicsExt/one_moment_microphysics.jl
glwagner: Weird. I think we should revert this.

src/TimeSteppers/acoustic_runge_kutta_3.jl
glwagner: Hmm, yes I think by "applied" we merely mean that the function is called. But as you point out that may do nothing (for most schemes). I do think the language can be improved.

src/AtmosphereModels/dynamics_interface.jl
```diff
@@ -65,8 +65,8 @@ $(TYPEDSIGNATURES)
 
 Return the default timestepper symbol for the given dynamics.
 
-For anelastic dynamics, returns `:SSPRungeKutta3`.
-For compressible dynamics, returns `:AcousticSSPRungeKutta3` (acoustic substepping).
+For anelastic dynamics or compressible dynamics with explicit timestepping, returns `:SSPRungeKutta3`.
+For compressible dynamics, returns `:AcousticRungeKutta3` (acoustic substepping).
```
kaiyuan-cheng: Looks like the acoustic solver is using Wicker-Skamarock RK3, and the SSP RK3 is not supported, am I correct?

test/latitude_longitude_grid.jl
```diff
@@ -0,0 +1,151 @@
...
+    @test N >= 1
+
+    # On a LatitudeLongitudeGrid, the smallest Δx is at the highest latitude.
+    # At 85°, Δx ≈ R cos(85°) Δλ ≈ 6371e3 * cos(85°*π/180) * (360/36)*π/180 ≈ 61 km
+    # Needs more substeps than at the equator.
+    Δx_equator = 6371e3 * (360 / 36) * π / 180  # ≈ 1113 km
+    Δx_85 = Δx_equator * cos(85 * π / 180) # ≈ 97 km
+
+    # Should need more substeps for the high-latitude grid since Δx is smaller
+    @test N >= 1
```
ewquon: This test appears to be missing a few steps

src/TimeSteppers/acoustic_substep_helpers.jl
```diff
@@ -0,0 +1,155 @@
...
+#####
+
+"""
+$(TYPEDSIGNATURES)
+
+Update non-acoustic scalar fields (moisture, tracers) using the given
+kernel. Iterates over prognostic fields, skipping the first 5
+(``ρ, ρu, ρv, ρw, ρθ``) which are handled by the acoustic substep loop.
+"""
+function acoustic_scalar_substep!(model, kernel!, Δt_implicit, kernel_args...)
```
kaiyuan-cheng: The term `acoustic` in `acoustic_scalar_substep` seems misleading, as tracers are not being updated during the acoustic substeps.

src/Thermodynamics/reference_states.jl
```diff
@@ -440,28 +444,133 @@ end
 
 Base.show(io::IO, ref::ExnerReferenceState) = print(io, summary(ref))
 
-@kernel function _compute_exner_reference!(π₀, pᵣ, ρᵣ, θ₀, grid, Nz, π₀_surface, pˢᵗ, cᵖᵈ, κ, Rᵈ, g)
+@kernel function _compute_isothermal_reference!(π₀, pᵣ, ρᵣ, θᵣ, grid, Nz, p₀, pˢᵗ, κ, Rᵈ, g, T₀)
     _ = @index(Global)
 
-    # Initialize π₀ at the bottom from the surface pressure
-    @inbounds π₀[1, 1, 1] = π₀_surface
+    # Discrete-balance recurrence (Standard S3, audit E2). Enforces
```
kaiyuan-cheng: I guess those tags (S3 and E2) are no longer needed.

test/substepper_structural.jl
```diff
@@ -0,0 +1,579 @@
...
+    update_state!(model)
+end
+
+@testset "Substepper structural correctness" begin
+
+#####
+##### S1 — Boundary-row of the column tridiag is trivial
+#####
+##### Solver row k = 1 must have b[1] = 1, c[1] = 0 so that the
+##### post-solve gives μw[1] = 0 (impenetrability at the rigid
```
ewquon: μ isn't meaningful in a Breeze context, I think it's a remnant from borrowed WRF code

src/Thermodynamics/reference_states.jl
```diff
@@ -440,28 +444,133 @@ end
...
+    #   (p[k] − p[k − 1]) / Δz_face[k] + g · (ρ[k] + ρ[k − 1]) / 2 = 0
+    # at every interior face k = 2..Nz, with ρ[k] = p[k] / (Rᵈ T[k]) and
+    # T[k] = θ̄[k] · Π[k] = θ̄[k] · (p[k]/pˢᵗ)^κ. Substituting gives a
+    # nonlinear equation in p[k] solved by Newton iteration.
+    #
+    # The previous implementation (MPAS-style up-then-down Π integration
+    # using `ΔΠ = -g Δz_face / (cᵖᵈ θ_face)`) satisfies the *continuous*
+    # hydrostatic equation but not the substepper's discrete operator,
+    # leaving a residual ~1e-3 N/m³ that seeds a factor-2 acoustic
+    # instability at production Δt. See test/substepper_validation/REPORT.md
```
ewquon: this report has not been committed

src/AtmosphereModels/dynamics_interface.jl
glwagner: thats right. It is possible to do substepping with SSPRK3, but it is slightly less stable and more expensive, so we have not implemented it yet

src/TimeSteppers/acoustic_substep_helpers.jl
glwagner: true, the substepping resolves acoustic dynamics for the primary variables, but is just regular substepping for everything else

test/substepper_structural.jl
glwagner: yes although a previous implementation may have had mu 🙉

src/AtmosphereModels/dynamics_kernel_functions.jl
glwagner: good idea

src/CompressibleEquations/time_discretizations.jl
```diff
@@ -1,60 +1,512 @@
...
+"""
+# `damping` may be a single `AcousticDampingStrategy` or a `Tuple` of them.
+# When it's a tuple, each strategy is applied in order at every substep:
+# explicit corrections accumulate (each adds to `momentum_perturbation`),
+# and implicit-vertical-damping factors sum into the column tridiag. Useful
+# for combining divergence damping with upper sponge or vorticity damping.
+"""
+$(TYPEDEF)
+
+FV3-style 2nd-order vorticity damping inside the substepper inner loop.
```
kaiyuan-cheng: That’s interesting. I don't think WRF uses vorticity damping, and I’m not sure if MPAS does either. Is vorticity damping necessary because we’re solving the momentum equation in vector-invariant form? BTW, can we change the order of the vorticity damping? I think the vorticity damping in FV3 is at least 4th order for most applications. 2nd order might be too diffusive.

glwagner: This "damping strategy" is actually substepper specific (ie it only applies to acoustic perturbations, and is applied over the acoustic substeps). It's probably pointless to have vorticity damping (this was added in a very exhaustive search for any way to damp the "spurious mode"). Perhaps we should delete it?

It is possible to add slow vorticity damping on the outer step through a standard closure though. And we do support biharmonic for that.

kaiyuan-cheng (approved):
Looks good. Great work!

src/CompressibleEquations/time_discretizations.jl
kaiyuan-cheng: Yes, it is probably not useful here.

src/CompressibleEquations/time_discretizations.jl
```diff
@@ -1,60 +1,512 @@
...
+``D``. If `damp_vertical = true`, the vertical component is folded
+implicitly into the column tridiag as a Laplacian on ``(ρw)′``. By
+default, `damp_vertical = false` and vertical acoustic damping comes from
+the off-centered implicit solve.
+
+Per-substep momentum correction (Klemp, Skamarock & Ha 2018 eq. 36, MPAS form):
+
+```math
+Δ(ρu)′ = -γ · ∂_x D , \\quad
+Δ(ρv)′ = -γ · ∂_y D .
```
ewquon, line 162:
```suggestion
Δ(ρu)′ = -γ_x · ∂_x D , \\quad
Δ(ρv)′ = -γ_y · ∂_y D .
```

Should check for consistency between this source file and `compressible_dynamics.md`. Claude found notation differences and missing features not documented: `VorticityDamping`, `UpperSponge`, sponge profile shapes.

Also, `damping` is allowed to be a tuple of strategies applied in sequence per substep and is independent from user-specified sponge layers. Are these experimental features?

src/TimeSteppers/acoustic_runge_kutta_3.jl
```diff
@@ -163,70 +166,84 @@ end
...
+##### `freeze_linearization_state!` seeded the field with `model.velocities`
+##### at outer-step start.
+#####
+
+function AtmosphereModels.transport_velocities(model::AtmosphereModel{<:Any, <:Any, <:Any, <:AcousticRungeKutta3})
+    sub = model.timestepper.substepper
+    return (u = sub.time_averaged_velocities.u,
+            v = sub.time_averaged_velocities.v,
+            w = sub.time_averaged_velocities.w)
+end
```
giordano: This method is ambiguous with https://github.com/NumericalEarth/Breeze.jl/blob/c126a4c42ae1bacb2ea6823260e1c76504d9a48a/src/CompressibleEquations/terrain_compressible_physics.jl#L129-L134
The error message suggests a possible fix
```
│ Ambiguity #1
│ transport_velocities(model::Breeze.AtmosphereModels.AtmosphereModel{<:Breeze.CompressibleEquations.CompressibleDynamics{<:Any, <:Any, <:Any, <:Any, <:Any, <:Breeze.TerrainFollowingDiscretization.TerrainMetrics}}) @ Breeze.CompressibleEquations /__w/Breeze.jl/Breeze.jl/src/CompressibleEquations/terrain_compressible_physics.jl:129
│ transport_velocities(model::Breeze.AtmosphereModels.AtmosphereModel{<:Any, <:Any, <:Any, <:Breeze.TimeSteppers.AcousticRungeKutta3}) @ Breeze.TimeSteppers /__w/Breeze.jl/Breeze.jl/src/TimeSteppers/acoustic_runge_kutta_3.jl:244
│ 
│ Possible fix, define
│   transport_velocities(::Breeze.AtmosphereModels.AtmosphereModel{<:Breeze.CompressibleEquations.CompressibleDynamics{<:Any, <:Any, <:Any, <:Any, <:Any, <:Breeze.TerrainFollowingDiscretization.TerrainMetrics}, <:Any, <:Any, <:Breeze.TimeSteppers.AcousticRungeKutta3})
└ 
```
but another possible fix is to make the signature one of the two methods tighter and more accurate, so that it doesn't overlap with the other.

src/CompressibleEquations/time_discretizations.jl
```diff
@@ -1,60 +1,512 @@
...
+$(TYPEDEF)
+
+Acoustic divergence damping that uses the **(ρθ)′ tendency as a
+discrete proxy for the momentum divergence**. From the linearized
+ρθ-continuity equation
+``\\partial_t (ρθ)' + \\nabla\\cdot(ρθ^0 u') = 0``, the per-substep
+quantity
+
+```math
+D \\equiv \\frac{(ρθ)' - (ρθ)'_\\mathrm{old}}{θ^0}
```
ewquon:
```suggestion
``\\partial_t (ρθ)' + \\nabla\\cdot(ρθ^L u') = 0``, the per-substep
quantity

```math
D \\equiv \\frac{(ρθ)' - (ρθ)'_\\mathrm{old}}{θ^L}
```

Notation differs from acoustic_substepping.jl and compressible_dynamics.md

src/CompressibleEquations/acoustic_substepping.jl
```diff
@@ -1,1205 +1,1527 @@
...
+##### cached per stage; ρᴸ, ρθᴸ, pᴸ, ρuᴸ, ρvᴸ, ρwᴸ are read live from the
+##### model since the substep loop never mutates those fields.
+#####
+##### The linearized perturbation equations integrated by the substep loop:
+#####
+#####   ∂t ρ′    +     ∇·((ρu)′, (ρv)′, (ρw)′)       = Gˢρ
+#####   ∂t (ρθ)′ +     ∇·(θᴸ · ((ρu)′, (ρv)′, (ρw)′)) = Gˢρθ
+#####   ∂t (ρu)′ + ∂x pᴸ + ∂x(Cᴸ (ρθ)′)             = Gˢρu
+#####   ∂t (ρv)′ + ∂y pᴸ + ∂y(Cᴸ (ρθ)′)             = Gˢρv
+#####   ∂t (ρw)′ +          ∂z(Cᴸ (ρθ)′) + g · ρ′   = Gˢρw
```
ewquon:
```suggestion
#####   ∂t ρ′    +     ∇·((ρu)′, (ρv)′, (ρw)′)        = Gˢρ
#####   ∂t (ρθ)′ +     ∇·(θᴸ · ((ρu)′, (ρv)′, (ρw)′)) = Gˢρθ
#####   ∂t (ρu)′ + ∂x pᴸ + ∂x(Cᴸ (ρθ)′)               = Gˢρu
#####   ∂t (ρv)′ + ∂y pᴸ + ∂y(Cᴸ (ρθ)′)               = Gˢρv
#####   ∂t (ρw)′ +         ∂z(Cᴸ (ρθ)′) + g · ρ′      = Gˢρw
```
minor nit

src/CompressibleEquations/acoustic_substepping.jl
```diff
@@ -1,1205 +1,1527 @@
...
+    cs  = sqrt(γᵈ * Rᵈ * FT(300))
+
+    Δx_min = let
+        TX, TY, _ = topology(grid)
+        Δx = TX === Flat ? typemax(FT) : minimum_xspacing(grid)
+        Δy = TY === Flat ? typemax(FT) : minimum_yspacing(grid)
+        min(Δx, Δy)
     end
+
+    return max(1, ceil(Int, 2 * FT(Δt) * cs / Δx_min))
```
ewquon: Could be useful to make the acoustic substep CFL (i.e., inverse of the safety factor) configurable, e.g., as a parameter stored in `SplitExplicitTimeDiscretization`.

docs/src/compressible_dynamics.md
```diff
@@ -1,413 +1,448 @@
...
-\frac{c_s \, Δt}{Δz} \lesssim \frac{\sqrt{3}}{2} \approx 0.87
-```
+   ```math
+   Δτ ≤ \frac{\min(Δx, Δy)}{c_s + |\boldsymbol{u}|} .
+   ```
 
-With ``c_s \approx 347\,``m/s and ``Δz = 1000\,``m, this restricts
-``Δt \lesssim 2.5\,``s---only marginally better than the unsplit acoustic CFL and far
-short of what the advective CFL would allow.
+   The vertical implicit solve removes the vertical acoustic CFL constraint entirely.
```
ewquon, line 387:
```suggestion
   The speed of sound ``c_s = \sqrt{\gamma^d R^d T_r}``, where the reference temperature is chosen to be ``T_r = 300\,``K.
   
   The vertical implicit solve removes the vertical acoustic CFL constraint entirely.
```

src/CompressibleEquations/acoustic_substepping.jl
```diff
@@ -1,1205 +1,1527 @@
...
+##### Section 4 — Adaptive substep computation (acoustic CFL)
+#####
+
+"""
+$(TYPEDSIGNATURES)
+
+Compute the number of acoustic substeps ``N`` from the horizontal
+acoustic CFL: ``N \\approx \\lceil 2 \\Delta t \\, c_s / \\Delta x_\\min \\rceil``,
+with ``c_s = \\sqrt{\\gamma^d R^d \\theta_0}`` for a nominal
+``\\theta_0 = 300\\,\\mathrm{K}``. The factor of 2 is the standard
```
ewquon:
```suggestion
with ``c_s = \\sqrt{\\gamma^d R^d T_r}`` for a nominal
``T_r = 300\\,\\mathrm{K}``. The factor of 2 is the standard
```
Strictly speaking, this should be a reference air temperature not the surface potential temperature.

src/CompressibleEquations/acoustic_substepping.jl
```diff
@@ -1,1205 +1,1527 @@
...
+    end
+    return (dᵐ⁺, dˢ⁻)
+end
 
-Execute the acoustic substep loop for a Wicker-Skamarock RK3 stage
-using the Exner pressure formulation.
+@inline function implicit_damping_factors(damping::ThermalDivergenceDamping, ω, one_minus_ω, grid, FT)
+    damping.damp_vertical || return (zero(FT), zero(FT))
+    α    = convert(FT, damping.coefficient)
+    Δz   = convert(FT, minimum_zspacing(grid))
```
ewquon, line 1049: Using min dz everywhere will under-damp the upper levels if the grid is vertically stretched. Maybe something to add to the to-do list / open an issue on.

src/CompressibleEquations/acoustic_substepping.jl
glwagner: Agree with that. I guess this is 0.5 or something.

ewquon (approved):
Monumental effort in such little time, nice work @glwagner!

src/CompressibleEquations/time_discretizations.jl
glwagner: Hmm, the whole thing is experimental 😂 --- what do you mean exactly? For slow dynamics, its common to use eg a combination of Laplacian diffusion + hyperdiffusion. Supporting a tuple allows testing that kind of strategy for substepping as well. Basically this recognizes that there is a much wider variety of strategies than we see in the literature... if the goal is simply to damp divergence, there's a lot of ways to do it...

That said, in all my investigation it seems that only ThermalDivergenceDamping is really needed.

One thought could be to keep the infrastructure for experimentation while we are still refining this (eg don't regard this PR as the last say on substepping).

We can also clean up and remove the extra strategies. This might be fine since I am not sure we will need them again. My only hesitation is not knowing if we will want to keep exploring that (esp getting to more complex moist simulations, terrain following coordinates, etc)

glwagner: As for the upper sponge layer, I think this is something that is supported in WRF. It also may not be needed.

There is perhaps a more general issue that we do have the need to apply some dynamics on the substep and some on the slow step. Perhaps we need a more general way to specify how to discretize a term in time.

Another example might be a need to apply relaxation / grid nudging on substeps...

src/TimeSteppers/acoustic_runge_kutta_3.jl
glwagner: hmm yes. we can preclude terrain following here. This will be added in a follow up PR.

src/CompressibleEquations/acoustic_substepping.jl
glwagner: This might be easy to use the local grid spacing

src/CompressibleEquations/acoustic_substepping.jl
glwagner: ok, implemented

src/CompressibleEquations/time_discretizations.jl
ewquon: Ha! By experimental, I was asking if those were features added for testing / debugging and not part of (current) best practices. I don't think we need to remove those extra strategies -- basically this:

> One thought could be to keep the infrastructure for experimentation while we are still refining this (eg don't regard this PR as the last say on substepping).

The remaining question on this topic is whether or not to spend time documenting the extra strategies and inviting usage.

src/CompressibleEquations/time_discretizations.jl
glwagner: I removed vorticity damping for substepping

glwagner:
figure i should drop this here. baroclinic wave (dry) at 1/4 degree, runs with CFL=1.4 (advective) and adaptive time stepping

https://github.com/user-attachments/assets/2cf61c9b-9987-4d4e-9181-be5a5dc2d497

src/CompressibleEquations/acoustic_substepping.jl
```diff
@@ -1,1205 +1,1456 @@
...
+#####
+##### References:
+#####   - Wicker, L. J. & Skamarock, W. C., 2002. *Time-splitting methods for
+#####     elastic models using forward time schemes.* MWR 130, 2088.
+#####   - Baldauf, M., 2010. *Linear stability analysis of Runge-Kutta-based
+#####     partial time-splitting schemes for the Euler equations.* MWR 138, 4475.
+#####   - Klemp, J. B., Skamarock, W. C. & Ha, S.-Y., 2018. *Damping acoustic
+#####     modes in compressible HEVI and split-explicit time integration
+#####     schemes.* MWR 146, 1911.
 #####
```
ewquon, line 65:
```suggestion
#####
##### Call graph (with [K]ernel functions and inlined [h]elper functions denoted):
#####
##### acoustic_rk3_substep_loop!(model, substepper, Δt, β, Uᴸ)
##### ├── acoustic_substeps(...)
##### ├── stage_substep_count_and_size(...)
##### ├── assemble_slow_vertical_momentum_tendency!
##### │   └── [K] _assemble_slow_vertical_momentum_tendency_(no_)ref!
##### ├── initialize_stage_perturbations!
##### │   └── [K] _initialize_perturbation_with_rewind!         (×5 fields)
##### ├── for substep in 1:Nτ
##### │   ├── apply_horizontal_pressure_gradient_substep gate
##### │   ├── [K] _explicit_horizontal_step!
##### │   │   └── [h] linearized_pressure_perturbation, on_x/y_boundary
##### │   ├── implicit_damping_factors(...)
##### │   ├── [K] _build_predictors_and_vertical_rhs!
##### │   │   └── [h] theta_face_x/y/z_flux, ℑb_wθ, sponge_rhs, ∂z²_ρw′
##### │   ├── solve!(vertical_solver)
##### │   │   └── get_coefficient(Lower/Diag/Upper)
##### │   │       └── [h] ℑbzᵃᵃᶠ, sponge_term_diag
##### │   ├── [K] _post_solve_recovery!
##### │   ├── apply_divergence_damping!
##### │   │   ├── [K] _thermal_divergence_damping!
##### │   │   └── [K] _vorticity_damping_2nd!
##### │   └── accumulate_momentum_perturbations!
##### ├── finalize_time_averaged_velocity!
##### │   └── [K] _finalize_time_averaged_velocity!
##### ├── [K] _recover_full_state!
##### └── compute_velocities!(model)
#####
```
Maybe add something to help the humans navigate this source file?

glwagner: I think the main issue with this is that it has to be maintained along with the code itself

ewquon: Totally agree. My expectation is that it _should_ be automatic with Claude/Codex. But probably minimal value add overall.

src/CompressibleEquations/acoustic_substepping.jl
giordano: > or FT directly?

That'd work too, yes

src/Thermodynamics/reference_states.jl
```diff
@@ -440,28 +444,131 @@ end
...
+        end
+        p⁻ = pᵏ
+        ρ⁻ = ρᵏ
     end
+end
 
-    # Derive pressure and density from π₀
-    for k in 1:Nz
+@kernel function _compute_exner_reference!(π₀, pᵣ, ρᵣ, θ₀, grid, Nz, π₀_surface, pˢᵗ, cᵖᵈ, κ, Rᵈ, g)
+    _ = @index(Global)
```
giordano, line 507: What's the point of this non-assignment?

src/Thermodynamics/reference_states.jl
```diff
@@ -440,28 +444,131 @@ end
...
+        Π¹ = π₀_surface - g * Δzᶜ₁ / (2 * cᵖᵈ * θ₀[1, 1, 1])
+        T¹ = θ₀[1, 1, 1] * Π¹
+        p¹ = pˢᵗ * Π¹^(1/κ)
+        ρ¹ = p¹ / (Rᵈ * T¹)
+
+        π₀[1, 1, 1] = Π¹
+        pᵣ[1, 1, 1] = p¹
+        ρᵣ[1, 1, 1] = ρ¹
+    end
+    p⁻ = pᵣ[1, 1, 1]
```
giordano: Not quite sure of why, but this line is apparently causing problems when running on GPU: https://github.com/NumericalEarth/Breeze.jl/actions/runs/25794244378/job/75767367354?pr=622#step:5:39
```
ERROR: LoadError: InvalidIRError: compiling MethodInstance for Breeze.Thermodynamics.gpu__compute_exner_reference!(::KernelAbstractions.CompilerMetadata{KernelAbstractions.NDIteration.StaticSize{(1,)}, KernelAbstractions.NDIteration.DynamicCheck, Nothing, Nothing, KernelAbstractions.NDIteration.NDRange{1, KernelAbstractions.NDIteration.StaticSize{(1,)}, KernelAbstractions.NDIteration.StaticSize{(1,)}, Nothing, Nothing}}, ::Field{Nothing, Nothing, Center, Nothing, Nothing, Nothing, OffsetArrays.OffsetArray{Float32, 3, CUDACore.CuDeviceArray{Float32, 3, 1}}, Float32, Nothing, Nothing, Nothing}, ::Field{Nothing, Nothing, Center, Nothing, Nothing, Nothing, OffsetArrays.OffsetArray{Float32, 3, CUDACore.CuDeviceArray{Float32, 3, 1}}, Float32, Nothing, Nothing, Nothing}, ::Field{Nothing, Nothing, Center, Nothing, Nothing, Nothing, OffsetArrays.OffsetArray{Float32, 3, CUDACore.CuDeviceArray{Float32, 3, 1}}, Float32, Nothing, Nothing, Nothing}, ::Field{Nothing, Nothing, Center, Nothing, Nothing, Nothing, OffsetArrays.OffsetArray{Float32, 3, CUDACore.CuDeviceArray{Float32, 3, 1}}, Float32, Nothing, Nothing, Nothing}, ::RectilinearGrid{Float32, Periodic, Periodic, Bounded, Oceananigans.Grids.StaticVerticalDiscretization{OffsetArrays.OffsetVector{Float32, StepRangeLen{Float32, Float64, Float64, Int64}}, OffsetArrays.OffsetVector{Float32, StepRangeLen{Float32, Float64, Float64, Int64}}, Float32, Float32}, Float32, Float32, OffsetArrays.OffsetVector{Float32, StepRangeLen{Float32, Float64, Float64, Int64}}, OffsetArrays.OffsetVector{Float32, StepRangeLen{Float32, Float64, Float64, Int64}}, Nothing}, ::Int64, ::Float32, ::Float32, ::Float32, ::Float32, ::Float32, ::Float32) resulted in invalid LLVM IR
Reason: unsupported dynamic function invocation (call to unsafe_store!)
Stacktrace:
 [1] setproperty!
   @ /usr/local/share/julia/packages/CUDACore/eDgUR/src/device/runtime.jl:89
 [2] throw_boundserror
   @ /usr/local/share/julia/packages/CUDACore/eDgUR/src/device/quirks.jl:13
 [3] throw_boundserror
   @ /usr/local/share/julia/packages/CUDACore/eDgUR/src/device/quirks.jl:53
 [4] checkbounds
   @ ./abstractarray.jl:699
 [5] getindex
   @ /usr/local/share/julia/packages/OffsetArrays/b7qpm/src/OffsetArrays.jl:419
 [6] getindex
   @ /usr/local/share/julia/packages/Oceananigans/41MiP/src/Fields/field.jl:575
 [7] macro expansion
   @ /__w/Breeze.jl/Breeze.jl/src/Thermodynamics/reference_states.jl:536
 [8] macro expansion
   @ /usr/local/share/julia/packages/KernelAbstractions/ecO4B/src/macros.jl:332
 [9] gpu__compute_exner_reference!
   @ ./none:0
```
***Edit***: oh, is it because of the bounds checks?  But also, isn't this just
```suggestion
    p⁻ = p¹
```
Similarly for the line below.

***Edit 2***: I implemented the suggested change, let me know if it was incorrect.

test/acoustic_substepping.jl
```diff
@@ -329,6 +598,35 @@ for arch in arches
         @test !any(isnan, parent(model.dynamics.density))
```
giordano, line 598: This is now failing on Metal GPU, didn't use to be the case before.  We could mark this test as broken
```suggestion
        @test !any(isnan, parent(model.dynamics.density)) broken=arch isa GPU{Metal}
```
but may be nicer to understand whether something bad is happening (`FT = Float32` on CPU works fine)

***Edit***: uh, this test fails only locally for me but not on CI? 😳 

***Edit 2***: uhm, not getting the error anymore locally either 😕

src/Thermodynamics/reference_states.jl
giordano: I see this pattern in many kernels, but I'm completely missing why it's being used.

giordano:
Do we expect the benchmark `CBL; Dynamics: compressible_splitexplicit; Microphysics: nothing [Float32]/Advection: WENO5/NVIDIA L4/512x512x256` to be [20% faster with this PR](https://github.com/NumericalEarth/Breeze.jl/actions/runs/25798661589?pr=622#summary-75787045228)? 🤔 


<details>
<summary>Details</summary>

| Benchmark suite | Current: 5d33b9cca9690a8b8becb124ba090c39d6ecd1ec | Previous: 05c8fafbb0522e9c7cbc72dc89d0523083cd6810 | Ratio |
|-|-|-|-|
| `CBL; Dynamics: anelastic; Grid: 512x512x256 [Float32]/Advection: WENO5/NVIDIA L4/MixedPhaseEquilibrium` | `120422459.4441466` points/s | `122507934.5906608` points/s | `1.02` |
| `CBL; Dynamics: anelastic; Grid: 512x512x256 [Float32]/Advection: WENO5/NVIDIA L4/1M_MixedEquilibrium` | `83521924.3082238` points/s | `84700551.76975007` points/s | `1.01` |
| `CBL; Dynamics: anelastic; Grid: 512x512x256 [Float32]/Advection: WENO5/NVIDIA L4/1M_MixedNonEquilibrium` | `44986898.12308219` points/s | `45829144.829319276` points/s | `1.02` |
| `CBL; Dynamics: anelastic; Microphysics: nothing [Float32]/Compare advections/NVIDIA L4/WENO5 [256, 256, 128]` | `132861991.40523262` points/s | `133516744.7687157` points/s | `1.00` |
| `CBL; Dynamics: anelastic; Microphysics: nothing [Float32]/Advection: WENO5/NVIDIA L4/256x256x128` | `132861991.40523262` points/s | `133516744.7687157` points/s | `1.00` |
| `CBL; Dynamics: anelastic; Grid: 512x512x256 [Float32]/Advection: WENO5/NVIDIA L4/nothing` | `127888572.78051268` points/s | `129126968.35381861` points/s | `1.01` |
| `CBL; Dynamics: anelastic; Microphysics: nothing [Float32]/Compare advections/NVIDIA L4/WENO5 [512, 512, 256]` | `127888572.78051268` points/s | `129126968.35381861` points/s | `1.01` |
| `CBL; Dynamics: anelastic; Microphysics: nothing [Float32]/Advection: WENO5/NVIDIA L4/512x512x256` | `127888572.78051268` points/s | `129126968.35381861` points/s | `1.01` |
| `CBL; Dynamics: anelastic; Microphysics: nothing [Float32]/Compare advections/NVIDIA L4/WENO5 [768, 768, 256]` | `114110860.66455352` points/s | `115701896.59807333` points/s | `1.01` |
| `CBL; Dynamics: anelastic; Microphysics: nothing [Float32]/Advection: WENO5/NVIDIA L4/768x768x256` | `114110860.66455352` points/s | `115701896.59807333` points/s | `1.01` |
| `CBL; Dynamics: anelastic; Microphysics: nothing [Float32]/Compare advections/NVIDIA L4/WENO9 [256, 256, 128]` | `91429284.54896933` points/s | `92420239.95661892` points/s | `1.01` |
| `CBL; Dynamics: anelastic; Microphysics: nothing [Float32]/Advection: WENO9/NVIDIA L4/256x256x128` | `91429284.54896933` points/s | `92420239.95661892` points/s | `1.01` |
| `CBL; Dynamics: anelastic; Microphysics: nothing [Float32]/Compare advections/NVIDIA L4/WENO9 [512, 512, 256]` | `84702305.9281727` points/s | `87532164.63927339` points/s | `1.03` |
| `CBL; Dynamics: anelastic; Microphysics: nothing [Float32]/Advection: WENO9/NVIDIA L4/512x512x256` | `84702305.9281727` points/s | `87532164.63927339` points/s | `1.03` |
| `CBL; Dynamics: anelastic; Microphysics: nothing [Float32]/Compare advections/NVIDIA L4/WENO9 [768, 768, 256]` | `75145942.68703921` points/s | `78509911.56156372` points/s | `1.04` |
| `CBL; Dynamics: anelastic; Microphysics: nothing [Float32]/Advection: WENO9/NVIDIA L4/768x768x256` | `75145942.68703921` points/s | `78509911.56156372` points/s | `1.04` |
| `CBL; Dynamics: compressible_explicit; Microphysics: 1M_MixedNonEquilibrium [Float64]/Compare backends/NVIDIA L4/vanilla 256x256x128` | `4020106.079616179` points/s | `4013075.9269985864` points/s | `1.00` |
| `CBL; Dynamics: compressible_explicit; Microphysics: 1M_MixedNonEquilibrium [Float64]/Compare backends/NVIDIA L4/reactant 256x256x128` | `4807600.515722543` points/s | `4897955.461021911` points/s | `1.02` |
| `CBL; AD; Dynamics: compressible_explicit; Microphysics: nothing [Float64]/Advection: WENO5/NVIDIA L4/64x64x32` | `2493654.1873611882` points/s | `2483673.8408254078` points/s | `1.00` |
| `CBL; Dynamics: compressible_splitexplicit; Microphysics: nothing [Float32]/Advection: WENO5/NVIDIA L4/512x512x256` | `25315966.693637125` points/s | `20461349.267835885` points/s | `0.81` |

</details>

src/Thermodynamics/reference_states.jl
glwagner: Can we delete it and the kernel is still valid?

glwagner:
> Do we expect the benchmark `CBL; Dynamics: compressible_splitexplicit; Microphysics: nothing [Float32]/Advection: WENO5/NVIDIA L4/512x512x256` to be [20% faster with this PR](https://github.com/NumericalEarth/Breeze.jl/actions/runs/25798661589?pr=622#summary-75787045228)? 🤔
> 
> Details

I would hope much more than 20%. But we have work to do I guess. This PR is a complete and utter refactor of acoustic substepping. This is also known as "SplitExplicitTimeStepping", which is used for CompressibleDynamics.

src/Thermodynamics/reference_states.jl
glwagner: that looks right!

src/Thermodynamics/reference_states.jl
giordano: I'm mildly we don't _have_ to reference `@index(Global)`.

kaiyuan-cheng:
There might be an issue with the moist compressible dynamics. I ran the supercell example using both dynamics options, and the results look quite different even in the early stages. For comparison, I also ran CM1, where both dynamics produced nearly the same pattern (not shown).

image

image

ewquon:
@kaiyuan-cheng to confirm: the Breeze anelastic agrees with the CM1 anelastic/compressible dynamics? For Breeze compressible would it make sense to turn off substepping as a sanity check?

kaiyuan-cheng:
> @kaiyuan-cheng to confirm: the Breeze anelastic agrees with the CM1 anelastic/compressible dynamics? For Breeze compressible would it make sense to turn off substepping as a sanity check?

No, Breeze anelastic does not agree with CM1, which is expected as the physics-dynamics coupling is different. Good idea, I can give explicit compressible dynamics a try.

src/CompressibleEquations/time_discretizations.jl
```diff
@@ -1,60 +1,475 @@
...
+end
+
+convert_acoustic_parameter(::Type{FT}, ::Nothing) where FT = nothing
+
+convert_acoustic_parameter(::Type{FT}, sponge::UpperSponge) where FT =
+    UpperSponge{FT, typeof(sponge.ramp)}(convert(FT, sponge.damping_rate),
+                                        convert(FT, sponge.depth),
+                                        sponge.ramp)
+
+struct SplitExplicitTimeDiscretization{N, FT, D, US, AD <: AcousticSubstepDistribution}
```
giordano, line 453: Adding a docstring to `SplitExplicitTimeDiscretization` would solve the only remaining issue in the docs (there are a few `@ref`s to `SplitExplicitTimeDiscretization` in the docs)

kaiyuan-cheng:
Looks like the discrepancy comes from initialization. The dynamics themselves should be fine.

https://github.com/user-attachments/assets/36e64fcd-e38d-472a-9b95-f6ba02fe98b0

PR #628, opened 2026-04-12
[CI] Update most Julia versions to v1.12.6

giordano:
Will update [our docker images](https://github.com/NumericalEarth/breeze-docker-images) when the [base Julia image](https://hub.docker.com/_/julia) is updated to v1.12.6.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -43,7 +43,7 @@ jobs:
       matrix:
         version:
           - "min"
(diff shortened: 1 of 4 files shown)


giordano:
Precompilation of the packages is messed up, even worse than v1.12.5, sigh. I think I'll add a step to precompile all the packages just once. ***Edit***: this will be fixed by https://github.com/JuliaLang/Pkg.jl/pull/4658

PR #630, opened 2026-04-14
Rework parallelism strategy for running docs

giordano:
Taken out of #615.


giordano:
I'm not sure this is necessarily the best solution, but I always wanted to better identify the GPU jobs, let's see how to make good use of that information for scheduling purposes

PR #631, opened 2026-04-16
Call `materialize_advection` for WENO GPU compatibility

ewquon:
## Summary

  - Adds missing `materialize_advection` call in `AtmosphereModel` constructor, fixing WENO advection on GPU
  - Oceananigans v0.106.6 refactored WENO to defer the weight computation type (`WCT`) as `Nothing` until the grid architecture is known. `materialize_advection` resolves it to
  `BackendOptimizedDivision` on GPU. All Oceananigans models call this, but `AtmosphereModel` was never updated — causing `InvalidIRError: unsupported dynamic function invocation (call to
   newton_div)` when running any WENO scheme on GPU.

  ## Test plan

  - [x] Run RICO example on GPU (previously crashed at `run!` with `InvalidIRError`)
  - [x] Run BOMEX or other WENO example on GPU to confirm no regression
  - [ ] CPU examples should be unaffected (`materialize_advection` is a no-op for non-deferred schemes)

### DIFF
--- a/src/AtmosphereModels/atmosphere_model.jl
+++ b/src/AtmosphereModels/atmosphere_model.jl
@@ -2,7 +2,7 @@ using ..Thermodynamics: Thermodynamics, ThermodynamicConstants
 
 using Oceananigans: Oceananigans, AbstractModel, Center, CenterField, Clock, Field,
                     Centered, fields, prognostic_fields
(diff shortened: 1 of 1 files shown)


kaiyuan-cheng (approved):
I am also running into the same issue. This fix looks good. Thanks!

PR #633, opened 2026-04-16
More 3D topologies for reactant tests

dkytezab:


giordano:
Keep here only the changes to reactant_centered_compilation and try to fix reactant_weno_compilation in a separate PR?

glwagner (approved):

dkytezab:
It's unclear to me whether the NaNs that we are hitting in the triply-bounded case are related to https://github.com/EnzymeAD/Enzyme-JAX/pull/2474 or are a separate issue; either way we will know soon when the next Reactant version gets pushed out

giordano:
> either way we will know soon when the next Reactant version gets pushed out

Not so fast: https://github.com/NumericalEarth/Breeze.jl/blob/a00986e2ede2543517bfbf7d93b01585a3e54fd8/test/Project.toml#L46 🫠  See https://github.com/NumericalEarth/Breeze.jl/pull/675#issuecomment-4395271826.  ***Edit***: reverted in #691, tests are passing now.

giordano:
> Keep here only the changes to reactant_centered_compilation and try to fix reactant_weno_compilation in a separate PR?

☝️

dkytezab:
So what's interesting is that the triply bounded Reactant case stops NaN-ing if we simply bump the grid size up a bit. Seeing as vanilla has no issue with the failing size `(8, 8, 8)`, smells like a correctness bug.

giordano:
Those should be "easy" to address in Reactant/Enzyme-JAX once we have a minimal reproducer (but getting that is probably the most time-consuming part)

dkytezab:
I say let's bring back the check and bump the grid size. Seeing at the test is primarily for compilation, and the pathological behavior is apparently only exhibited by unrealistically small grids. Then will flag this as something to look into.

giordano:
This is still failing

dkytezab:
Interesting. So `Periodic, Periodic, Bounded` is also unhappy but the previous grid size was too small to uncover it? I don't hit issues if I run this locally (macOS). Maybe something to do with CUDA intrinsics?

test/Project.toml
```diff
@@ -46,7 +47,7 @@ Oceananigans = "0.110.2"
 ParallelTestRunner = "2.0.1"
 Printf = "<0.0.1, 1"
 RRTMGP = "0.21"
-Reactant = "=0.2.266"
+Reactant = "=0.2.267"
```
giordano: We are blocked by #801 at the moment (at least for `reactant/advection_diffusion_ad`, but maybe you can see if that PR helps with the other tests)

dkytezab: Everything is still failing cause time out. An enemy has sent not one but three bugs after us, so that in pursuing one we are always confounded by the other two...

giordano: No, timeout should have been worked around by #799.  But now #801 is making the other Reactant jobs terminate, so either temporarily delete the `reactant/advection_diffusion_ad` file, or temporarily remove the `--quickfail` flag https://github.com/NumericalEarth/Breeze.jl/blob/587faa6c6898c8dfd47f855521bcdb778928a758/.github/workflows/CI.yml#L118

giordano:
Tests are finally passing, only thing is that [test times now doubled](https://github.com/NumericalEarth/Breeze.jl/actions/runs/28180626876/job/83468968064?pr=633#step:14:1141)
```
    reactant/advection_diffusion_ad           |   12     12  17m08.0s
    reactant/weno_compilation                 |   48     48  38m04.0s
    reactant/centered_compilation             |   72     72  50m02.7s
```
I guess there aren't easy way to split up `reactant/weno_compilation` and `reactant/centered_compilation` without having duplication?

dkytezab:
I'm ok with removing the 2d topology tests from `reactant/centered_compilation`. Seems like register spillage has gotten significantly worse. I suspect we need
`XLA_FLAGS: '--xla_disable_hlo_passes=multi_output_fusion'` for test env as well

dkytezab:
A little better!
```
    reactant/advection_diffusion_ad           |   12     12  15m39.4s
    reactant/centered_compilation             |   72     72  20m10.0s
    reactant/weno_compilation                 |   48     48  26m00.8s
```
For reference, CI is (as of last commit)
```
    reactant/advection_diffusion_ad           |   12     12  16m36.8s
    reactant/centered_compilation             |   48     48  12m39.4s
    reactant/weno_compilation                 |   48     48  22m57.9s
```
Also for reference we bumped the size of the grids in the test cases which could explain the small increase. Assuming last CI passes, I think this is good to go cc @giordano

giordano (approved):
Yes, a lot better 🚀

PR #644, opened 2026-04-20
Register atan, atand, mod as binary operations on Fields

ewquon:
## Summary

- Adds `src/AbstractOperations.jl`, a small module that registers `atan`, `atand`, and `mod` as `Oceananigans.AbstractOperations.@binary` operators so they can be composed into `BinaryOperation` trees on `Field`s.
- Motivated by the wind-direction diagnostic in `examples/neutral_atmospheric_boundary_layer.jl`, where `mod(270 - atand(v, u), 360)` currently has to drop to `interior(...)` with `@.` broadcasting because `atand` and `mod` aren't registered operators.
- Adds `test/abstract_operations.jl` covering `atan`, `atand`, and `mod` on `default_arch` (CPU locally, GPU in CI).

A `TODO` in the new source file notes a potential upstream Oceananigans cleanup: `define_binary_operator` currently references non-exported helpers (`apply_op`, `grid_metric_operation`, `choose_location`, `BinaryOperation`, `GridMetric`) as bare symbols in its expansion, so downstream packages have to import all of them by hand. Fully qualifying those references upstream would let this file shrink to just `@binary` and the `Base` functions being registered.

## Test plan

- [x] \`julia --project=test test/runtests.jl abstract_operations\` passes (5/5 assertions)
- [x] CI picks up the new file automatically via \`ParallelTestRunner.find_tests\`
- [x] Follow-up: update \`examples/neutral_atmospheric_boundary_layer.jl\` to use the registered operator form once this lands


src/AbstractOperations.jl
```diff
@@ -0,0 +1,25 @@
...
+
+import Base: atan, atand, mod
+
+# Register two-argument elementary functions useful for atmospheric diagnostics
+# (e.g. wind direction from velocity components, periodic wrapping) so they can
+# be composed with Fields in `BinaryOperation` trees without dropping to
+# `interior(...)` / `@.` escape hatches.
+@binary atan
+@binary atand
+@binary mod
```
giordano: Sorry, I'm a bit allergic to the use of `import`, because it's a spooky action at a distance.  Just passing `Module.function_name` to `@binary` should be equivalent 
```suggestion
# Register two-argument elementary functions useful for atmospheric diagnostics
# (e.g. wind direction from velocity components, periodic wrapping) so they can
# be composed with Fields in `BinaryOperation` trees without dropping to
# `interior(...)` / `@.` escape hatches.
@binary Base.atan
@binary Base.atand
@binary Base.mod
```
Side note, I just noticed the current implementation of `@binary` requires you to import into scope all those extra symbols, that's unnecessary, I'll have a look separately.

giordano:
> A `TODO` in the new source file notes a potential upstream Oceananigans cleanup: `define_binary_operator` currently references non-exported helpers (`apply_op`, `grid_metric_operation`, `choose_location`, `BinaryOperation`, `GridMetric`) as bare symbols in its expansion, so downstream packages have to import all of them by hand. Fully qualifying those references upstream would let this file shrink to just `@binary` and the `Base` functions being registered.

https://github.com/CliMA/Oceananigans.jl/pull/5528

giordano:
Mumble mumble mubmle...something is holding Oceananigans back to v0.106

giordano:
> Mumble mumble mubmle...something is holding Oceananigans back to v0.106

Ah!  The problem was that the `[sources]` entry I added for Oceananigans in the top-level project file wasn't propagated to all others, and so the package manager wouldn't find Oceananigans v0.107.3 in the registry, and the only other option left was to downgrade it to v0.106 (which was also wrong, because with this PR we explicitly require v0.107.3).  Problem solved, and now all tests are running.  Also, the tests with Julia v1.11 actually ran before already, successfully.

All in all, this PR should be good, and we only need to wait for https://github.com/CliMA/Oceananigans.jl/pull/5528 (basically validated here already) to make it into a new release

giordano (approved):

navidcy:
Wouldn't this changes fit better in Oceananigans?

giordano:
I thought about that a little bit (but admittedly not long enough!): I suspect it may be application-specific what operators need to be registered, right?  Otherwise Oceananigans could define dozens of such operators by default, which isn't currently doing, and I _thought_ just to keep the default lean.

navidcy:
True, good point.

giordano:
Maybe @simone-silvestri has an opinion about where these should live? :eyes:

simone-silvestri:
I would be ok either way, but I don't think it is so much of a burden adding these methods to Oceananigans?

src/AbstractOperations.jl
```diff
@@ -0,0 +1,13 @@
...
+
+using Oceananigans.AbstractOperations: @binary
+
+# Register two-argument elementary functions useful for atmospheric diagnostics
+# (e.g. wind direction from velocity components, periodic wrapping) so they can
+# be composed with Fields in `BinaryOperation` trees without dropping to
+# `interior(...)` / `@.` escape hatches.
+@binary Base.atan
+@binary Base.atand
+@binary Base.mod
```
glwagner, line 11: Shouldn't these be registered in Oceananigans rather than Breeze? What's the reason to register them here but not make them available in Oceananigans?

giordano: See the discussion above 😄

glwagner: I'm actually not really ok with this being in Breeze, it opens a pandora's box. We should have clarity about what belongs in Oceananigans and Breeze. `atan` and `mod` are not "atmosphere specific" functionality.

PR #645, opened 2026-04-20
Drop :grid from default_included_properties; test no-warning init

### LINKED ISSUE #643: JLD2Writer warns about duplicate serialized/grid during init with AtmosphereModel
## Summary

Every `JLD2Writer` attached to an `AtmosphereModel` emits a warning during initialization:
```
┌ Warning: error ArgumentError("a group or dataset named grid is already present within this group") thrown when
trying to serialize the grid at serialized/grid
└ @ Oceananigans.OutputWriters ~/.julia/.../OutputWriters/jld2_writer.jl:234
```

Observed while running `examples/neutral_atmospheric_boundary_layer.jl` (two writers → two warnings). Simulation otherwise runs fine and the output file contains a valid `serialized/grid`.

## Cause

Breeze overrides `default_included_properties` to include `:grid`:

https://github.com/NumericalEarth/Breeze.jl/blob/main/src/AtmosphereModels/atmosphere_model.jl#L464
```julia
Oceananigans.OutputWriters.default_included_properties(::AtmosphereModel) = [:grid, :thermodynamic_constants]
```

Oceananigans PR CliMA/Oceananigans.jl#5486 ("Save grid per field in JLD2Writer") added automatic grid serialization. JLD2Writer initialization now:

1. Iterates including and writes each property to serialized/$property — Breeze's override lands the grid at serialized/grid.
2. Collects unique grids from the outputs and writes them to serialized/grid (single-grid case) — this errors because the path already exists, caught and surfaced as the warning.

## Fix

Drop :grid from the override and let Oceananigans' new code path handle it:

```julia
Oceananigans.OutputWriters.default_included_properties(::AtmosphereModel) = [:thermodynamic_constants]
```

Worth adding a small test that a JLD2Writer attached to an AtmosphereModel initializes without warnings.

ewquon:
## Summary
- Drop `:grid` from `default_included_properties(::AtmosphereModel)` so Oceananigans' automatic grid serialization (CliMA/Oceananigans.jl#5486) isn't duplicated.
- Add a regression test (`test/output_writers.jl`) that initializes a `JLD2Writer` attached to an `AtmosphereModel` and asserts no warnings, plus verifies the file contains `serialized/grid` (matching the model grid) and `serialized/thermodynamic_constants`.

Closes #643.

## Test plan
- [x] `julia --project=test test/runtests.jl --jobs=1 output_writers` passes (5/5 assertions)
- [x] CI green on main

### DIFF
--- a/test/output_writers.jl
+++ b/test/output_writers.jl
@@ -32,6 +32,6 @@ using Logging: Warn
             @test file["serialized/grid"] == on_architecture(CPU(), grid)
         end
     finally
-        isfile(writer.filepath) && rm(writer.filepath, force=true)
+        rm(writer.filepath; force=true)
     end
 end
--- a/test/output_writers.jl
+++ b/test/output_writers.jl
@@ -1,0 +1,37 @@
+using Test
+using Breeze
+using Oceananigans
+using Oceananigans.Architectures: on_architecture, CPU
+using Oceananigans.OutputWriters: jldopen
+using Logging: Warn
+
+# Regression test for https://github.com/NumericalEarth/Breeze.jl/issues/643.
+# Before the fix, Breeze's `default_included_properties` override included `:grid`,
+# which collided with Oceananigans' automatic grid serialization and produced a
+# "grid is already present within this group" warning during JLD2Writer init.
+@testset "JLD2Writer initializes without warnings [$(FT)]" for FT in test_float_types()
+    Oceananigans.defaults.FloatType = FT
+    grid = RectilinearGrid(default_arch; size=(2, 2, 4), extent=(100, 100, 100))
+    model = AtmosphereModel(grid)
+
+    filepath = tempname() * ".jld2"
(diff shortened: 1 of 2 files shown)


test/output_writers.jl
```diff
@@ -0,0 +1,37 @@
...
+        @test_logs min_level=Warn Oceananigans.initialize!(writer, model)
+        @test isfile(writer.filepath)
+
+        jldopen(writer.filepath, "r") do file
+            @test haskey(file, "serialized/grid")
+            @test haskey(file, "serialized/thermodynamic_constants")
+            @test file["serialized/grid"] == on_architecture(CPU(), grid)
+        end
+    finally
+        isfile(writer.filepath) && rm(writer.filepath, force=true)
```
giordano: Either don't use force=true or don't check that the file exists (I vote forthe  latter), doing both is redundant 
```suggestion
        rm(writer.filepath; force=true)
```

giordano (approved):

PR #648, opened 2026-04-21
Fix moisture initialization in splitting supercell example

kaiyuan-cheng:
## Summary

Fixes the moisture initialization in `examples/splitting_supercell.jl` to match the Klemp et al. (2015) specification (Eqs. 11–12).

- **Before**: the shape function `1 − 3/4 (z/zᵖ)^{5/4}` was passed as relative humidity with no cap. At the 300 K / 1000 hPa surface this gave ℋ = 1, i.e. saturation with `qᵥ ≈ 22.8 g/kg` — far wetter than Klemp's sounding and yielding ~6000 J/kg CAPE.
- **After**: apply the RH profile as specified, compute `qᵥ = ℋ · qᵛ⁺(T, p)`, and cap at `qᵛ_max = 0.014 kg/kg` (Klemp's surface limit). Local T, p, ρ are estimated from a hydrostatic Exner profile `Π(z) = 1 − g z/(cᵖᵈ θ₀)` combined with the actual `θ(z)`. Back-of-envelope CAPE drops to ~1400 J/kg, in the expected range for this test.

The initial narrative, the profile visualization (now showing qᵛ in g/kg), and the `set!` call (now uses `qᵛ=` instead of `ℋ=`) were updated correspondingly.

## Test plan

- [ ] Build the example locally and confirm the initial-conditions figure shows the expected qᵛ profile (≈14 g/kg at surface, tapering to ~0 in the stratosphere)
- [ ] Run the 2-hour simulation and verify the storm splits and the max-w time series is closer to DCMIP2016 reference values than the previous ~60 m/s extreme
- [ ] Inspect the first output slice at t ≈ 2 min to confirm no spurious stratospheric cloud layer from initial supersaturation

### DIFF
--- a/examples/splitting_supercell.jl
+++ b/examples/splitting_supercell.jl
@@ -14,7 +14,7 @@
 # The simulation initializes a conditionally unstable atmosphere with a warm bubble perturbation
 # that triggers deep convection. The environment includes:
 # - A realistic tropospheric potential temperature profile with a tropopause at 12 km
-# - Relative humidity that decreases with height, with the resulting water vapor mixing ratio capped at 14 kg/kg to 
+# - Relative humidity that decreases with height, with the resulting water vapor mixing ratio capped at 14 g/kg to
 # approximate a well-mixed boundary layer in the lowest kilometer.
 # - Wind shear in the lower 5 km to promote storm rotation and supercell development
 #
--- a/examples/splitting_supercell.jl
+++ b/examples/splitting_supercell.jl
@@ -132,7 +132,7 @@ function θ_background(z)
     return (z ≤ zᵖ) * θᵗ + (z > zᵖ) * θˢ
 end
 
-# Relative humidity profile (Equations 11–12 in [KlempEtAl2015](@citet)) combined with
+# Relative humidity profile (Equations 11–12 by [KlempEtAl2015](@citet)) combined with
 # a cap on the water vapor mixing ratio at ``qᵛ_max``. The local temperature and density
 # are obtained by numerically integrating the hydrostatic balance with the actual
 # ``θ(z)`` profile:
--- a/examples/splitting_supercell.jl
+++ b/examples/splitting_supercell.jl
@@ -14,7 +14,7 @@
 # The simulation initializes a conditionally unstable atmosphere with a warm bubble perturbation
 # that triggers deep convection. The environment includes:
 # - A realistic tropospheric potential temperature profile with a tropopause at 12 km
-# - Moisture that decreases with height, with relative humidity dropping above the tropopause
+# - Water vapor mixing ratio that decreases with height, falling to one-quarter of the surface value above the tropopause
 # - Wind shear in the lower 5 km to promote storm rotation and supercell development
 #
 # ### Potential temperature profile
@@ -106,6 +106,7 @@ dynamics = AnelasticDynamics(reference_state)
 θᵖ = 343       # K - tropopause potential temperature
 zᵖ = 12000     # m - tropopause height
 Tᵖ = 213       # K - tropopause temperature
+qᵛ⁰ = 0.014    # kg/kg - surface water vapor mixing ratio
 nothing #hide
 
 # Wind shear parameters control the low-level environmental wind profile:
@@ -119,6 +120,7 @@ nothing #hide
 
 g = constants.gravitational_acceleration
 cᵖᵈ = constants.dry_air.heat_capacity
+Rᵈ = dry_air_gas_constant(constants)
 nothing #hide
 
 # Background potential temperature profile (Equation 14 in [KlempEtAl2015](@citet)):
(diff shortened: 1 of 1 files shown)


examples/splitting_supercell.jl
```diff
@@ -14,7 +14,8 @@
 # The simulation initializes a conditionally unstable atmosphere with a warm bubble perturbation
 # that triggers deep convection. The environment includes:
 # - A realistic tropospheric potential temperature profile with a tropopause at 12 km
-# - Moisture that decreases with height, with relative humidity dropping above the tropopause
+# - Relative humidity that decreases with height, with the resulting water vapor mixing ratio capped at 14 g/kg to
```
navidcy:
```suggestion
# - Relative humidity that decreases with height, with the resulting water vapor mixing ratio capped at 0.014 kg/kg to
```

(only to match the code value below)

examples/splitting_supercell.jl
```diff
@@ -129,9 +132,26 @@ function θ_background(z)
     return (z ≤ zᵖ) * θᵗ + (z > zᵖ) * θˢ
 end
 
-# Relative humidity profile (decreases with height, 25% above tropopause):
+# Relative humidity profile (Equations 11–12 in [KlempEtAl2015](@citet)) combined with
```
navidcy:
```suggestion
# Relative humidity profile (Equations 11–12 by [KlempEtAl2015](@citet)) combined with
```

examples/splitting_supercell.jl
```diff
@@ -129,9 +132,26 @@ function θ_background(z)
...
+# a cap on the water vapor mixing ratio at ``qᵛ_max``. The local temperature and density
```
navidcy:
```suggestion
# a cap on the water vapor mixing ratio at ``qᵛ_{max}``. The local temperature and density
```

examples/splitting_supercell.jl
```diff
@@ -171,11 +191,11 @@ uᵢ(x, y, z) = u_background(z)
...
 # ## Visualization of initial conditions and warm bubble perturbation
 #
-# We visualize the background potential temperature, relative humidity, and wind shear profiles
-# that define the environmental stratification:
+# We visualize the background potential temperature, water vapor mixing ratio, and wind shear
+# profiles that define the environmental stratification:
 
 θ_profile = set!(Field{Nothing, Nothing, Center}(grid), z -> θ_background(z))
-ℋ_profile = set!(Field{Nothing, Nothing, Center}(grid), z -> ℋ_background(z) * 100)
+qᵛ_profile = set!(Field{Nothing, Nothing, Center}(grid), 1000 * qᵛ_column)
```
navidcy: why `1000 *`? change units? if yes, why do we need to change units, just for plotting?

perhaps 


```suggestion
qᵛ_profile = set!(Field{Nothing, Nothing, Center}(grid), 1000 * qᵛ_column) # convert kg/kg -> g/kg
```

examples/splitting_supercell.jl
```diff
@@ -225,9 +245,7 @@ model = AtmosphereModel(grid; dynamics, microphysics, advection, thermodynamic_c
 #
 # We initialize the model with the previously described initial conditions, including a warm-bubble perturbation.
 
-ℋᵢ(x, y, z) = ℋ_background(z)
-
-set!(model, θ=θᵢ, ℋ=ℋᵢ, u=uᵢ)
+set!(model, θ=θᵢ, qᵛ=qᵛ_column, u=uᵢ)
```
navidcy, line 248: this means that here we are using g/kg units?? Is this correct?

kaiyuan-cheng: We're using kg/kg.

examples/splitting_supercell.jl
kaiyuan-cheng: qᵛ_profile is just for plotting.

examples/splitting_supercell.jl
navidcy: oh yeah, it's q_profile that was in g/kg; my bad!

examples/splitting_supercell.jl
```diff
@@ -129,9 +132,26 @@ function θ_background(z)
     return (z ≤ zᵖ) * θᵗ + (z > zᵖ) * θˢ
 end
 
-# Relative humidity profile (decreases with height, 25% above tropopause):
+# Relative humidity profile (Equations 11–12 by [KlempEtAl2015](@citet)) combined with
+# a cap on the water vapor mixing ratio at ``qᵛ_{max}``. The local temperature and density
+# are obtained by numerically integrating the hydrostatic balance with the actual
+# ``θ(z)`` profile:
+
+function qᵛ_background(z)
```
glwagner: if you can convert this to math notation, ie `qᵛ_bg` or `qᵛᵇ`. Same for other symbols that mix words and math notation

examples/splitting_supercell.jl
```diff
@@ -106,6 +108,7 @@ dynamics = AnelasticDynamics(reference_state)
 θᵖ = 343       # K - tropopause potential temperature
 zᵖ = 12000     # m - tropopause height
 Tᵖ = 213       # K - tropopause temperature
+qᵛ_max = 0.014    # kg/kg - cap on water vapor mixing ratio
```
```suggestion
qᵛ_max = 0.014  # kg/kg - cap on water vapor mixing ratio
```

can this be linked to a specific line or equation in the Klemp paper?

glwagner (approved):

examples/splitting_supercell.jl
glwagner: note i tried to fix the alignment with above comments but not sure I succeeded

examples/splitting_supercell.jl
```diff
@@ -129,9 +132,26 @@ function θ_background(z)
     return (z ≤ zᵖ) * θᵗ + (z > zᵖ) * θˢ
 end
 
-# Relative humidity profile (decreases with height, 25% above tropopause):
+# Relative humidity profile (Equations 11–12 by [KlempEtAl2015](@citet)) combined with
+# the water vapor cap ``qᵛ_{max}``. The local temperature and density
+# are obtained by numerically integrating the hydrostatic balance with the actual
+# ``θ(z)`` profile:
+
+function qᵛ_bg(z)
```
navidcy, line 140:
```suggestion
function qᵛ_background(z)
```

examples/splitting_supercell.jl
```diff
@@ -129,9 +132,26 @@ function θ_background(z)
...
+    qᵛ⁺ = saturation_specific_humidity(T, ρ, constants, PlanarLiquidSurface())
+    return min(ℋ * qᵛ⁺, qᵛ_max)
+end
+
+# Evaluate ``qᵛ`` on a column field so the hydrostatic integration runs only once
+# per vertical level rather than once per horizontal grid point:
 
-ℋ_background(z) = (1 - 3/4 * (z / zᵖ)^(5/4)) * (z ≤ zᵖ) + 1/4 * (z > zᵖ)
+qᵛ_column = Field{Nothing, Nothing, Center}(grid)
+set!(qᵛ_column, qᵛ_bg)
```
navidcy, line 154:
```suggestion
set!(qᵛ_column, qᵛ_background)
```

glwagner: I had just suggested @kaiyuan-cheng change that... do you prefer _background? My thought is that it mixes words and math notation.

navidcy: there is another variable in the script that is `_background`.
I just thought better to homogenize; so perhaps we change the other one to `_bg`? That was my only intention.

navidcy: I think it's the `θ_background`?

navidcy: Perhaps let's leave it as is!

navidcy:
Let's merge this; looks good!

kaiyuan-cheng:
I can't merge using the command line. Can someone help with this?

PR #651, opened 2026-04-22
Fix bulk sensible heat flux potential temperature conversion

bischtob:
## Summary

This PR fixes the `PotentialTemperatureFlux` branch of `BulkSensibleHeatFlux` so it compares atmospheric potential temperature against the surface-equivalent potential temperature, not the raw surface temperature.

The change is intentionally narrow:

- thread `standard_pressure` into the materialized bulk sensible heat boundary condition
- convert `T₀` to `θ₀` before forming the bulk temperature difference
- add a symmetric `potential_temperature_from_temperature` helper and use it at the boundary-condition call site
- add regression coverage for the zero-flux matched-`θ` case, plus helper round-trip tests

## Problem

For `ρθ` boundary conditions, the bulk sensible heat flux path was effectively using

`Δθ = θ - T₀`

instead of comparing like with like.

That is only correct when the actual surface pressure `p₀` is equal to the fixed reference pressure `pˢᵗ` used to define potential temperature. In Breeze those pressures are generally not the same:

- `p₀` is the actual surface pressure
- `pˢᵗ` is the fixed reference pressure used in the definition of `θ`

When `p₀ ≠ pˢᵗ`, this produces a systematic bias in the potential-temperature bulk flux.

## What Changed

- `BulkSensibleHeatFluxFunction` now carries `standard_pressure` through construction, adaptation, and materialization
- the `PotentialTemperatureFlux` path now converts the configured surface temperature to surface-equivalent potential temperature with

`θ₀ = potential_temperature_from_temperature(T₀, p₀, constants; pˢᵗ)`

- the `StaticEnergyFlux` path is unchanged
- `BulkSensibleHeatFlux(...)` remains unchanged from the caller’s perspective

## Why This Is The Right Fix

The bug lives at the boundary-condition interface between a user-provided surface temperature and a prognostic potential-temperature formulation. The fix therefore stays local to that interface:

- the thermodynamics conversion is implemented once in a helper that mirrors the existing `temperature_from_potential_temperature` convenience function
- the bulk flux code reuses that helper instead of carrying a second inline conversion
- no broader thermodynamics refactor or public API redesign is introduced

## Validation

- `Pkg.test("Breeze"; test_args=\`unit_tests\`)`
- `Pkg.test("Breeze"; test_args=\`forcing_and_boundary_conditions\`)`
- `Pkg.test("Breeze"; test_args=\`quality_assurance\`)`

The new regression test sets the near-surface atmospheric `θ` equal to the dry-equivalent surface `θ₀` under nonzero wind and verifies that the materialized bottom `ρθ` bulk sensible heat flux is zero to floating-point tolerance. Before this fix, that case was biased by the `θ - T₀` mismatch.

### DIFF
--- a/src/BoundaryConditions/bulk_scalar_fluxes.jl
+++ b/src/BoundaryConditions/bulk_scalar_fluxes.jl
@@ -5,11 +5,12 @@
 struct PotentialTemperatureFlux end
 struct StaticEnergyFlux end
 
-struct BulkSensibleHeatFluxFunction{C, G, T, P, TC, F, FV, FS}
+struct BulkSensibleHeatFluxFunction{C, G, T, P, SP, TC, F, FV, FS}
     coefficient :: C
     gustiness :: G
     surface_temperature :: T
     surface_pressure :: P
+    standard_pressure :: SP
     thermodynamic_constants :: TC
     formulation :: F
     filtered_velocities :: FV  # Nothing or FilteredSurfaceVelocities
@@ -17,7 +18,7 @@ struct BulkSensibleHeatFluxFunction{C, G, T, P, TC, F, FV, FS}
 end
 
 """
-    BulkSensibleHeatFluxFunction(; coefficient, gustiness=0, surface_temperature)
+$(TYPEDSIGNATURES)
 
 A bulk sensible heat flux function. The flux is computed as:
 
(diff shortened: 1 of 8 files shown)


glwagner (approved):
Good catch. Previously we defined the standard pressure for potential temperature as p(z=0). When we changed that, the temperature and potential temperature are no longer equivalent at z=0.

src/BoundaryConditions/bulk_scalar_fluxes.jl
```diff
@@ -63,22 +69,33 @@ Base.summary(bf::BulkSensibleHeatFluxFunction) =
...
 
 # With filtered scalar: read from the 2D filtered field
-@inline function bulk_sensible_heat_difference(i, j, grid, ::PotentialTemperatureFlux, T₀, constants, fields, fs::FilteredSurfaceScalar)
+@inline function bulk_sensible_heat_difference(i, j, grid, ::PotentialTemperatureFlux, bf, T₀, fields, fs::FilteredSurfaceScalar)
     θ = @inbounds fs.field[i, j, 1]
-    return θ - T₀
+    p₀ = bf.surface_pressure
+    pˢᵗ = bf.standard_pressure
+    constants = bf.thermodynamic_constants
+    θ₀ = potential_temperature_from_temperature(T₀, p₀, constants; pˢᵗ)
```
glwagner: I don't think we want to use a kwarg in a kernel. Can we pass in `pˢᵗ` as a positional argument?

glwagner (changes requested):
I realized there is one change we need --- use

```julia
potential_temperature_from_temperature(T₀, p₀, pˢᵗ, constants)
```

instead of

```julia
potential_temperature_from_temperature(T₀, p₀, constants; pˢᵗ)
```

ie use positional arguments only in functions that appear in kernels.

glwagner (approved):

glwagner:
thank you @bischtob !

Issue #625, opened 2026-04-09
Bug Crawl: Overview Issue of Potential Bugs


bischtob:
False positives are possible; treat every item as a candidate.

[Part A](#a-checklist) is a flat checklist. [Part B](#b-details) has evidence and a suggested fix for each item.

**51 items** · 7 critical · 11 high · 24 medium · 9 low

| Module | C | H | M | L |
|:---|:---:|:---:|:---:|:---:|
| Thermodynamics |  | 1 | 3 | 1 |
| Compressible Equations |  | 2 | 1 | 1 |
| Atmosphere Models |  | 1 | 3 | 3 |
| Atmosphere Models (Diagnostics) | 2 | 1 |  |  |
| Boundary Conditions | 2 |  | 2 |  |
| Static Energy Formulations |  | 1 |  |  |
| Potential Temperature Formulations |  |  | 1 |  |
| Microphysics |  | 1 | 2 |  |
| Moist Air Buoyancies | 1 | 1 |  | 1 |
| Kinematic Driver | 2 |  |  |  |
| Forcings |  |  | 3 |  |
| Parcel Models |  |  | 1 |  |
| Celestial Mechanics |  |  | 1 |  |
| Time Stepping |  |  | 1 |  |
| Vertical Grids |  |  | 1 | 1 |
| Extensions (RRTMGP) |  | 1 | 1 |  |
| Extensions (CloudMicrophysics) |  |  | 1 |  |
| Examples |  | 2 | 3 | 2 |

---

#### Thermodynamics

- [x] **high** · `src/Thermodynamics/reference_states.jl` · adiabatic_hydrostatic_pressure uses potential temperature θ₀ where it should use the actual surface temperature T₀ = [#1](#issue-1)
- [ ] **medium** · `src/Thermodynamics/vapor_saturation.jl:302` · The secant iteration convergence criterion degenerates for low vapor pressure, producing silently incorrect dewpoint [#2](#issue-2)
- [x] **medium** · `src/Thermodynamics/clausius_clapeyron.jl` · The intermediate Clausius-Clapeyron equation in the docstring is mathematically wrong and internally inconsistent with [#3](#issue-3)
- [ ] **medium** · `src/Thermodynamics/dynamic_states.jl` · temperature_from_potential_temperature fails with MethodError when called with Float32 arguments due to a hardcoded [#4](#issue-4)
[...]

PR #626, opened 2026-04-10
Fix physics bugs found by Bug Crawl (#625)

kaiyuan-cheng:
## Summary

Fixes seven physics bugs identified in #625:

- **Item 1** (high): `adiabatic_hydrostatic_pressure` used potential temperature `θ₀` directly instead of converting to actual temperature `T₀ = θ₀(p₀/pˢᵗ)^κ`, and was missing the `pˢᵗ` argument
- **Item 10** (high): `max(2, Nz)` → `min(2, Nz)` in bottom-up moisture borrowing — the old code indexed out of bounds when `Nz < 2`
- **Item 17** (critical): ice latent heat sign in moist static energy was wrong: `+ℒⁱᵣqⁱ` → `−ℒⁱᵣqⁱ`
- **Item 18** (critical): `saturation_total_specific_moisture` was missing a factor of `ϵᵈᵛ = Rᵈ/Rᵛ` in the numerator, overestimating saturation specific humidity by ~1/0.622
- **Item 15** (low): `∂z_b` divided a face-located derivative `∂zᶜᶜᶠ(ϑ)` by a center-located `ϑ`, creating a grid-staggering inconsistency; replaced with the equivalent `g · ∂zᶜᶜᶠ(log ϑ)`
- **Item 14** (low): `moisture_fractions` fallback for `NamedTuple` microphysical states accumulated only liquid species (`qᶜˡ`, `qʳ`), silently setting ice-phase species to zero; now includes `qᶜⁱ` and `qˢ`
- **Item 3** (medium): Clausius-Clapeyron docstring had `log[(Δcᵝ/Rᵛ)(T/Tᵗʳ)]` instead of the correct `(Δcᵝ/Rᵛ) log(T/Tᵗʳ)`

Note: the `virtual_potential_temperature` Poisson exponent fix (item 11) was attempted but reverted — needs further investigation.

Also includes doctest value updates cascading from these fixes and a refactor of `moisture_fractions` to use `get` dispatch instead of `haskey` + ternary.

## Test plan

- [x] New test in `test/reference_states.jl` verifies closed-form hydrostatic pressure respects `pˢᵗ`
- [x] Updated `test/diagnostics.jl` passes with corrected `adiabatic_hydrostatic_pressure` signature
- [ ] Verify turbulence closure buoyancy gradients are physically consistent (BOMEX or RICO)
- [ ] Confirm mixed-phase microphysics states produce correct `MoistureMassFractions` with nonzero ice

[...]

PR #656, opened 2026-04-24
Fix formulation of virtual potential temperature / bugs from #625 bug crawl

kaiyuan-cheng:
## Summary

Addresses four items from the bug crawl in #625:

- **[item 24](https://github.com/NumericalEarth/Breeze.jl/issues/625#issue-24) (high)** · `src/StaticEnergyFormulations/static_energy_tendency.jl` · Flip the sign of the buoyancy flux term in the moist static energy tendency so it matches ∂ₜ(ρᵣe) + ∇·(ρᵣe u) = −ρᵣwb + Sₑ (Pauluis 2008), and correct the matching sign in `docs/src/anelastic_dynamics.md`.
- **[item 11](https://github.com/NumericalEarth/Breeze.jl/issues/625#issue-11) (medium)** · `src/AtmosphereModels/atmosphere_model_buoyancy.jl` · `virtual_potential_temperature` was using the moist Poisson exponent `Rᵐ/cᵖᵐ` instead of the standard dry-air exponent `Rᵈ/cᵖᵈ`. Switch to the dry exponent per the standard meteorological definition.
- **[item 25](https://github.com/NumericalEarth/Breeze.jl/issues/625#issue-25) (medium)** · `src/PotentialTemperatureFormulations/potential_temperature_formulation.jl` · Correct the sign of the exponential term in the `LiquidIcePotentialTemperatureFormulation` docstring (positive → negative).
- **[item 28](https://github.com/NumericalEarth/Breeze.jl/issues/625#issue-28) (medium)** · `src/Microphysics/dcmip2016_kessler.jl` · The Kessler kernel used `surface_pressure` (101325 Pa default) as the Exner function reference pressure, but Π = (p/pˢᵗ)^(Rᵐ/cᵖᵐ) is defined against `standard_pressure` (1e5 Pa default) — the convention used everywhere else in the codebase. With defaults, this produced a ~0.38% error in Π at p = 80 kPa and a ~1.2 K bias in the T ↔ θˡⁱ conversion driving saturation adjustment, condensation, and rain evaporation. The existing Kessler test masked the bug by setting `surface_pressure = 1e5`, making it equal to `standard_pressure`.

## Test plan

- [ ] CI passes (unit tests + doctests)
- [ ] Verify `virtual_potential_temperature` usage still round-trips at expected precision
- [ ] Confirm the sign change in `static_energy_tendency` does not break any existing regression reference
- [ ] Confirm Kessler microphysics tests still pass with the corrected reference pressure


kaiyuan-cheng:
@glwagner We discussed item 11 previously in #626. After further examination, I believe the bug is real. The purpose of virtual temperature is to allow moist air to be treated using the dry-air thermodynamic constants. The definition of virtual potential temperature can also be found in Atmospheric Convection by Kerry Emanuel:
image

glwagner:
> @glwagner We discussed item 11 previously in #626. After further examination, I believe the bug is real. The purpose of virtual temperature is to allow moist air to be treated using the dry-air thermodynamic constants. The definition of virtual potential temperature can also be found in Atmospheric Convection by Kerry Emanuel: image

ok, let me think about this. I guess this means that virtual temperature is not conserved, but perhaps it is the quantity that shows up in the buoyancy computation (which I believe is how we use it)

glwagner:
Okay I see, so the tricky part is that _potential temperature_ does include the vapor dependence of the exponent, but "virtual potential temperature" does not (by definition). Is that what you see @kaiyuan-cheng ?

I honestly don't completely understand why this would be the case (eg the argument that Emmanuel develops involving the virtual potential temperature around equation 6.1.5 assumes that two air masses have the same pressure anyways), but it is the case that Emmanuel claims this version of virtual potential temperature with dry exponent is the relevant quantity for buoyancy.

glwagner:
> The purpose of virtual temperature is to allow moist air to be treated using the dry-air thermodynamic constants.

I don't think this is quite right. The purpose of virtual potential temperature is that it can be used to compute stability / buoyancy frequency, right? Do you agree @kaiyuan-cheng ?

kaiyuan-cheng:
> > @glwagner We discussed item 11 previously in #626. After further examination, I believe the bug is real. The purpose of virtual temperature is to allow moist air to be treated using the dry-air thermodynamic constants. The definition of virtual potential temperature can also be found in Atmospheric Convection by Kerry Emanuel: image
> 
> ok, let me think about this. I guess this means that virtual temperature is not conserved, but perhaps it is the quantity that shows up in the buoyancy computation (which I believe is how we use it)

It depends on the scope of conservation. Virtual potential temperature is conserved for unsaturated air under adiabatic processes.

docs/src/anelastic_dynamics.md
```diff
@@ -63,11 +63,11 @@ and
 According to [Pauluis2008](@citet), the moist static energy obeys
 
 ```math
-\partial_t(ρᵣ e) + \boldsymbol{\nabla \cdot}\, (ρᵣ e \boldsymbol{u}) = ρᵣ w b + S_e ,
+\partial_t(ρᵣ e) + \boldsymbol{\nabla \cdot}\, (ρᵣ e \boldsymbol{u}) = - ρᵣ w b + S_e ,
 ```
 
 with vertical velocity ``w``, buoyancy ``b`` as above, and ``S_e`` including microphysical and external energy sources/sinks.
-The ``ρᵣ w b`` term is the buoyancy flux that links the energy and momentum budgets in the anelastic limit.
+The ``-ρᵣ w b`` term is the buoyancy flux that links the energy and momentum budgets in the anelastic limit.
```
glwagner:
```suggestion
The ``ρᵣ w b`` term is the buoyancy flux that links the energy and momentum budgets in the anelastic limit.
```

I agree with the sign change in the equation, but I propose defining "ρᵣ wb" as the buoyancy flux.

glwagner (approved):

kaiyuan-cheng:
> > The purpose of virtual temperature is to allow moist air to be treated using the dry-air thermodynamic constants.
> 
> I don't think this is quite right. The purpose of virtual potential temperature is that it can be used to compute stability / buoyancy frequency, right? Do you agree @kaiyuan-cheng ?

Certainly, virtual potential temperature can serve to compute stability, but I believe the main purpose of introducing virtual temperature is the simplicity it brings to thermodynamic equations by using dry-air constants.

glwagner:
What equations are those? (I'm curious)

kaiyuan-cheng:
> What equations are those? (I'm curious)

Anyone related to the equation of state, such as virtual potential temperature and CAPE calculation.

glwagner:
CAPE calculation is in terms of buoyancy right? So for that application, its important that virtual potential temperature is correctly formulated wrt to buoyancy.

giordano:
Any tests we could have to prevent regressions? I'm surprised this change didn't even require touching any existing test, sounds like something isn't covered by tests?

glwagner:
Good point @giordano, we don't have a test for this. We could have a simple one that builds a state in `AtmosphereModel` and then evaluates that $\theta_v$ takes the expected value(s)

PR #440, opened 2026-01-30
[Example] Idealized tropical cyclone with rainbands from Moon and Nolan (2010)

nrb171:
WIP implementation of Moon and Nolan (2010) with Yu and Didlake's (2019) modified rainband heating profiles applied. 

Closes #439 .

PR #657, opened 2026-04-24
[Example] Tropical cyclone with stationary stratiform rainband (YD19)

### LINKED ISSUE #439: Modified Moon and Nolan (2010) idealized TC with rainband setup
[Moon and Nolan](https://journals.ametsoc.org/doi/10.1175/2010JAS3171.1) and later [Yu and Didlake (2019)](https://doi.org/10.1175/JAS-D-18-0335.1) utilize an axisymmetric vortex superimposed with an asymmetric rainband. These studies utilized 3DVAPS and WRF. Specifically of interest to my research, Yu and Didlake utilized this framework to study the impact of different TC rainband heating profiles on the larger-scale storm and circulation. 

Here's a figure from Moon and Nolan:
Image
Moon and Nolan (2010), Figure 23

And another from Yu and Didlake:
Image

Given the complexity of setting up these types of TC simulations, an example using these initial conditions would be a great benefit to the research community, while also providing additional testing scenarios for Breeze development.

bischtob:
## Summary

Self-contained literate example that reproduces the idealized tropical-cyclone rainband experiment of Yu & Didlake (2019, *J. Atmos. Sci.* **76**, 3169–3189) — modeled on the form of our mature examples (`bomex.jl`, `splitting_supercell.jl`, `dry_thermal_bubble.jl`).

The simulation asks YD19's question: given a mature hurricane, what happens when you paint a steady stationary heating pattern in one of its stratiform rainbands? The answer — both in YD19's full-physics WRF run and in the Breeze anelastic reproduction here — is a quadrupole of secondary-circulation anomalies and a few-m/s dipole in the tangential wind.

Buidls on earlier WIP #440 by @nrb171. Closes #439.

@nrb171 feel free to do whatever you want with this.

## What's in the PR (4 files, +1465 / −0)

| File | Change |
|---|---|
| `examples/tropical_cyclone_with_rainband.jl` | +1384 — the literate example |
| `docs/src/breeze.bib` | +66 — 6 new bib entries (YuDidlake2019, MoonNolan2010, SternNolan2009, Nolan2001, Jordan1958, Emanuel1986) |
| `examples/.gitignore` | +14 — keeps `output_tc_*/`, `*.jld2`, `*.mp4`, `*.png` out of git (the example writes ~5 GB of output) |
| `docs/make.jl` | +1 — registers the example (`build_always=false, gpu=true`, consistent with the other heavy GPU examples) |

## Simulation

Single hardwired configuration, 214² × 75 cells on a 642 km × 642 km × 25 km periodic-in-x-y box at 3 km horizontal resolution (Δz ≈ 333 m). Three back-to-back 24 h stages run end-to-end in one script:

1. **Spinup** — balanced vortex IC relaxes under the upper-level sponge only.
2. **Control** — 24 h continuation from the post-spinup state, no heating.
3. **Heated** — 24 h continuation from the same post-spinup state with the MN10 stratiform heating switched on.

The response is (heated − control) in the hour 5–7 analysis window. An F06 diagnostic tracks the response amplitude vs time; saturation past ~7 h reflects the absence of explicit horizontal diffusivity (documented known limitation).

## Physics

- **Environment**: Jordan (1958) West Indies mean hurricane-season sounding, tabulated inline and interpolated onto `z_centers`.
- **Vortex**: modified-Rankine profile (YD19 Eq. 2) with Stern & Nolan (2009) Eq. 4.4 RMW(z) slope, cos² taper outside 250 km for periodic-box compatibility.
- **Initial condition**: Picard iteration of the gradient-wind + hydrostatic + ideal-gas fixed point (Nolan 2001 / WRF `em_tropical_cyclone` initializer — the procedure YD19 §3a2 prescribe). Converges in ~15 iterations to gradient-wind residual ~10⁻³ Pa/m and hydrostatic residual ~10⁻⁵ Pa/m — a 10⁴× collapse over a one-shot linearized baseline.
- **Rainband heating**: YD19 Eq. 3 — spiral radial centerline `r_bs(λ,z) = [60 − 10(λ/(π/4))] km + z`, Gaussian radial shape with `σ_r = 6 km`, sinusoidal vertical shape with `σ_zs = 2 km` centered at `z_bs = 4 km`, super-Gaussian azimuthal window at `λ = −π/4`, 1-h linear ramp from zero to full strength.
- **Upper-level sponge**: WRF `damp_opt=2` analog. Rayleigh damping of `ρu`/`ρv`/`ρw` toward zero and `ρe` toward the reference dry-static-energy-density profile `ρᵣ(z)[cᵖᵈ Tᵣ(z) + gz]`, sin² ramp from `z = 20 km` to max at `z = 25 km`.

## Scientific outcome

Reproduced on NVIDIA H200 (GPU), 214² × 75, Float64:

| Metric | YD19 target | This reproduction |
|---|---|---|
| F02a surface v̄ peak | ≈ 40 m/s | **37.94 m/s** at r = 31.5 km |
| F02b θ′ peak | ≈ 10–12 K at 10–12 km | **7.63 K** at (r, z) = (1.5, 9.8) km |
| F03a response range (hour 5–7) | ≈ ±1.5 m/s | **[−1.00, +1.61] m/s** |
| F04 per-panel \|w′\| | < 1 m/s | **[0.213, 0.183, 0.217] m/s** at z = 6 / 3.6 / 2 km |

The quadrupole pattern and dipole in v̄′ are reproduced. The θ′ magnitude shortfall is a known formulation caveat (mass-coord vs z-coord Picard closure) that survives the 24 h spinup roughly unchanged.

## What the example teaches

- Building a balanced-vortex initial condition via Picard iteration (Nolan 2001 / WRF `em_tropical_cyclone`).
- Wiring a spatially structured, time-varying heat source into the energy equation with `Forcing`.
- Running three related stages in one script with in-memory state handoff between them.
- Isolating a forced response via a control experiment.

## Figures produced

- `F01_preflight.png` — vortex IC sanity check
- `F02ab_vortex.png` — basic-state vortex (YD19 Fig 2a,b)
- `F02cd_heating.png` — analytic heating (YD19 Fig 2c,d)
- `F03a_axisym_response.png` — axisymmetric response (YD19 Fig 3a)
- `F04_plan_response.png` — plan-view response (YD19 Fig 4a-c)
- `F05_cross_sections.png` — cross sections (YD19 Fig 5)
- `F06_response_timeseries.png` — response amplitude vs time (diagnostic)
- `response_w_z3km.mp4` — 25-frame animation of the w′ response at z ≈ 3 km

## Mature-example form

- Self-contained in one file: Jordan sounding, YD19 constants, kinematics, Picard solver, residuals diagnostic — no external `include`s.
- Dense Literate prose with five LaTeX math blocks (gradient wind, modified-Rankine, Picard fixed point, YD19 Eq. 3 heating, dry-static-energy target).
- 24 inline `[Author](@citet)` / `@cite` markers for every scientific reference; six bib entries added.
- `@ref` cross-refs for Breeze API (`ReferenceState`).
- `` embed for the animation so Documenter renders it inline.
- Analysis block uses streaming `FieldTimeSeries(...; backend=Oceananigans.OnDisk())` and preallocated `Float32` scratch buffers — peak RSS during analysis is bounded to a handful of 3D fields (~50 MB at this resolution) regardless of snapshot count.

## Implementation notes & caveats

- **Float64 default.** `Float32` was attempted and deferred: the anelastic pressure solve NaN's in `ρu` within 100–300 spinup steps at 3 km on GPU with the YD19 vortex intensity. Smoke-domain (12 km) Float32 runs were stable. Worth a separate Breeze-level stability investigation.
- **Explicit horizontal diffusivity is absent.** Without it, the response amplitude likely inflates past the paper's quasi-steady ±1.5 m/s after ~7 h (F06 diagnostic). YD19's WRF setup has a subgrid closure that damps the response; we don't. Analysis window is chosen accordingly.
- The example is gated `build_always=false, gpu=true`, it only runs when `BREEZE_BUILD_ALL_EXAMPLES=true` on a GPU runner.
- Worth also looking if stretched grid is appropriate here.

## Test plan

- [x] Full GPU run reproduces the numbers in the table above.
- [x] `Meta.parseall` passes.
- [x] Local Documenter draft build: Literate parses the script, all six new `@cite` markers resolve, `@ref` resolves for `ReferenceState`, no example-specific errors.
- [ ] CI green on a GPU-enabled run (requires reviewer to trigger `BREEZE_BUILD_ALL_EXAMPLES=true` or equivalent).
    - [x] Determine why `Float32` isn't working.
- [x] Run at a high resolution for benchmarking/pretty pictures purposes. 

## References

- Yu, C.-K. and Didlake, A. C. (2019). *J. Atmos. Sci.*, 76, 3169–3189. [doi:10.1175/JAS-D-18-0259.1](https://doi.org/10.1175/JAS-D-18-0259.1)
- Moon, Y. and Nolan, D. S. (2010). *J. Atmos. Sci.*, 67, 1779–1805. [doi:10.1175/2010JAS3284.1](https://doi.org/10.1175/2010JAS3284.1)
- Stern, D. P. and Nolan, D. S. (2009). *J. Atmos. Sci.*, 66, 3579–3600. [doi:10.1175/2009JAS3021.1](https://doi.org/10.1175/2009JAS3021.1)
- Nolan, D. S. (2001). *Physics of Fluids*, 13, 1724–1738. [doi:10.1063/1.1369646](https://doi.org/10.1063/1.1369646)
- Jordan, C. L. (1958). *J. Meteor.*, 15, 91–97.
- Emanuel, K. A. (1986). *J. Atmos. Sci.*, 43, 585–605.


bischtob:
F01_preflight
F02ab_vortex-2
F02cd_heating
F03a_axisym_response
F04_plan_response
F05_cross_sections
F06_response_timeseries

bischtob:
https://github.com/user-attachments/assets/05916bef-c02e-4ece-8cf4-3864952269bd

examples/tropical_cyclone_with_rainband.jl
```diff
@@ -0,0 +1,1384 @@
+# # [Tropical cyclone with a stationary stratiform rainband](@id tc_rainband_example)
+#
+# This example reproduces the idealized tropical-cyclone rainband experiments
+# of [YuDidlake2019](@citet), who asked a beautifully concrete question: if
+# you take a mature hurricane and paint a steady, stationary heating pattern
+# in one of its stratiform rainbands, what happens to the storm?
+#
+# Their answer — obtained from a full-physics WRF simulation — is a quadrupole
```
navidcy:
```suggestion
# Their answer — obtained from a full-physics Weather Research and Forecasting (WRF) model simulation — is a quadrupole
```

examples/tropical_cyclone_with_rainband.jl
```diff
@@ -0,0 +1,1384 @@
...
+# pattern of secondary-circulation anomalies: rising/sinking pairs that flank
+# the imposed heat source, and an accompanying dipole in the tangential wind
+# of a few m/s. We get the same pattern here with the Breeze anelastic core,
+# a Jordan ([Jordan1958](@cite)) tropical sounding, a Stern–Nolan balanced
```
navidcy:
```suggestion
# a Jordan [Jordan1958](@cite) tropical sounding, a Stern–Nolan balanced
```

examples/tropical_cyclone_with_rainband.jl
```diff
@@ -0,0 +1,1384 @@
...
+# vortex ([SternNolan2009](@citet)), and the Moon–Nolan ([MoonNolan2010](@cite))
```
navidcy:
```suggestion
# vortex [SternNolan2009](@citet), and the Moon–Nolan [MoonNolan2010](@cite)
```

examples/tropical_cyclone_with_rainband.jl
```diff
@@ -0,0 +1,1384 @@
+# # [Tropical cyclone with a stationary stratiform rainband](@id tc_rainband_example)
+#
+# This example reproduces the idealized tropical-cyclone rainband experiments
+# of [YuDidlake2019](@citet), who asked a beautifully concrete question: if
```
navidcy:
```suggestion
# of [YuDidlake2019](@citet) (hereafter YD19), who asked a beautifully concrete question: if
```

examples/tropical_cyclone_with_rainband.jl
```diff
@@ -0,0 +1,1384 @@
...
+# ## Figures produced
+#
+# | File                             | Content                                           |
+# |----------------------------------|---------------------------------------------------|
+# | `F01_preflight.png`              | vortex IC sanity check                            |
+# | `F02ab_vortex.png`               | basic-state vortex ([YuDidlake2019](@cite) Fig 2a,b) |
+# | `F02cd_heating.png`              | analytic heating ([YuDidlake2019](@cite) Fig 2c,d)  |
+# | `F03a_axisym_response.png`       | axisymmetric response ([YuDidlake2019](@cite) Fig 3a) |
+# | `F04_plan_response.png`          | plan-view response ([YuDidlake2019](@cite) Fig 4a-c)  |
+# | `F05_cross_sections.png`         | cross sections ([YuDidlake2019](@cite) Fig 5)     |
```
navidcy:
```suggestion
# | `F02ab_vortex.png`               | basic-state vortex ([YuDidlake2019](@citet); Fig. 2a,b) |
# | `F02cd_heating.png`              | analytic heating ([YuDidlake2019](@citet); Fig. 2c,d)  |
# | `F03a_axisym_response.png`       | axisymmetric response ([YuDidlake2019](@citet); Fig. 3a) |
# | `F04_plan_response.png`          | plan-view response ([YuDidlake2019](@citet); Fig. 4a-c)  |
# | `F05_cross_sections.png`         | cross sections ([YuDidlake2019](@citet); Fig. 5)     |
```

examples/tropical_cyclone_with_rainband.jl
```diff
@@ -0,0 +1,1384 @@
...
+using Random
+using CairoMakie
+using CUDA
+
+Oceananigans.defaults.FloatType = Float64
+
+Random.seed!(42)
+
+example_dir = @__DIR__
+
```
navidcy: I don't think we need this, right?
```suggestion
```

examples/tropical_cyclone_with_rainband.jl
```diff
@@ -0,0 +1,1384 @@
...
+using Oceananigans: Oceananigans
+using Oceananigans.Units
+using Oceananigans.Fields: interior
+using Printf
+using Random
+using CairoMakie
+using CUDA
+
+Oceananigans.defaults.FloatType = Float64
+
```
navidcy:
```suggestion
```

examples/tropical_cyclone_with_rainband.jl
```diff
@@ -0,0 +1,1384 @@
...
+Random.seed!(42)
+
+example_dir = @__DIR__
+
+# ## Jordan (1958) hurricane-season mean sounding
+#
+# The environment is a hydrostatic dry column taken from the Jordan
+# ([Jordan1958](@cite)) Table 5 mean West Indies sounding for the
```
navidcy:
```suggestion
# The environment is a hydrostatic dry column taken from Table 5 by
# [Jordan1958](@citet) mean West Indies sounding for the
```

examples/tropical_cyclone_with_rainband.jl
```diff
@@ -0,0 +1,1384 @@
...
+
+θ_env(z) = _linear_interpolate(jordan_z_m, jordan_θ_K,           z)
+T_env(z) = _linear_interpolate(jordan_z_m, jordan_T_C .+ 273.15, z)
+p_env(z) = _linear_interpolate(jordan_z_m, jordan_p_mb .* 100.0, z)
+
+# ## YD19 physical parameters
+#
+# The vortex parameters come straight out of [YuDidlake2019](@citet) §3a2:
+# a surface maximum wind of 43 m/s sitting at a radius of 31 km, decaying
+# outward as ``r^{-0.5}`` in the modified-Rankine sense. Above ``z ≈ 16`` km
```
navidcy:
```suggestion
# outward as ``r^{-1/2}`` in the modified-Rankine sense. Above ``z ≈ 16`` km
```

examples/tropical_cyclone_with_rainband.jl
```diff
@@ -0,0 +1,1384 @@
...
+# (the outflow level) the tangential wind is zero. The outer cos² taper
```
navidcy:
```suggestion
# (the outflow level) the tangential wind is zero. The outer ``\cos^2`` taper
```

examples/tropical_cyclone_with_rainband.jl
```diff
@@ -0,0 +1,1384 @@
...
+# between 250 and 300 km is *not* in the original paper — YD19 use a huge
+# non-periodic mother domain — but we need it to keep the vortex from
+# wrapping around on our periodic box.
+
+f             = 5.0e-5        # f-plane Coriolis parameter ([YuDidlake2019](@cite) §3a1)
+v_max_surface = 43.0          # initial surface v_max, m/s ([YuDidlake2019](@cite) §3a2)
+a_decay       = 0.5           # modified-Rankine decay exponent ([YuDidlake2019](@cite) Eq. 2)
+rmw_surface   = 31_000.0      # surface radius of maximum wind, m ([MoonNolan2010](@cite) App. A)
+z_vortex_top  = 16_000.0      # outflow reference level, m ([MoonNolan2010](@cite): v = 0 at RMW at z ≈ 15.9 km)
```
navidcy:
```suggestion
f             = 5.0e-5        # f-plane Coriolis parameter ([YuDidlake2019](@citet); §3a1)
v_max_surface = 43.0          # initial surface v_max, m/s ([YuDidlake2019](@citet); §3a2)
a_decay       = 0.5           # modified-Rankine decay exponent ([YuDidlake2019](@citet); Eq. 2)
rmw_surface   = 31_000.0      # surface radius of maximum wind, m ([MoonNolan2010](@citet); Appendix A)
z_vortex_top  = 16_000.0      # outflow reference level, m ([MoonNolan2010](@citet); v = 0 at RMW at z ≈ 15.9 km)
```

examples/tropical_cyclone_with_rainband.jl
```diff
@@ -0,0 +1,1384 @@
...
+#   - `figures/`   — F01-F06 PNGs (paper-reproduction outputs)
+
+output_dir    = joinpath(example_dir, "output_tc_rainband")
+snapshots_dir = joinpath(output_dir, "snapshots")
+figures_dir   = joinpath(output_dir, "figures")
+mkpath(snapshots_dir); mkpath(figures_dir)
+
+# ## Grid and architecture
+#
+# [YuDidlake2019](@citet) §3a1 uses a 3 km inner-nest resolution with a
```
navidcy:
```suggestion
# [YuDidlake2019](@citet) §3a1 use a 3 km inner-nest resolution with a
```

examples/tropical_cyclone_with_rainband.jl
```diff
@@ -0,0 +1,1384 @@
...
+# 25 km deep domain. We match that on a 642 km × 642 km periodic-in-x,y box:
+# 214² cells horizontally and 75 levels vertically (``Δz ≈ 333`` m). The run
+# prefers GPU and falls back to CPU if CUDA isn't functional.
+
+Δx              = 3_000.0
```
navidcy:
```suggestion
Δx              = 3kilometers
```

examples/tropical_cyclone_with_rainband.jl
```diff
@@ -0,0 +1,1384 @@
...
+# | `F02cd_heating.png`              | analytic heating ([YuDidlake2019](@cite) Fig 2c,d)  |
+# | `F03a_axisym_response.png`       | axisymmetric response ([YuDidlake2019](@cite) Fig 3a) |
+# | `F04_plan_response.png`          | plan-view response ([YuDidlake2019](@cite) Fig 4a-c)  |
+# | `F05_cross_sections.png`         | cross sections ([YuDidlake2019](@cite) Fig 5)     |
+# | `F06_response_timeseries.png`    | response amplitude vs time (diagnostic)           |
+
+using Breeze
+using Oceananigans: Oceananigans
+using Oceananigans.Units
+using Oceananigans.Fields: interior
```
navidcy:
```suggestion
```

examples/tropical_cyclone_with_rainband.jl
```diff
@@ -0,0 +1,1384 @@
...
+# [YuDidlake2019](@citet) §3a1 uses a 3 km inner-nest resolution with a
+# 25 km deep domain. We match that on a 642 km × 642 km periodic-in-x,y box:
+# 214² cells horizontally and 75 levels vertically (``Δz ≈ 333`` m). The run
+# prefers GPU and falls back to CPU if CUDA isn't functional.
+
+Δx              = 3_000.0
+Nx = Ny         = 214
+Nz              = 75
+Lx              = Nx * Δx
+Lz              = 25_000.0                    # YD19 §3a1
```
navidcy:
```suggestion
Lz              = 25kilometers                 # YD19 §3a1
```

examples/tropical_cyclone_with_rainband.jl
```diff
@@ -0,0 +1,1384 @@
...
+
+grid = RectilinearGrid(arch;
+                       size = (Nx, Ny, Nz), halo = (5, 5, 5),
+                       x = (-Lx/2, Lx/2), y = (-Lx/2, Lx/2), z = (0, Lz),
+                       topology = (Periodic, Periodic, Bounded))
+
+# ## Reference state and thermodynamic constants
+#
+# The [`ReferenceState`](@ref Breeze.Thermodynamics.ReferenceState) is
+# constructed from the Jordan θ(z) profile and anchored at the observed
```
navidcy:
```suggestion
# constructed from the Jordan ``θ(z)`` profile and anchored at the observed
```

examples/tropical_cyclone_with_rainband.jl
```diff
@@ -0,0 +1,1384 @@
...
+    i   = searchsortedfirst(xs, x_c)
+    i   = clamp(i, 2, length(xs))
+    x0, x1 = xs[i-1], xs[i]
+    y0, y1 = ys[i-1], ys[i]
+    return y0 + (y1 - y0) * (x_c - x0) / (x1 - x0)
+end
+
+θ_env(z) = _linear_interpolate(jordan_z_m, jordan_θ_K,           z)
+T_env(z) = _linear_interpolate(jordan_z_m, jordan_T_C .+ 273.15, z)
+p_env(z) = _linear_interpolate(jordan_z_m, jordan_p_mb .* 100.0, z)
```
navidcy:
```suggestion
T_env(z) = _linear_interpolate(jordan_z_m, jordan_T_C .+ 273.15, z)    # convert C -> K
p_env(z) = _linear_interpolate(jordan_z_m, jordan_p_mb .* 100.0, z)    # convert mb -> Pa
```

examples/tropical_cyclone_with_rainband.jl
```diff
@@ -0,0 +1,1384 @@
...
+    else
+        @warn "Missing control and/or heated snapshots — skipping F03a, F04, F05, animation"
+    end
+
+    @info "=== Analysis complete ==="
+end
+
+# ![](output_tc_rainband/figures/response_w_z3km.mp4)
+
+@info "=== Done ==="
```
navidcy:
```suggestion

```

giordano:
> Float64 default. Float32 was attempted and deferred: the anelastic pressure solve NaN's in ρu within 100–300 spinup steps at 3 km on GPU with the YD19 vortex intensity. Smoke-domain (12 km) Float32 runs were stable. Worth a separate Breeze-level stability investigation.

Unfortunately the L4 really struggles with intensive Float64 workloads.  This example alone (I disabled all the other GPU ones) took 80 minutes (before erroring during the plotting part due to a literate syntax error)

giordano:
https://github.com/NumericalEarth/Breeze.jl/actions/runs/24929552701/job/73004993549#step:7:82
```
┌ Error: Generated HTML over size_threshold limit: literated/tropical_cyclone_with_rainband.md
│     Generated file size: 568.2 (MiB)
│     size_threshold_warn: 512.0 (KiB)
│     size_threshold:      1.0 (MiB)
│     HTML file:           literated/tropical_cyclone_with_rainband/index.html
└ @ Documenter.HTMLWriter /usr/local/share/julia/packages/Documenter/AXNMp/src/html/HTMLWriter.jl:2000
```
Half a GB for a webpage is indeed quite a lot 😳

nrb171:
Ok. I think I found the issue. I'm testing it on my end before I push anything up.

giordano:
Docs building finally works: https://numericalearth.github.io/BreezeDocumentation/previews/PR657/literated/tropical_cyclone_with_rainband/  🥳 only problem now is that the new example takes over one hour to run (and it'll be a lot worse when we restore all other examples), because fp64 performance is pretty bad on this GPU 🫠

examples/tropical_cyclone_with_rainband.jl
```diff
@@ -0,0 +1,1626 @@
...
+using Oceananigans: Oceananigans
+using Oceananigans.Units
+using Oceananigans.Fields: interior
+using Printf
+using Random
+using CairoMakie
+using CUDA
+
+Oceananigans.defaults.FloatType = Float64
+Random.seed!(42)
```
giordano, line 49: For [actual reproducibility](https://github.com/NumericalEarth/Breeze.jl/pull/550) (at least on same device), need also to set the GPU seed: https://github.com/NumericalEarth/Breeze.jl/blob/b0e3e87afe3e46fba49060b032f96fc9f16f3a1e/examples/bomex.jl#L27-L29

nrb171: done :D

navidcy:
> Docs building finally works: https://numericalearth.github.io/BreezeDocumentation/previews/PR657/literated/tropical_cyclone_with_rainband/ 🥳 only problem now is that the new example takes over one hour to run (and it'll be a lot worse when we restore all other examples), because fp64 performance is pretty bad on this GPU 🫠

What if we run on Float32?

giordano:
Apparently it wasn't stable: https://github.com/NumericalEarth/Breeze.jl/pull/657#issuecomment-4317049819

examples/tropical_cyclone_with_rainband.jl
```diff
@@ -152,7 +152,7 @@ Nz = 75
 Lx = Nx * Δx
 Lz = 25kilometers                 # YD19 §3a1
 Δz = Lz / Nz
-sponge_rate = 0.003                       # ≈ WRF damp_opt=2 `dampcoef` (~333 s timescale)
+sponge_rate = 0.003               # ≈ WRF damp_opt=2 `dampcoef` (~333 s timescale)
```
glwagner: express this as 1/timescale?

navidcy:
```suggestion
sponge_rate = 1/333seconds # ≈ WRF damp_opt=2 `dampcoef`
```

nrb171:
## 500 m benchmark — full reproduction at convective scale

A second, higher-resolution reproduction was run end-to-end on the same example script, using a single GPU, to confirm that the 3 km result holds at a resolution at which deep-convective updrafts begin to be resolved and to give the figures greater visual fidelity.

image


### Configuration

|                | 3 km (PR baseline)       | 500 m (this benchmark)                    |
|----------------|--------------------------|-------------------------------------------|
| Grid           | 214 × 214 × 75          | 1284 × 1284 × 75 (~1.24 × 10⁸ cells)     |
| Δx, Δy         | 3 km                     | 500 m                                     |
| Δz             | 333 m                    | 333 m                                     |
| Domain         | 642 × 642 × 25 km       | 642 × 642 × 25 km                         |
| Precision      | Float64                  | Float64                                   |
| Hardware       | NVIDIA H200 (1 GPU)     | NVIDIA H100 80 GB (1 GPU)                 |
| Stages         | spinup + control + heated, 24 h each | spinup + control + heated, 24 h each |

No physics, IC, sponge, or analysis-window settings were changed — only the resolution. The Picard IC solver converged in the same 15 iterations to the same residuals (gradient-wind 1.48 × 10⁻³ Pa/m, hydrostatic 2.18 × 10⁻⁵ Pa/m).

### Performance

| Stage          | Steps  | Δt range       | Wall time |
|----------------|--------|----------------|-----------|
| Spinup (24 h)  | 16,333 | 2.2 → 5.7 s   | 1.70 h    |
| Control (24 h) | 15,099 | 2.2 → 5.9 s   | 1.57 h    |
| Heated (24 h)  | 15,842 | 2.2 → 5.6 s   | 1.68 h    |
| **Total**      | 47,274 | —              | **≈ 5.0 h for 72 h simulated** |

≈ 14× faster than real time on one H100 at this grid (~3.3 × 10⁸ cell-updates/s). Snapshot output is 3 × 57 GB = 170 GB.

### Scientific outcome

| Metric                              | YD19 target                  | 3 km                                    | 500 m                                        |
|-------------------------------------|------------------------------|-----------------------------------------|----------------------------------------------|
| F02a global v̄ peak                  | ≈ 40 m/s                    | 37.94 m/s @ r = 31.5 km, z = surf      | 30.19 m/s @ (r, z) = (43.2, 5.8) km         |
| F02a surface v̄ peak                 | ≈ 40 m/s                    | 37.94 m/s                               | 18.88 m/s @ r = 18.8 km                     |
| F02b θ′ peak                        | ≈ 10–12 K at z = 10–12 km   | 7.63 K @ (1.5, 9.8) km                 | 7.08 K @ (0.8, 11.2) km                     |
| F03a response range (hour 5–7)      | ≈ ±1.5 m/s                  | [−1.00, +1.61] m/s                      | [−1.97, +1.98] m/s                           |
| F04 per-panel \|w′\| (z = 6 / 3.6 / 2 km) | < 1 m/s               | [0.213, 0.183, 0.217] m/s              | [0.576, 0.433, 0.431] m/s                   |

The qualitative outcome — quadrupole secondary-circulation anomalies and a tangential-wind dipole of a few m/s — is reproduced and is visually cleaner at 500 m.

**Quantitative differences worth flagging:**

- **Storm-scale weakening**: This is due to the inclusion of boundary-layer fluxes in the 500 m resolution simulation. Without (1) open boundary conditions and (2) radial inflow initial conditions, the boundary layer essentially spins down. This doesn't affect the simulation results because the rainband parameterization is largely above the boundary layer, but the boundary-layer processes will now be toggled off by default. 

- **F04 color range auto-scales** from ±0.75 m/s (3 km) to ±4 m/s (500 m); a raw-w plan-view figure (`F04b_plan_w_raw.png`) is included alongside the response figure at this resolution.

- Outside of these differences, the results are in line with the literature, with the caveats mentioned above. 

### Figures produced (500 m)

Same set as the 3 km run, plus F04b: 
`F01_preflight`
image

 `F02ab_vortex`
image

 
 `F02cd_heating`
 (same as above, omitted)
 
 `F03a_axisym_response`
 image

 `F04_plan_response`
image 

 `F04b_plan_w_raw`
 (excluded, mostly diagnostic for now)
 
 `F05_cross_sections`
 (above)
  
 `F06_response_timeseries`
image


## Some final thoughts
- There's a lot of missing physics in here, and adding them is a substantial task. While the paper's results are replicated, the final product is highly idealized.
- Here's a brief list of what I think we need before real TC simulations can become a thing in Breeze. Many of these are being actively developed or may already exist in some form, so the list may become obsolete at some point. 
    - [ ] open boundary conditions (allowing more complex idealized simulations with inflow/outflow resolved)
    - [ ] domain nesting or temporally evolving boundary conditions informed by other forecast/reanalysis products (real TC support)
    - [x] sea-surface flux/friction models
    - [ ] land flux/friction models
    - [ ] a high-quality, stable, two-moment (or better) microphysics scheme
    - [x] accurate horizontal/vertical diffusion

glwagner:
We should figure out what makes this unstable at Float32 I think.

glwagner:
> Here's a brief list of what I think we need before real TC simulations can become a thing in Breeze. Many of these are being actively developed or may already exist in some form, so the list may become obsolete at some point.

* COARE bulk formula for air-sea fluxes are supported in NumericalEarth, and land surface is coming soon
* Are we sure we want to use horizontal diffusion for TC simulations? (We have LES closures, but I am not sure this is what's meant.)

nrb171:
I'll see what I can do about Float32.

> * COARE bulk formula for air-sea fluxes are supported in NumericalEarth, and land surface is coming soon

Cool. I'll take a look at using those air-sea fluxes. 

> * Are we sure we want to use horizontal diffusion for TC simulations? (We have LES closures, but I am not sure this is what's meant.)

Depends on the resolution. For a 500 m resolution simulation, no, the LES itself should be enough. I'm a bit worried about the intermediate scales (3-10 km resolution). If you already have a solution for this, just let me know. There's a lot in this ecosystem, and it's hard to keep track of everything.

glwagner:
> Depends on the resolution. For a 500 m resolution simulation, no, the LES itself should be enough. I'm a bit worried about the intermediate scales (3-10 km resolution). If you already have a solution for this, just let me know. There's a lot in this ecosystem, and it's hard to keep track of everything.

Turbulence closure development is definitely planned, including surface layer modeling (applicable also to LES), as well as boundary layer vertical mixing parameterization (such as a Mellor-Yamada type parameterization, which can be re-used from Oceananigans). In Oceananigans this is called `TKEDissipationVerticalDiffusivity()`. The particular implementation is called "k-epsilon" in some communities.

Both of the above options should be classified as "vertical flux parameterization". I would expect them to apply to models with dx > 100 m, up to perhaps dx ~ 1-4 km.

Beyond that, further accuracy from true _terra incognita_ type turbulence modeling that includes horizontal fluxes as well is a research direction that may be pursued. However, there is no roadmap for turbulence/convection modeling (or cloud parameterization) with dx = O(10 km) and coarser, yet. It's possible that we will pursue an EDMF scheme similar to what has been implemented in GFDL's model. There is some hope that this will have some scale-awareness, but it remains to be proven, eg:

https://agupubs.onlinelibrary.wiley.com/doi/full/10.1029/2025MS005347

nrb171:
> Turbulence closure development is definitely planned, including surface layer modeling (applicable also to LES), as well as boundary layer vertical mixing parameterization (such as a Mellor-Yamada type parameterization, which can be re-used from Oceananigans). In Oceananigans this is called `TKEDissipationVerticalDiffusivity()`. The particular implementation is called "k-epsilon" in some communities.

Ah, this is exactly what I needed. Thank you! :D

nrb171:
> We should figure out what makes this unstable at Float32 I think.

I reran a couple of tests to see if I can nail down where the `Float32` instability comes from.

## Float32 (in)stability

Tested whether the Float32 anelastic NaN survives the recent main updates (notably #680, consistent `FloatType` in `_update_pressure_and_average!`).

**Setup.** 3 km / 214² × 75 / Float32 / no surface physics, on the merged tree. T4 GPU. Per-iteration sample of `max|ρu|, max|ρv|, max|ρw|, max|ρe|` for the first 250 spinup steps.

**Result.** Unfortunately, the simulation NaN'd at iter 100.

| iter | max\|ρu\| | max\|ρv\| | max\|ρw\| | max\|ρe\| |
|---|---|---|---|---|
| 95 | 4.83e+01 | 4.83e+01 | 2.64e-01 | 3.565e+05 |
| 96 | 4.83e+01 | 4.83e+01 | 2.75e-01 | 3.565e+05 |
| **97** | **1.74e+02** | **1.65e+02** | **4.90e+01** | 3.565e+05 |
| 98 | 1.25e+06 | 1.92e+06 | 4.87e+05 | 1.32e+10 |
| 99 | NaN | NaN | NaN | NaN |

ρw jumps by a factor of about 184 at iter 97, causing the rest of the simulation to destabilize. Could this be an issue with the pressure solver? @glwagner what do you recommend I try out?

glwagner:
> Could this be an issue with the pressure solver? @glwagner what do you recommend I try out?

I think because it works with Float64 but not Float32, it is a round-off error somewhere. For another case, we found that rewriting the buoyancy term stabilized a simulation that was blowing up with F32... so my guess would be something along those lines, possibly in the microphysics.

I think it is unlikely to be a pressure solver specific issue, because many other cases work with this pressure solver at F32. Also the pressure solver algorithm is fairly simple arithmetic so I don't think there are many opportunities to rewrite expressions to reduce round off error.

Apart from inspecting the source code a few test simulations might yield some insight:

- increase resolution
- reduce WENO order
- change the microphysics
- try acoustic substepping?

nrb171:
> * try acoustic substepping?

The simulation worked when I switched to acoustic substepping! 

(everything else failed at the same iteration)

glwagner:
> > * try acoustic substepping?
> 
> The simulation worked when I switched to acoustic substepping!
> 
> (everything else failed at the same iteration)

wow! that could point to several things, but the buoyancy calculation may actually be implicated here (they are different between the two, because anelastic buoyancy is a slightly different formulation). I think the divergence damping in the substepping is also stabilizing.

nrb171:
@glwagner What else do we need to merge this?

docs/make.jl
```diff
@@ -36,11 +36,11 @@ examples = [
     Example("Stratified dry thermal bubble", "dry_thermal_bubble"; build_always=true, gpu=false),
     Example("Cloudy thermal bubble", "cloudy_thermal_bubble"; build_always=true, gpu=false),
     Example("Cloudy Kelvin-Helmholtz instability", "cloudy_kelvin_helmholtz"; build_always=true, gpu=false),
-    Example("Shallow cumulus convection (BOMEX)", "bomex"; build_always=true, gpu=true),
+    # Example("Shallow cumulus convection (BOMEX)", "bomex"; build_always=true, gpu=true),
     Example("Precipitating shallow cumulus (RICO)", "rico"; build_always=false, gpu=false),
     Example("Convection over prescribed sea surface temperature (SST)", "prescribed_sea_surface_temperature"; build_always=true, gpu=false),
     Example("Inertia gravity wave: many time steppers", "inertia_gravity_wave"; build_always=true, gpu=false),
-    Example("Neutral atmospheric boundary layer", "neutral_atmospheric_boundary_layer"; build_always=false, gpu=true),
+    # Example("Neutral atmospheric boundary layer", "neutral_atmospheric_boundary_layer"; build_always=false, gpu=true),
```
giordano: These changes should eventually be reverted before merge.

examples/tropical_cyclone_with_rainband.jl
```diff
@@ -0,0 +1,1989 @@
...
+    20743.0, 22139.0, 23971.0,
+]
+
+jordan_T_C = Float64[
+    26.3, 26.0, 23.0, 19.8, 17.3, 14.6, 11.8, 8.6, 5.1, 1.4,
+    -2.5, -6.9, -11.9, -17.7, -24.8, -33.2, -43.3, -55.2, -61.5, -67.6,
+    -72.2, -73.5, -69.8, -63.9, -60.6, -57.3, -54.0,
+]
+
+jordan_θ_K = Float64[
```
giordano:
```suggestion
const jordan_p_mb = Float64[
    1015.1, 1000.0, 950.0, 900.0, 850.0, 800.0, 750.0, 700.0, 650.0, 600.0,
    550.0, 500.0, 450.0, 400.0, 350.0, 300.0, 250.0, 200.0, 175.0, 150.0,
    125.0, 100.0, 80.0, 60.0, 50.0, 40.0, 30.0,
]

const jordan_z_m = Float64[
    0.0, 132.0, 583.0, 1054.0, 1547.0, 2063.0, 2609.0, 3182.0,
    3792.0, 4442.0, 5138.0, 5888.0, 6703.0, 7595.0, 8581.0, 9682.0,
    10935.0, 12396.0, 13236.0, 14177.0, 15260.0, 16568.0, 17883.0, 19620.0,
    20743.0, 22139.0, 23971.0,
]

const jordan_T_C = Float64[
    26.3, 26.0, 23.0, 19.8, 17.3, 14.6, 11.8, 8.6, 5.1, 1.4,
    -2.5, -6.9, -11.9, -17.7, -24.8, -33.2, -43.3, -55.2, -61.5, -67.6,
    -72.2, -73.5, -69.8, -63.9, -60.6, -57.3, -54.0,
]

const jordan_θ_K = Float64[
```
Also, do these really need to be `Float64`?  Most of these are integers

giordano (commented):
The example still takes 60 minutes, which is an improvement from 80 minutes 🥲  I see there are `Float64` constants, I suspect they're getting inside the kernels?

Can we reduce the wall of prints in https://numericalearth.github.io/BreezeDocumentation/previews/PR657/literated/tropical_cyclone_with_rainband/#F01-%E2%80%94-vortex-IC-preflight-(fast-sanity-check;-always-runs-in-spinup)?  I think that involves reducing the frequency of the prints.  One has to scroll a lot, not fun when reading the docs.

examples/tropical_cyclone_with_rainband.jl
```diff
@@ -0,0 +1,1989 @@
...
+#
+# Run outputs live under `examples/output_tc_rainband/figures/` — F01-F06
+# PNGs (paper-reproduction outputs). Hourly state snapshots stay in host RAM
+# during the run (~6.6 GB per stage of host memory; fits comfortably on a
+# 128 GB host) and are released after stage 4's analysis, so the on-disk
+# footprint is only the figures themselves.
+
+output_dir = joinpath(@__DIR__, "output_tc_rainband")
+figures_dir = joinpath(output_dir, "figures")
+mkpath(figures_dir)
```
giordano:
```suggestion
mkpath(figures_dir)
nothing #hide
```

bischtob:
@nrb171 @glwagner is this thing ready? asking cause it might be good jumping off point for conversations with collaborators. would be wonderful if we could get it in.

nrb171:
> @nrb171 @glwagner is this thing ready? asking cause it might be good jumping off point for conversations with collaborators. would be wonderful if we could get it in.

Yes, it should be ready now. 
It was crashing yesterday. It seems the pull from main that @giordano just did resolved those issues.

giordano:
https://github.com/NumericalEarth/Breeze.jl/pull/657#discussion_r3275115983 still needs to be addressed.  https://numericalearth.github.io/BreezeDocumentation/previews/PR657/literated/tropical_cyclone_with_rainband/ looks overall good (can be later tweaked if needed)

docs/make.jl
nrb171: Thanks for the reminder. Fixed in [1f0264a](https://github.com/NumericalEarth/Breeze.jl/pull/657/commits/1f0264ab67f8467ec9384f7d91ff66eb52c83baf).

glwagner:
Are we still running F64?

giordano:
No, it's fp32 now: https://github.com/NumericalEarth/Breeze.jl/blob/5b1fe064159c97655b2f99f09714bcb7e4246e3d/examples/tropical_cyclone_with_rainband.jl#L62  Still takes an hour, but I guess the simulation is just expensive.

examples/tropical_cyclone_with_rainband.jl
```diff
@@ -0,0 +1,1989 @@
...
+    26.3, 26.0, 23.0, 19.8, 17.3, 14.6, 11.8, 8.6, 5.1, 1.4,
+    -2.5, -6.9, -11.9, -17.7, -24.8, -33.2, -43.3, -55.2, -61.5, -67.6,
+    -72.2, -73.5, -69.8, -63.9, -60.6, -57.3, -54.0,
+]
+
+const jordan_θ_K = Float32[
+    298.0, 299.0, 300.0, 302.0, 304.0, 307.0, 309.0, 312.0, 315.0, 318.0,
+    321.0, 324.0, 328.0, 332.0, 335.0, 338.0, 342.0, 345.0, 348.0, 354.0,
+    364.0, 386.0, 418.0, 468.0, 500.0, 542.0, 597.0,
+]
```
glwagner: It's probably not necessary to use `const` or convert to `Float32` here. It happens under the hood.

examples/tropical_cyclone_with_rainband.jl
```diff
@@ -0,0 +1,1989 @@
...
+
+function linear_interpolate(xs::AbstractVector, ys::AbstractVector, x::Real)
+    x_c = clamp(x, first(xs), last(xs))
+    i = searchsortedfirst(xs, x_c)
+    i = clamp(i, 2, length(xs))
+    x0, x1 = xs[i - 1], xs[i]
+    y0, y1 = ys[i - 1], ys[i]
+    return y0 + (y1 - y0) * (x_c - x0) / (x1 - x0)
+end
```
glwagner: This should not be necessary. Oceananigans has interpolation capabilities.

examples/tropical_cyclone_with_rainband.jl
```diff
@@ -0,0 +1,1989 @@
...
+
+## Precompute the K- and Pa-converted soundings so `T_env`/`p_env` don't
+## allocate a fresh broadcasted array on every call, and so the type
+## propagates through interpolation as Float32.
+const jordan_T_K = jordan_T_C .+ 273.15f0
+const jordan_p_Pa = jordan_p_mb .* 100f0
```
glwagner: This is ok, but it might be more sane simply to define the sounding in K in the first place. Also I don't think `const` is necessary here

examples/tropical_cyclone_with_rainband.jl
```diff
@@ -0,0 +1,1989 @@
...
+# July–October hurricane season. It's the same climatological profile YD19
+# use (and about a million other idealized tropical cyclone studies since).
+# Columns are pressure (mb), geopotential height (m), temperature (°C), and
+# potential temperature (K).
+
+const jordan_p_mb = Float32[
+    1015.1, 1000.0, 950.0, 900.0, 850.0, 800.0, 750.0, 700.0, 650.0, 600.0,
+    550.0, 500.0, 450.0, 400.0, 350.0, 300.0, 250.0, 200.0, 175.0, 150.0,
+    125.0, 100.0, 80.0, 60.0, 50.0, 40.0, 30.0,
+]
```
glwagner: can you define the z-levels instead of p here?

examples/tropical_cyclone_with_rainband.jl
```diff
@@ -0,0 +1,1989 @@
...
+
+## Precompute the K- and Pa-converted soundings so `T_env`/`p_env` don't
+## allocate a fresh broadcasted array on every call, and so the type
+## propagates through interpolation as Float32.
+const jordan_T_K = jordan_T_C .+ 273.15f0
+const jordan_p_Pa = jordan_p_mb .* 100f0
+
+θ_env(z) = linear_interpolate(jordan_z_m, jordan_θ_K, z)
+T_env(z) = linear_interpolate(jordan_z_m, jordan_T_K, z)
+p_env(z) = linear_interpolate(jordan_z_m, jordan_p_Pa, z)
```
glwagner: A more robust way to do this might be

```julia
sounding_grid = RectilinearGrid(size=Nz; z=jordan_z, topology=(Flat, Flat, Bounded))
jordan_θ = CenterField(sounding_grid)
interior(jordan_θ) .= [298.0, 299.0, 300.0, 302.0, 304.0, 307.0, 309.0, 312.0, 315.0, 318.0,
    321.0, 324.0, 328.0, 332.0, 335.0, 338.0, 342.0, 345.0, 348.0, 354.0,
    364.0, 386.0, 418.0, 468.0, 500.0, 542.0, 597.0]
```

then later on

```julia
set!(model, θ=jordan_θ)
```

will just work.

examples/tropical_cyclone_with_rainband.jl
```diff
@@ -0,0 +1,1989 @@
...
+#
+# The vortex parameters utilize classic modified-Rankine vortex structure, following the parameters used in [YuDidlake2019](@citet) §3a2:
+# a surface maximum wind of 43 m/s with a radius of maximum wind set at 31 km, decaying
+# outward as ``r^{-1/2}`` in the modified-Rankine sense. Above ``z ≈ 16`` km
+# (the outflow level) the tangential wind is zero. The outer ``\cos^2()`` taper
+# between 250 and 300 km is *not* in the original paper. Since our domain is periodic,
+# we need to impose this to limit unrealistic stress at the domain boundaries.
+
+f = 5.0e-5                     # f-plane Coriolis parameter, 1/s ([YuDidlake2019](@citet); §3a1)
+v_max_surface = 43.0           # initial surface v_max, m/s ([YuDidlake2019](@citet); §3a2)
```
glwagner:
```suggestion
f = 5e-5                     # f-plane Coriolis parameter, 1/s ([YuDidlake2019](@citet); §3a1)
v_max_surface = 43           # initial surface v_max, m/s ([YuDidlake2019](@citet); §3a2)
```

examples/tropical_cyclone_with_rainband.jl
```diff
@@ -0,0 +1,1989 @@
...
+Δx = 3000meters
+Lx = 642kilometers
+Nx = Ny = Int(Lx / Δx)
+Nz = 75
+Lz = 25kilometers                 # YD19 §3a1
+Δz = Lz / Nz
+sponge_rate = 1f0 / Float32(333seconds) # ≈ WRF damp_opt=2 `dampcoef`
+stage_stop_time = 24hours
+
+arch = CUDA.functional() ? GPU() : CPU()
```
glwagner: should we delete this? @giordano

examples/tropical_cyclone_with_rainband.jl
```diff
@@ -0,0 +1,1989 @@
...
+reference_state = ReferenceState(
+    grid, constants;
+    surface_pressure = p_env(0.0),
+    potential_temperature = θ_env
+)
+
+## `interior(...)` below is for host-side reductions (balance iteration,
+## azimuthal binning, state save/restore), NOT for Makie plotting — so the
+## examples-rules ban on `interior(field, ...)` in heatmap!/lines! calls
+## doesn't apply. Plotting is done from already-reduced Arrays.
```
glwagner: wut

nrb171: lol. the robot speaks. I'm going through and cleaning up the comments (for like the 4th time...).

examples/tropical_cyclone_with_rainband.jl
nrb171: We need both z and p to perform the interpolation later.

bischtob:
I know this is a long example, but it's also epic.

examples/tropical_cyclone_with_rainband.jl
nrb171: Runic prefers 5.0e-5, so I left it that way.

examples/tropical_cyclone_with_rainband.jl
nrb171: Ok, I did something similar. Thanks for the suggestion. I am not used to having nice grids like Oceananigans haha

examples/tropical_cyclone_with_rainband.jl
giordano: Probably yes, I assume this would be way too expensive on CPU anyway.

examples/tropical_cyclone_with_rainband.jl
nrb171: Ok, I cleaned that up.

giordano:
Images aren't actually shown.  I guess 719e9a2a9b445cc577ee5e25fb06760639956f72 removed _all_ images?

examples/tropical_cyclone_with_rainband.jl
```diff
@@ -702,265 +702,259 @@ nothing # hide
...
+v_sc = similar(u_sc)
+w_sc = similar(u_sc)
+T_sc = similar(u_sc)
+
+## In-place centered interpolation (Arakawa C-grid → Centers).
+function center_u!(out::AbstractArray{Float32, 3}, src)
+    Nxs, Nys, Nzs = size(out)
+    return @inbounds for k in 1:Nzs, j in 1:Nys, i in 1:Nxs
+        ip = i == Nxs ? 1 : i + 1
+        out[i, j, k] = Float32((src[i, j, k] + src[ip, j, k]) / 2)
```
glwagner: you can also write

```julia
u_fcc = XFaceField(grid)
u_ccc = CenterField(grid)
set!(u_ccc, u_fcc)
```

this will interpolate under the hood.

nrb171:
@glwagner what does this example need to be merged?

glwagner:
@nrb171 this is ready to go EXCEPT that it takes too long (the docs build takes 86 min). We either need more computational resources / more powerful GPU for the docs, or we need to reduce the cost of the example.

I think we should take the latter approach for now.

There is a bit of clean up that I could do as well, but we can defer for a follow up PR.

examples/tropical_cyclone_with_rainband.jl
```diff
@@ -0,0 +1,1059 @@
...
+z_centers = collect(range(Δz / 2, step = Δz, length = Nz))
+
+@info @sprintf(
+    "Grid: %d × %d × %d   Δx = %.1f km   Lz = %.1f km   Δz = %.1f m",
+    Nx, Ny, Nz, Δx / 1.0e3, Lz / 1.0e3, Δz
+)
+@info @sprintf(
+    "Sponge: WRF damp_opt=2 analog, sin²() ramp from z = %.1f to %.1f km, rate %.3f s⁻¹ (ρu, ρv, ρw → 0; ρθ → ρᵣ·θᵣ)",
+    20.0, 25.0, sponge_rate
+)
```
glwagner: I think we probably should not call sprintf within examples (it is better style to rely on the notebook / block output that is intrinsic to Literate, probably)

examples/tropical_cyclone_with_rainband.jl
```diff
@@ -0,0 +1,1059 @@
...
+function rmw_analytic(z; ΔT_floor = ΔT_floor_default)
+    T_out = T_env(z_vortex_top)
+    ΔT_0 = T_env(0.0) - T_out
+    ΔT_z = max(T_env(z) - T_out, ΔT_floor)
+    return rmw_surface * sqrt(ΔT_0 / ΔT_z)
+end
+
+function tangential_wind(r, z; ΔT_floor = ΔT_floor_default)
+    r ≥ r_taper_end  && return 0.0
+    z ≥ z_vortex_top && return 0.0
```
glwagner: this hardcodes F64, which it probably shouldn't

examples/tropical_cyclone_with_rainband.jl
```diff
@@ -0,0 +1,1059 @@
...
+
+function dpdz_centered(p, k, i, z_centers, Nz)
+    if k == 1
+        return (p[i, 2] - p[i, 1]) / (z_centers[2] - z_centers[1])
+    elseif k == Nz
+        return (p[i, Nz] - p[i, Nz - 1]) / (z_centers[Nz] - z_centers[Nz - 1])
+    else
+        return (p[i, k + 1] - p[i, k - 1]) / (z_centers[k + 1] - z_centers[k - 1])
+    end
+end
```
glwagner: This can be implemented with operators, which is better. It should also act on fields.

examples/tropical_cyclone_with_rainband.jl
```diff
@@ -0,0 +1,1059 @@
...
+    history = Float64[]
+    for iter in 1:max_iter
+        T_prev = copy(T)
+        ρ = p ./ (Rᵈ .* T)
+        for k in 1:Nz
+            p[end, k] = p_col[k]
+            for i in (Nr - 1):-1:1
+                ρ_face = 0.5 * (ρ[i, k] + ρ[i + 1, k])
+                v_face = 0.5 * (v2d[i, k] + v2d[i + 1, k])
+                r_face = 0.5 * (r_grid[i] + r_grid[i + 1])
```
glwagner: More hardcoded operators

examples/tropical_cyclone_with_rainband.jl
```diff
@@ -0,0 +1,1059 @@
...
+    fr = (r_c - r0) / (r1 - r0)
+    fz = (z_c - z0) / (z1 - z0)
+    v00 = table[ir - 1, iz - 1]; v10 = table[ir, iz - 1]
+    v01 = table[ir - 1, iz];     v11 = table[ir, iz]
+    return (1 - fr) * (1 - fz) * v00 + fr * (1 - fz) * v10 +
+        (1 - fr) * fz * v01 + fr * fz * v11
+end
+
+uᵢ(x, y, z) = (r = sqrt(x^2 + y^2); r < 1.0 ? 0.0 : -(y / r) * lookup_rz(vortex.v, r, z))
+vᵢ(x, y, z) = (r = sqrt(x^2 + y^2); r < 1.0 ? 0.0 : (x / r) * lookup_rz(vortex.v, r, z))
```
glwagner: spurious f64

examples/tropical_cyclone_with_rainband.jl
```diff
@@ -0,0 +1,1059 @@
...
+    A_λ = exp(-(λ_c / π_4_F32)^8)
+
+    Q = p.Q_max * G_r * V_z * A_λ * ramp
+    ## ρθ tendency: ρᵣ · Q / Πᵣ. WRF uses full ρ; under the WRF/MN10 idealized
+    ## framework ρ ≈ ρᵣ(z) inside the rainband, so the difference is second
+    ## order in p'/p.
+    ρᵣ_k = @inbounds p.ρᵣ[k]
+    Πᵣ_k = @inbounds p.Πᵣ[k]
+    return ρᵣ_k * Q / Πᵣ_k
+end
```
glwagner, line 430: this can be implemented with the continuous form which might be clearer

examples/tropical_cyclone_with_rainband.jl
```diff
@@ -0,0 +1,1059 @@
...
+
+# ## Surface fluxes — omitted
+#
+# YD19 §3a1 uses Emanuel-1986 bulk aerodynamic drag (Cᴰ) and sensible heat
+# (Cᵀ) over a 300 K SST. We omit them here so the response field is the
+# direct dynamical response to the prescribed rainband heating.
+
+# ## Model builder
+
+function build_model(; with_heating::Bool)
```
glwagner: why do we need this?

examples/tropical_cyclone_with_rainband.jl
```diff
@@ -0,0 +1,1059 @@
...
+
+    @info "Running $stage_label stage for $(prettytime(stop_time))"
+    run!(simulation)
+
+    post = (
+        u = Array(interior(model.velocities.u)),
+        v = Array(interior(model.velocities.v)),
+        T = Array(interior(model.temperature)),
+        ρ = Array(interior(model.dynamics.density)),
+    )
```
glwagner: We should use output writers

examples/tropical_cyclone_with_rainband.jl
```diff
@@ -0,0 +1,1059 @@
...
+    end
+end
+function center_w!(out::AbstractArray{Float32, 3}, src)
+    ## w lives on z-Face (size Nz+1 in the vertical). Average adjacent
+    ## faces to get cell-centered values with size (Nx, Ny, Nz).
+    Nxs, Nys, Nzs = size(out)
+    return @inbounds for k in 1:Nzs, j in 1:Nys, i in 1:Nxs
+        out[i, j, k] = Float32((src[i, j, k] + src[i, j, k + 1]) / 2)
+    end
+end
```
glwagner: It isn't necessary to manually interpolate data --- Oceananigans built in post processing can handle this.

glwagner (commented):
@nrb171 made some comments -- with your blessing I will address them. It is possible that some of the cost is coming from the initialization and manual post processing. We can speed this up possibly by using built-in operations instead.

examples/tropical_cyclone_with_rainband.jl
glwagner: For example, if you look at the built example you will see this:

image

Therefore, it is not necessary to duplicate printing grid information. It is already there!

giordano:
> the docs build takes 86 min

Note that 86 minutes is the total, we are already at around 75 minutes on main. This new example alone [takes about 40 minutes](https://github.com/NumericalEarth/Breeze.jl/actions/runs/26830250995/job/79109190852#step:10:86), which increases the overall build time by 10 minutes (we do things concurrently, although load balancing is pretty bad).

glwagner:
> > the docs build takes 86 min
> 
> Note that 86 minutes is the total, we are already at around 75 minutes on main. This new example alone [takes about 40 minutes](https://github.com/NumericalEarth/Breeze.jl/actions/runs/26830250995/job/79109190852#step:10:86), which increases the overall build time by 10 minutes (we do things concurrently, although load balancing is pretty bad).

well, reducing the resolution seems to help a lot! we can possibly increase it to somewhere between 128 and 214^2. But check it out at 5km:

image

not that dissimilar to the 3km run.

test/diagnostics.jl
```diff
@@ -288,4 +288,22 @@ end
...
+    # center recovers a clean monotonic radial profile.
+    xc, yc = 0.3, -0.2
+    ρᵒ = CenterField(grid)
+    set!(ρᵒ, (x, y, z) -> sqrt((x - xc)^2 + (y - yc)^2))
+    ρ̄ᵒ = azimuthal_mean(ρᵒ; radius = 0.5, Nr = 8, center = (xc, yc))
+    offset_profile = Array(interior(ρ̄ᵒ, :, 1, 1))
+    @test issorted(offset_profile)
+    @test all(0 .< offset_profile .< 1)
+
+    # Rings finer than the grid leave some empty; those are filled with zero, not NaN,
```
nrb171: Feel free to push back on this, but I don't think empty rings should be filled with zeros. In cases where radial averages are also wanted after the azimuthal averages, the zeros would throw off the mean, and popping the empty rows would be an extra step in postprocessing. Oceananigans grids may have a way to handle empty voxels implicitly, but this could cause a problem.

src/AtmosphereModels/Diagnostics/azimuthal_mean.jl
```diff
@@ -0,0 +1,76 @@
...
+
+"""
+$(TYPEDSIGNATURES)
+
+Azimuthally average `field` into radial rings about `center = (xc, yc)`, returning a
+`Field` on a one-dimensional-in-radius grid — `Bounded` in radius, `Flat` in azimuth, and
+`Bounded` in z — with `Nr` uniform rings spanning ``[0, \\texttt{radius}]`` and the same
+vertical grid as `field`.
+
+This bins the Cartesian `field` by radius, a pragmatic stand-in for a reduction on a true
```
nrb171: We may want to put an upper limit on `Nr`.

Related to my previous comment, this is where the issue with very fine grids is coming from, and ultimately why I'm using interpolation now . As `Nr` increases, the accuracy of the azimuthal mean decreases since the number of points in the radial band decreases. Near the limit, there are lots of empty rings.

glwagner: can you make a specific suggestion?

test/diagnostics.jl
glwagner: what change do you suggest? (this comment is on a test, but I think we need a change somewhere in the source code)

nrb171: Yeah, saw the test and wanted to flag this behavior. Thanks for fixing it.

src/AtmosphereModels/Diagnostics/azimuthal_mean.jl
nrb171: Looks like you addressed it in [48c23d5](https://github.com/NumericalEarth/Breeze.jl/pull/657/commits/48c23d5bf495c6c68fe5ab7c1549d15a6c950b92). Thanks!

glwagner: Took this a step further in b5d8513: rather than just NaN-filling, the binning itself is now conservative. Each Cartesian cell is split into an `m × m` block of sub-cells (default `m = 4`) binned by radius, so a cell contributes — area-weighted — to *every* ring it overlaps instead of landing wholesale in the one ring under its center.

That's a first-order conservative remap onto the rings, which is exactly the accuracy concern you raised: a cell is no longer "starved" at fine `Nr`, and near-center rings stay populated. Empty rings effectively don't arise for sensible `Nr` (`radius/Nr ≳` grid spacing), so no `Nr` cap is needed; the rare ones that still catch nothing (only when `radius/Nr` is finer than a sub-cell) stay `NaN`.

It stays on the GPU and needs no dependency — because the targets are full annuli, the overlap is a 1-D radial calculation, so we don't need `ConservativeRegridding.jl`'s polygon machinery. Validated: example figure unchanged (<1% from the center-binned version), simulation bit-identical, and CPU tests cover the conservative fill / monotonic / off-center cases.

PR #658, opened 2026-04-25
Revert "Register atan, atand, mod as binary operations on Fields (#644)"

giordano:
Code now [lives upstream in Oceananigans](https://github.com/CliMA/Oceananigans.jl/pull/5538).


giordano:
I'm going to merge this, because precompilation is currently broken without this.

PR #660, opened 2026-04-27
[benchmarking] Support AMD GPUs

giordano:
Simple adaptation to support AMD GPUs when running the benchmarks.  Only downside is that it requires adding the `AMDGPU` package explicitly to the benchmarking environment.

### DIFF
--- a/benchmarking/src/metadata.jl
+++ b/benchmarking/src/metadata.jl
@@ -19,7 +19,7 @@ function BenchmarkMetadata(arch)
     gpu_name = nothing
     cuda_version = nothing
 
-    if arch isa Oceananigans.Architectures.GPU{CUDABackend}
+    if arch isa GPU{CUDABackend}
         try
             gpu_name = CUDA.name(CUDA.device())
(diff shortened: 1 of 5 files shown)


benchmarking/src/metadata.jl
```diff
@@ -19,14 +19,21 @@ function BenchmarkMetadata(arch)
...
     cuda_version = nothing
 
-    if arch isa Oceananigans.Architectures.GPU{CUDABackend}
+    if arch isa GPU{CUDABackend}
         try
             gpu_name = CUDA.name(CUDA.device())
             cuda_version = string(CUDA.runtime_version())
         catch
             gpu_name = "Unknown GPU"
             cuda_version = "Unknown"
```
giordano, line 28: `cuda_version` should be changed to be more generic, but I don't know how to get the AMD GPU runtime version, so I'm not touching this at the moment.

glwagner (approved):
adding the package is worth it!

navidcy (approved):

PR #661, opened 2026-04-27
Add `eltype` and `architecture` methods for `AtmosphereModel`

giordano:
Without these methods, we were hitting very generic fallback methods in Oceananigans which always return `Float64` for the eltype and `nothing` for the architecture.  This was particularly bad on Metal GPUs because the `eltype` of the model was always `Float64` despite the model being `Float32`

### DIFF
--- a/test/atmosphere_model_construction.jl
+++ b/test/atmosphere_model_construction.jl
@@ -2,6 +2,7 @@ using Breeze
 using Breeze.Thermodynamics: TetensFormula
 using GPUArraysCore: @allowscalar
 using Oceananigans
+using Oceananigans.Architectures: architecture
 using Oceananigans.Diagnostics: erroring_NaNChecker!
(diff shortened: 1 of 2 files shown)


test/atmosphere_model_construction.jl
```diff
@@ -23,11 +24,13 @@ function run_nan_checker_test(arch; erroring)
     return nothing
 end
 
-@testset "AtmosphereModel [$(FT)]" for FT in test_float_types()
+@testset "AtmosphereModel [$(FT)]" for FT in all_float_types()
```
giordano, line 27: This adds some overhead when running the tests (on my macbook 60 -> 90 seconds) but I think it's worth it to make sure the very basic stuff of constructing a model works with Float32 correctly.

glwagner (approved):

navidcy (approved):

PR #662, opened 2026-04-29
Fix summary and docs for FilteredBCs

navidcy:


ext/BreezeCloudMicrophysicsExt/two_moment_microphysics.jl
giordano, line 1: 🥲

navidcy: Damn...

mmr0 (approved):
Thanks, that's clearer!

I ran the example and enjoyed the `...), with filtering` addition

PR #664, opened 2026-04-30
Microphysics scheduler

kaiyuan-cheng:
## Motivation

  Microphysics can be expensive, yet the right *coupling frequency* with the dynamics is still an open question (substep vs. superstep, and at what cadence). This PR exposes that knob via a `microphysics_schedule` keyword. The default (`nothing`) is a
  no-op, byte-identical to before; setting a schedule lets you super-step microphysics and study the trade-off without touching the dycore call site.

## Summary

  Adds opt-in scheduled microphysics to `AtmosphereModel` via a new `microphysics_schedule` keyword (e.g., `IterationInterval(N)`, `TimeInterval(Δt)`).

  When set, microphysics tendencies are cached to per-prognostic `CenterField`s and refilled only when the schedule fires. In between firings, the dycore reads the held cache, allowing the dynamics to super-step microphysics. The operator-split entry point `microphysics_model_update!` is plumbed with an explicit `Δt_eff = clock.time - last_fire_time`, so schemes (e.g. `DCMIP2016KesslerMicrophysics`) integrate over the actual elapsed window rather than a single dycore step.

  With `microphysics_schedule = nothing` (the default), the path is bit-identical to the previous code: cache fields are `nothing`, dispatch resolves to the existing inline `grid_microphysical_tendency`, and the operator-split call reduces to `microphysics_model_update!(μ, model, clock.last_Δt)`.

  ## Changes

  - New `microphysics_schedule` keyword on `AtmosphereModel` + cached tendency fields + `MicrophysicsScheduleState`
  - `microphysics_model_update!` refactored to a 3-arg form `(microphysics, model, Δt_eff)` with a backward-compatible 2-arg shim; `DCMIP2016KM` (grid + parcel) now reads `Δt_eff` instead of the clock
  - Cache-aware `grid_microphysical_tendency` overload (compile-time `haskey` via `Val{N}` dispatch)
  - New `compute_microphysics_tendencies!` kernel that builds `𝒰` and `ℳ` once per grid point and writes the tendency for every cached name via static `Val` iteration
  - New `update_microphysics!(model)` driver replaces the bare `microphysics_model_update!` call in `update_state!`
  - Three dycore tendency kernels (`scalar_tendency`, `static_energy_tendency`, `potential_temperature_tendency`) thread the cache through `common_args`
  - `Base.show(io, ::AtmosphereModel)` displays the schedule when set; constructor docstring updated
  - New example `examples/splitting_supercell_scheduled_microphysics.jl` (DCMIP2016 Kessler + `TimeInterval(20)` at fixed `Δt = 4` s → fires every 5 dycore steps)

### DIFF
--- a/examples/splitting_supercell_scheduled_microphysics.jl
+++ b/examples/splitting_supercell_scheduled_microphysics.jl
@@ -1,0 +1,264 @@
+# # Splitting supercell with scheduled microphysics
+#
+# This example mirrors the [splitting supercell](@ref) case but exercises Breeze's
+# `microphysics_schedule` keyword on [`AtmosphereModel`](@ref). Microphysics fires
+# every 20 simulated seconds (= 5 dycore steps at the fixed ``Δt = 4`` s) rather
+# than every step. The operator-split state update and the cached tendency fields
+# are both gated by the same schedule, which lets the dynamics super-step a
+# potentially expensive microphysics package without breaking thermodynamic
+# consistency.
+#
+# On firing, the schedule passes ``Δt_{\rm eff} = t - t_{\rm last\,fire}`` to
+# [`microphysics_model_update!`](@ref Breeze.AtmosphereModels.microphysics_model_update!),
+# so the operator-split Kessler scheme integrates over the actual elapsed window
+# rather than a single dycore step. In between firings, the dycore tendency kernels
+# read precomputed `CenterField` tendencies that were filled at the last firing.
+#
+# !!! warning
+#     Holding microphysical rates constant across 20 s windows is physically
+#     reasonable only for the slowly-varying part of the scheme. For Kessler-style
+#     sedimentation and rain evaporation in active convection, results may differ
+#     from the every-step reference. Use this example to evaluate the trade-off,
+#     not as a recommended production setting for deep convection.
+#
+# Other than the microphysics schedule, the physical setup, initial conditions,
+# and diagnostics are identical to the unscheduled `splitting_supercell` example.
+
+using Breeze
+using Breeze: DCMIP2016KesslerMicrophysics, TetensFormula
+using Breeze.Thermodynamics: hydrostatic_density, hydrostatic_temperature
+using Oceananigans: Oceananigans
+using Oceananigans.Units
+using Oceananigans.Grids: znodes
+
+using CairoMakie
+using CUDA
+using Printf
+
+# ## Domain and grid
+
+Oceananigans.defaults.FloatType = Float32
+
+Nx, Ny, Nz = 168, 168, 40
+Lx, Ly, Lz = 168kilometers, 168kilometers, 20kilometers
+
+grid = RectilinearGrid(GPU(),
+                       size = (Nx, Ny, Nz),
+                       x = (0, Lx),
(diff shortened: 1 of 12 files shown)


glwagner:
Microphysics isn't expensive for us (its a tiny fraction of the cost of WENO advection). I propose not adding something like this until we need it. I suspect we won't ever need it --- the main issue for microphysics is that it adds tracers, which need to be advected. It would be great if we can avoid these kinds of hacks too!

glwagner:
I do think it will be interesting to investigate evaluating microphysics on the acoustic substep though. But we need to wait for #622 for that.

I think it may be reasonable to evaluate microphysics on the acoustic substep if/when it is relatively cheap, which seems to be the case for most of the schemes right now.

kaiyuan-cheng:
> Microphysics isn't expensive for us (its a tiny fraction of the cost of WENO advection). I propose not adding something like this until we need it. I suspect we won't ever need it --- the main issue for microphysics is that it adds tracers, which need to be advected. It would be great if we can avoid these kinds of hacks too!

I understand the concern. While this approach may appear to be a step backward, I argue that it is actually a necessary functionality. One motivation is to align our splitting supercell simulation with the literature. Our supercell splits and becomes unorganized more quickly than reported by  [Zarzycki et al. (2019)](https://gmd.copernicus.org/articles/12/879/2019/gmd-12-879-2019.html). Furthermore, the supercell develops too vigorously. I suspect this occurs because we update the microphysics at every timestep, which may not be suitable for moment schemes that are all designed for larger spatial and temporal scales. Indeed, updating the Kessler microphysics at a lower frequency yields much better results.

Kessler called every 20 seconds

image

vs. every 4 seconds

image

glwagner:
Are you sure that the Kessler parameters shouldn't just be re-tuned? I don't think that it's possible that calling the scheme less frequently can make it more accurate. I think its more likely that you have simply revealed a deficiency in the parameterization.

As a thought experiment, wouldn't you simply recover the same problem if you used a smaller time-step? (This could be a further experiment you might try). It's very important that all numerical schemes are "consistent", in the sense that they converge as the time-step or grid spacing is reduced. When a model does not have this property, it is very hard to make sense of its results.

glwagner:
@kaiyuan-cheng one major difference about our case is that we don't have any explicit diffusion, whereas the test case suggests a Laplacian diffusion:

image

I suspect diffusion will tend to spread out the convective cores and make them weaker. Note the variety of schemes that many models used (in addition to the recommended one, which is a second order diffusion applied to momentum and tracers separately)

glwagner:
This is interesting:

image

we should perform a resolution study as well (perhaps the example can add one?)

here's the same for us, but the example only has one resolution right now:

image

I think it is a point of pride to have stronger vertical velocities! This is what we want --- note that vertical velocity increases with finer resolution. Having stronger vertical velocities is a sign of higher effective resolution. However, we should also add the Laplacian diffusion specified by the test, because that may reduce our vertical velocity somewhat.

glwagner:
Seems like the closure is WAY overkill. I don't understand the other model results?

image

(there could also be a problem with how closures are implemented in Breeze)

PR #674, opened 2026-05-06
Add regression test for virtual potential temperature formulation

### LINKED ISSUE #659: Add regression tests for formulation of virtual potential temperature bugs
> Good point @giordano, we don't have a test for this. We could have a simple one that builds a state in `AtmosphereModel` and then evaluates that $\theta_v$ takes the expected value(s)

_Originally posted by @glwagner in https://github.com/NumericalEarth/Breeze.jl/issues/656#issuecomment-4319592841_

kaiyuan-cheng:
## Summary

- Adds a regression test that pins down the dry-air Poisson exponent `Rᵈ/cᵖᵈ` used by the buoyancy-side `virtual_potential_temperature` kernel (fixed in #656), guarding against a revert to the mixture exponent `Rᵐ/cᵖᵐ`.
- The test builds an `AtmosphereModel` with a 5 km column so reference pressure at the top differs noticeably from `pˢᵗ`, sets `θ=300`, `qᵗ=0.01`, and at every level checks the kernel against the analytical formula `T·(Rᵐ/Rᵈ)·(pˢᵗ/pᵣ)^(Rᵈ/cᵖᵈ)` with `rtol = 100 eps(FT)`.

Closes #659.

## Test plan

- [x] `Pkg.test("Breeze"; test_args=\`diagnostics\`)` passes (57/57) on CPU
- [x] `Pkg.test("Breeze"; test_args=\`quality_assurance\`)` passes (17/17), confirming explicit-imports are correct
- [ ] Full CI passes

## Note for reviewers

While writing this test I noticed that the public `VirtualPotentialTemperature(model)` diagnostic in `src/AtmosphereModels/Diagnostics/potential_temperatures.jl` still computes `θ·(1 + δqᵛ − qˡ − qⁱ)` with `θ` based on the moist Exner exponent `Rᵐ/cᵖᵐ`, while the buoyancy kernel now uses the dry exponent. Worth a follow-up to decide whether the diagnostic should match the buoyancy definition.


giordano (approved):
I can confirm the new tests fail on 4c9cfb299748d6accb098b51b89af92cc7893541, the parent of the merge commit of #656 🥳 Thanks for adding the tests!

giordano:
> ## Note for reviewers
> 
> While writing this test I noticed that the public `VirtualPotentialTemperature(model)` diagnostic in `src/AtmosphereModels/Diagnostics/potential_temperatures.jl` still computes `θ·(1 + δqᵛ − qˡ − qⁱ)` with `θ` based on the moist Exner exponent `Rᵐ/cᵖᵐ`, while the buoyancy kernel now uses the dry exponent. Worth a follow-up to decide whether the diagnostic should match the buoyancy definition.

@glwagner I guess this is comment for you, but this looks good to me otherwise.

glwagner:
> while the buoyancy kernel now uses the dry exponent. Worth a follow-up to decide whether the diagnostic should match the buoyancy definition.

There's no question the diagnostic and internal utility should be identical!

kaiyuan-cheng:
> > ## Note for reviewers
> > While writing this test I noticed that the public `VirtualPotentialTemperature(model)` diagnostic in `src/AtmosphereModels/Diagnostics/potential_temperatures.jl` still computes `θ·(1 + δqᵛ − qˡ − qⁱ)` with `θ` based on the moist Exner exponent `Rᵐ/cᵖᵐ`, while the buoyancy kernel now uses the dry exponent. Worth a follow-up to decide whether the diagnostic should match the buoyancy definition.
> 
> @glwagner I guess this is comment for you, but this looks good to me otherwise.

Good catch! Fixed.

PR #675, opened 2026-05-06
Pin Reactant and GPUCompiler to known compatible versions

giordano:
Same as https://github.com/CliMA/Oceananigans.jl/pull/5566


wsmoses:
the issue isn't the reactant side, its the gpucompiler side [aka newer reactants with older gpucompiler work].

in any case proper release has been made.

giordano:
Oceananigans tests are still broken with new Reactant (although that may not be related to GPUCompiler anymore?): https://github.com/CliMA/Oceananigans.jl/actions/runs/25472371819/job/74738686218?pr=5567#step:8:3130

PR #676, opened 2026-05-06
Fuse multi-output microphysical tendencies for GPU performance

kaiyuan-cheng:
## Summary
- Adds a `compute_microphysical_tendencies!(model)` interface (no-op by default) so schemes whose tendency computation bundles many process rates feeding multiple prognostics can fuse the work into a single launch instead of recomputing the bundle once per tracer.
- Implements the fused path for **MPNE1M** (mixed-phase non-equilibrium 1M) and **WPNE2M** (warm-phase non-equilibrium 2M). Each scheme overrides `grid_microphysical_tendency` to zero and writes all bundled outputs to `Gⁿ` from one kernel.
- Switches `examples/splitting_supercell.jl` to WPNE2M with a 5-panel animation (w, qᶜˡ, nᶜˡ, qʳ, nʳ) to exercise the new path.

## Why
The per-tracer dispatchers like `microphysical_tendency(::MPNE1M, ::Val{:ρqˢ}, …)` each call `mpne1m_tendencies(…)` and discard 4 of 5 outputs. With one tendency-kernel launch per prognostic field and AcousticSSPRungeKutta3 having 3 stages, that's 15 full bundle evaluations per cell per timestep for MPNE1M (18 for WPNE2M). The compiler's DCE prunes some of it, but the conservation-routed numerical guards inside the bundle make most rates live for any single field extraction, so most of that work is genuinely redundant.

## Measured speedup (H100, Float32, 168×168×40, AcousticSSPRungeKutta3)

| Scheme | Unfused ms/iter | Fused ms/iter | Speedup |
|---|---|---|---|
| MPNE1M | 11.04 | 8.75 | **1.26×** |
| WPNE2M | 15.28 | 10.08 | **1.52×** |

The 2M case sees a larger speedup because the bundle has one extra prognostic, the coupled α-limiters cross-couple all process rates (so DCE has very little to remove), and aerosol activation (Abdul-Razzak–Ghan) contains expensive transcendentals (`erf`, `log`).

## Test plan
- [x] `cloud_microphysics_1M` test set passes (81 tests, including MPNE1M time-stepping)
- [x] `cloud_microphysics_1M_warm_ice_deposition` test set passes
- [x] Smoke test on CPU: 2-step `time_step!` with WPNE2M succeeds end-to-end
- [x] `examples/splitting_supercell.jl` runs to completion on H100 with WPNE2M
- [ ] Reviewer: confirm no regression in `cloud_microphysics_2M` test set
- [ ] Reviewer: profile or `@time_ns` to confirm fused kernel is one of the cheaper kernels (sanity check that we didn't blow up register usage)

## Notes for reviewers
- The fused 2M kernel uses `grid_microphysical_state` (which interpolates u/v/w to centers) because WPNE2M's `microphysical_state` stores velocities (aerosol activation depends on vertical velocity). The 1M kernel reads the prognostic fields directly since MPNE1M's `microphysical_state` doesn't use velocities.
- The per-tracer `microphysical_tendency(::MPNE1M, ::Val{:ρq?}, …)` and `microphysical_tendency(::WPNE2M, ::Val{:ρ??}, …)` methods are **kept** — they are still used by parcel models and the unit tests, which call them directly with a constructed state. The grid path now bypasses them via the `grid_microphysical_tendency` override.
- Float32 accumulation order changes between the two paths (per-tracer kernel does `Gρq = adv + diff + 𝓜 + forcing` in one expression; fused path does `Gρq = adv + diff + forcing` then `Gρq += 𝓜`), so trajectories diverge at ULP level and amplify chaotically over thousands of timesteps. Bulk diagnostics still agree — see Float64 tests for bit-level correctness.


src/AtmosphereModels/microphysics_interface.jl
```diff
@@ -242,6 +242,32 @@ end
 # Explicit Nothing fallback (for backward compatibility)
 @inline grid_microphysical_tendency(i, j, k, grid, microphysics::Nothing, name, ρ, μ, 𝒰, constants, velocities) = zero(grid)
 
+#####
+##### Fused multi-output microphysical tendency interface
+#####
+
+"""
+    compute_microphysical_tendencies!(model)
```
giordano:
```suggestion
$(TYPEDSIGNATURES)
```

ext/BreezeCloudMicrophysicsExt/one_moment_microphysics.jl
```diff
@@ -930,3 +930,60 @@ end
...
+#
+# `mpne1m_tendencies` bundles all 5 prognostic tendencies in one call. To avoid
+# recomputing process rates 5 times per cell, we (1) zero out the per-tracer term
+# and (2) apply all tendencies in a single fused kernel below.
+
+# Skip the per-tracer microphysics term: handled by the fused kernel below.
+@inline AM.grid_microphysical_tendency(i, j, k, grid, microphysics::MPNE1M,
+                                       name, ρ, fields, 𝒰, constants, velocities) = zero(grid)
+
+@kernel function _compute_mpne1m_tendencies!(Gρqᵛ, Gρqᶜˡ, Gρqᶜⁱ, Gρqʳ, Gρqˢ,
```
glwagner, line 942: Is there any possibility to re-use the `grid_microphysical_tendency` interface? This design basically negates that interface.

Another possibility is to excise the `grid_microphysical_tendency` interface in favor of this new interface. The downside is that we permanently bake-in splitting of the microphysics tendency from advection tendency. However, if that is faster in most situations, I think it's a good approach.

I think more generally it might be nice to provide just "one way" of extending microphysics, for simplicity.

kaiyuan-cheng:
Pushed a follow-up commit that addresses the reviewer's question about the dual interface (5e0ff88a).

Replaced `grid_microphysical_tendency` + `compute_microphysical_tendencies!` with a single entry point. The atmosphere model now invokes microphysics solely through `compute_microphysical_tendencies!`, called after the per-tracer dynamics kernels write advection + diffusion + forcing to `Gⁿ`.

Schemes plug in by extending one of two methods:

- **`microphysical_tendency`** (state-based, per-name): the default `compute_microphysical_tendencies!` launches a single fused kernel that builds `ℳ` and `𝒰` once per cell and `+=`s the result into each prognostic `G` field. Simple schemes (Kessler, saturation adjustment, ZMCM) plug in here.
- **`compute_microphysical_tendencies!(microphysics, model)`** directly: bundle schemes (MPNE1M, WPNE2M) override this with their fused-bundle kernel.

The MPNE1M/WPNE2M zero-overrides on `grid_microphysical_tendency` (and the `grid_microphysical_tendency` calls inside `scalar_tendency`, `potential_temperature_tendency`, `static_energy_tendency`) are gone. Verified locally:

- `cloud_microphysics_0M` + `cloud_microphysics_1M`: 95/95 pass
- `cloud_microphysics_2M`: 71/71 pass
- ExplicitImports: clean

ext/BreezeCloudMicrophysicsExt/one_moment_microphysics.jl
```diff
@@ -930,3 +930,56 @@ end
...
+
+    # Build the microphysical state directly to avoid the velocity interpolation
+    # in `grid_microphysical_state` (MPNE1M's `microphysical_state` does not use it).
+    @inbounds qᶜˡ = microphysical_fields.ρqᶜˡ[i, j, k] / ρ
+    @inbounds qᶜⁱ = microphysical_fields.ρqᶜⁱ[i, j, k] / ρ
+    @inbounds qʳ  = microphysical_fields.ρqʳ[i, j, k]  / ρ
+    @inbounds qˢ  = microphysical_fields.ρqˢ[i, j, k]  / ρ
+    ℳ = MixedPhaseOneMomentState(qᶜˡ, qᶜⁱ, qʳ, qˢ)
+
+    G = mpne1m_tendencies(microphysics, ρ, ℳ, 𝒰, constants)
```
glwagner, line 964: Should we remove this interface and put the body of this function straight into the kernel?

ext/BreezeCloudMicrophysicsExt/two_moment_microphysics.jl
```diff
@@ -914,3 +914,53 @@ end
...
+
+    # Reconstruct moisture fractions and thermodynamic state at this cell.
+    q = AtmosphereModels.grid_moisture_fractions(i, j, k, grid, microphysics, ρ, qᵛ, microphysical_fields)
+    𝒰 = AtmosphereModels.diagnose_thermodynamic_state(i, j, k, grid, formulation, dynamics, q)
+
+    # WPNE2M's microphysical_state stores velocities (aerosol activation depends on w),
+    # so we need the full grid_microphysical_state which interpolates u, v, w to centers.
+    ℳ = AtmosphereModels.grid_microphysical_state(i, j, k, grid, microphysics, microphysical_fields, ρ, 𝒰, velocities)
+
+    G = wpne2m_tendencies(microphysics, ρ, ℳ, 𝒰, constants)
```
glwagner, line 944: ditto here --- the new interface for building microphysics schemes requires defining this kernel, I think...

glwagner (commented):
Do we have to update the microphysics interface docs to reflect the fact that now, after this PR, building a microphysics scheme cannot be done merely by defining a set of functions? This PR changes the interface so that developers have to write a kernel. The documentation should also explain how to write kernels with links to KernelAbstractions and `launch!` in Oceananigans I think.

giordano:
Is this good to go once CI is happy?  Performance improvement for `CBL; Dynamics: compressible_explicit; Microphysics: 1M_MixedNonEquilibrium [Float64]/Compare backends/NVIDIA L4/vanilla 256x256x128` is pretty solid, over 70%, which also makes vanilla Breeze a lot faster than the Reactant backend 😛

docs/src/developer/microphysics/fused_example.md
```diff
@@ -0,0 +1,243 @@
...
+
+To make the comparison concrete, we re-implement the same `ExplicitMicrophysics` scheme from
+the [per-name walkthrough](example.md) via the override path. The physics is identical;
+only the extension point differs.
+
+The shared portions of the implementation — the struct, prognostic field names, field
+materialization, the state type, and `microphysical_state` — are unchanged from the per-name
+walkthrough. We list them here so the page is self-contained:
+
+```@example fused_microphysics_example
```
giordano: I don't think these should be `@example`s, no code is actually executed, all blocks are only defining methods
````suggestion
```julia
````

File: docs/src/developer/microphysics/fused_example.md [hunk 6]
```diff
@@ -0,0 +1,243 @@
...
+own `microphysical_tendency` method. Notice that the liquid and ice tendencies each recompute
+`temperature(𝒰, constants)` independently, and the vapor tendency calls back into the
+liquid and ice methods — so a per-name kernel dispatch evaluates `temperature` three times
+per cell. For this small scheme the overhead is modest, but the structure scales poorly:
+each new prognostic adds another redundant pass through the same intermediates.
+
+The override path begins by defining a single helper that returns all three tendencies as
+a NamedTuple, with the shared intermediates computed exactly once:
+
+```@example fused_microphysics_example
```
giordano, line 78:
````suggestion
```julia
````

File: docs/src/developer/microphysics/fused_example.md [hunk 7]
```diff
@@ -0,0 +1,243 @@
...
+    Sᵛ = -Sˡ - Sⁱ
+
+    return (ρqᵛ = Sᵛ, ρqˡ = Sˡ, ρqⁱ = Sⁱ)
+end
+```
+
+The helper is gridless: it works equally well in a parcel model context. If a parcel model
+needs the per-name interface, the helper can also be wrapped:
+
+```@example fused_microphysics_example
```
giordano, line 101:
````suggestion
```julia
````

File: docs/src/developer/microphysics/fused_example.md [hunk 8]
```diff
@@ -0,0 +1,243 @@
...
+state-based unit tests.
+
+## Write the fused kernel
+
+The fused kernel is responsible for everything the default kernel does — reading density,
+extracting the prognostic moisture, reconstructing the thermodynamic state ``𝒰``, building
+the microphysical state ``ℳ`` — and then for `+=`ing the bundled tendencies into all
+`Gⁿ` fields:
+
+```@example fused_microphysics_example
```
giordano, line 124:
````suggestion
```julia
````

File: docs/src/developer/microphysics/fused_example.md [hunk 9]
```diff
@@ -0,0 +1,243 @@
...
+2. We *do not* call `grid_microphysical_state` because we want to skip its velocity
+   interpolation: this scheme's `microphysical_state` ignores velocities. Schemes that
+   need velocity-dependent activation should call `grid_microphysical_state` instead,
+   passing `transport_velocities(model)` from the override below.
+
+## Override `compute_microphysical_tendencies!`
+
+The override threads the model fields the kernel needs and launches it:
+
+```@example fused_microphysics_example
```
giordano, line 160:
````suggestion
```julia
````

File: docs/src/developer/microphysics/fused_example.md [hunk 10]
```diff
@@ -0,0 +1,243 @@
...
+specific dispatch on `::ExplicitMicrophysics` takes priority whenever the model carries this
+scheme.
+
+## Auxiliary, moisture, and saturation hooks (unchanged)
+
+The remaining interface — `update_microphysical_auxiliaries!`, `moisture_fractions`, and
+`maybe_adjust_thermodynamic_state` — is independent of how tendencies are dispatched, and
+the implementations from the [per-name walkthrough](example.md) carry over verbatim:
+
+```@example fused_microphysics_example
```
giordano, line 193:
````suggestion
```julia
````

**Review by giordano — COMMENTED** [review]
Docs build is failing because the `@example fused_microphysics_example` blocks require `KernelAbstractions` which isn't available in the `docs/` environment.  While we could easily add the package to the environment, I don't think all these blocks should be `@example`s to start with, they don't run any code anyway.

File: docs/src/developer/microphysics/fused_example.md [hunk 11]
```diff
@@ -205,7 +205,7 @@ The remaining interface — `update_microphysical_auxiliaries!`, `moisture_fract
 `maybe_adjust_thermodynamic_state` — is independent of how tendencies are dispatched, and
 the implementations from the [per-name walkthrough](example.md) carry over verbatim:
 
-```@example fused_microphysics_example
+```julia
```
glwagner: this has to be an `@example` otherwise the output won't be shown

giordano: What output?  Nothing is run, these are all method definitions.

kaiyuan-cheng: The rendering of the new doc looks good to me.

glwagner: Oh! The code in the doc should be run. Otherwise it may not run if a user copy pastes it. Can we add a block after these method definitions that exercises the code?

giordano:
@glwagner https://numericalearth.github.io/BreezeDocumentation/previews/PR676/developer/microphysics/fused_example/#Exercising-the-Scheme

glwagner:
beautiful! 

one final suggestion is to change "Example implementation" to "Per-name implementation" (distinguished from "Fused-kernel implementation"). 

image

glwagner:
Also "White path sHould I pick" probably belongs in the overview section, not buried in the Fused-kernel implementation

giordano:
Since all comments above have been addressed, I'm going to merge this PR 🚀

PR #677, opened 2026-05-06
Adds benchmarking for Reactant vs. Vanilla `AtmosphereModel`

dkytezab:
Timings for various grid sizes for Compressible Model with one moment mixed-phase microphysics


.github/workflows/Benchmarks.yml
giordano, line 1: Need also to adapt https://github.com/NumericalEarth/Breeze.jl/blob/29b798cf066d4bbf555350df33b11d6f91338106/.github/workflows/Benchmarks.yml#L112-L159 The number of grouping levels is fixed, can't change that, you can see in https://numericalearth.github.io/BreezeBenchmarks/ how they're used.

.github/workflows/Benchmarks.yml
```diff
@@ -40,14 +40,36 @@ jobs:
...
+          # field-dependent surface drag BCs (which do not currently
+          # materialize on ReactantState). Runs in Float64 because Reactant's
+          # @trace loop trips on the Float32-grid / Float64-clock mismatch.
+          - dynamics: 'compressible_explicit'
+            microphysics: '1M_MixedNonEquilibrium'
+            grid: '128x128x64, 256x256x128, 384x384x192'
+            advection: 'WENO5'
+            backend: 'vanilla, reactant'
+            topology: 'PBB'
+            extra_flags: '--simplified --float_type=Float64'
```
giordano, line 72: I'd prefer for the mismatch to be fixed (which would be good performance-wise anyway).  Float64 is a massive performance hit on this GPU.

dkytezab: I agree, I plan on opening a separate PR to apply our fixes from GB for this

giordano: Not entirely related to this benchmark because that's about the compressible dynamics, but in #680 I had to track down a few accidental `Float64`s (mostly coming from `Δt`), that's probably relevant elsewhere though.  We need a better solution that what I currently have there though (that was mainly for testing).

dkytezab: I think we had the same problem in GB. If I recall the hacky solution was to add FP32 0 to `Δt` in the first time step or something like that. Why not build the model with `FT(Δt)`?

giordano: > Why not build the model with `FT(Δt)`?

Yeah, I was discussing this option with @glwagner earlier.  There are a couple of things to consider:

* is that a problem for Reactant?  In the sense that do we need to allow it to be traced or not?
* `Δt` may also be a `Dates.Period`, we need to convert it to seconds beforehand

giordano: Well, `Δt` is part of `Simulation`, not the model, but `Simulation` is defined in Oceananigans so I'd say it's an upstream issue.

giordano: https://github.com/NumericalEarth/Breeze.jl/blob/764fe0a800bd74262a6bbef267e1acca2e249b0f/benchmarking/src/utils.jl#L286 uhm, this looks bad :disappointed:

dkytezab: In our GB runs `Δt` became a `ConcreteRNumber` – I imagine that baking in `Δt` as a constant should also be fine

dkytezab: Could we have Oceananigans `Defaults` propagate through `Clock`? 
https://github.com/CliMA/Oceananigans.jl/blob/05c263e5950c544741d01d1add60b9f5d20dc17e/src/TimeSteppers/clock.jl#L104

giordano:
There are a couple of conflicts with #680 now :innocent:

.github/workflows/Benchmarks.yml
giordano: Rather than using oceananigans default (that'd be type-unstable, depending on a non-const global), that could more simply use the eltype of the grid, which is always well defined: https://github.com/CliMA/Oceananigans.jl/blob/638a259e2e362139473a436a30c6241402830815/src/Grids/abstract_grid.jl#L78

benchmarking/src/metadata.jl
```diff
@@ -34,6 +34,15 @@ function BenchmarkMetadata(arch)
             gpu_name = "Unknown GPU"
         end
         cuda_version = "unknown"
+    elseif arch isa ReactantState && CUDA.functional()
```
giordano: The body of this branch is the same as the CUDA one, right? If so, you can simply extend the condition above to include this one, instead of duplicating the code

giordano:
> # :warning: **Performance Alert** :warning:
> 
> Possible performance regression was detected for benchmark **'Breeze.jl Benchmarks'**.
> Benchmark result of this commit is worse than the previous benchmark result exceeding threshold `1.10`.
> 
> | Benchmark suite | Current: 8a7263326a5ac554d83376e26c53993abc2e8eb9 | Previous: d978c46862d40a209283993a3b8576607944749d | Ratio |
> |-|-|-|-|
> | `CBL; Dynamics: compressible_splitexplicit; Microphysics: nothing [Float32]/Advection: WENO5/NVIDIA L4/512x512x256` | `14893673.49162833` points/s | `19844063.114872545` points/s | `1.33` |
> 
> This comment was automatically generated by [workflow](https://github.com/NumericalEarth/Breeze.jl/actions?query=workflow%3ABenchmarks) using [github-action-benchmark](https://github.com/marketplace/actions/continuous-benchmark).

I haven't investigated, but it's very likely some `Float64`s are back in the compressible benchmarks, because ~1.5e7 gridpoints/s was precisely what we had before #680.

dkytezab:
Ideally next we should add benchmarking of Reactant with FP32 against 2 FP16 (leveraging the multifloat work inside of Enzyme-JAX and GB). However there are some issues with raising multifloat that I need to investigate further. Also, multifloat seems presently only supported for 2 'limbs', i.e. we can't yet do FP32 represented as 4 FP8. 

I'm not completely certain that multifloat will give us a significant speed up considering that we're usually bounded my memory rather than arithmetic intensity. But worth looking into still

giordano:
Preview: https://numericalearth.github.io/BreezeBenchmarks/previews/PR677/.

Is there anything else we need to do? Ill definitely need to revert the preview commit. Multifloat sounds like a potential follow up, not a blocker for this PR. I still think using Float32 here would be nicer (for consistency with the other benchmarks and runtime performance, although the Reactant benchmarks are probably bottlenecked by compilation...), but if that's too complicated for now we can postpone it for later.

dkytezab:
Preview looks good to me. I think we should add a follow up to restore both Reactant benchmarks to FP32 once we resolve https://github.com/CliMA/Oceananigans.jl/pull/5570

PR #678, opened 2026-05-07
Allow passing the float type as argument to `SplitExplicitTimeDiscretization`

giordano:
This makes it easier to construct the object with a different float type.

### DIFF
--- a/src/CompressibleEquations/time_discretizations.jl
+++ b/src/CompressibleEquations/time_discretizations.jl
@@ -50,10 +50,12 @@ function SplitExplicitTimeDiscretization(FT=Float64;
                                          forward_weight=FT(0.6),
                                          divergence_damping_coefficient=FT(0.10),
                                          acoustic_damping_coefficient=FT(0.0))
-    return SplitExplicitTimeDiscretization(substeps,
-                                           forward_weight,
-                                           divergence_damping_coefficient,
-                                           acoustic_damping_coefficient)
+    return SplitExplicitTimeDiscretization(
+        substeps,
+        FT(forward_weight),
+        FT(divergence_damping_coefficient),
+        FT(acoustic_damping_coefficient),
+    )
 end
 
 """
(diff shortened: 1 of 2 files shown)


giordano:
With this PR and a fix for #679 (for the time being I forced `Float32` in the referenced kernel launches) I can successfully run all the tests in https://github.com/NumericalEarth/Breeze.jl/blob/29b798cf066d4bbf555350df33b11d6f91338106/test/acoustic_substepping.jl (after some tweaking) on Metal GPU using Float32

glwagner (approved):

glwagner:
nice catch!

glwagner:
Metal tests would catch this I think. Although one would think it would also be possible through some GPUCompiler trick, but the only way I know how to catch tricky missing FT is by running on Metal.

giordano:
> Metal tests would catch this I think.

Yes, that's what I'm working towards 😁

src/CompressibleEquations/time_discretizations.jl
```diff
@@ -45,14 +45,17 @@ struct SplitExplicitTimeDiscretization{N, FT}
...
 
-function SplitExplicitTimeDiscretization(; substeps=nothing,
-                                           forward_weight=0.6,
-                                           divergence_damping_coefficient=0.10,
-                                           acoustic_damping_coefficient=0.0)
-    return SplitExplicitTimeDiscretization(substeps,
-                                           forward_weight,
-                                           divergence_damping_coefficient,
-                                           acoustic_damping_coefficient)
+function SplitExplicitTimeDiscretization(FT=Float64;
```
glwagner, line 48: I'm sorry I missed this 😢 but this should be

```julia
FT = Oceananigans.defaults.FloatType
```

that way users do not have to change this manually to Float32, they can change the global default instead

glwagner: cc @giordano

PR #680, opened 2026-05-07
Use consistent float type when launching `_update_pressure_and_average!`

giordano:
Fix #679.  Follow up to #678 to make Metal GPUs work better.

PR #682, opened 2026-05-07
Support Metal GPUs in benchmarks

giordano:
After #678 and #680 we can finally run benchmarks on Metal GPUs successfully!

This is similar to #660, but for Metal GPUs.  ***Note***: I think only the compressible dynamics benchmark will work, because on Metal GPUs we can't do FFTs.  For example you can run
```
julia --project run_benchmarks.jl --device 'MetalGPU' --dynamics "compressible_splitexplicit" --microphysics "nothing" --advection "WENO5" --size "512x512x256" --warmup_steps 1 --time_steps 2
```
(I used very low number of steps because performance can be pretty underwhelming on M1, should hopefully be better with later generations)


glwagner (approved):
nice!

PR #687, opened 2026-05-08
Add Reactant correctness check for AtmosphereModel.

### LINKED ISSUE #683: Add Reactant correctness check for `AtmosphereModel` timestepping
Check parity between vanilla and Reactant `CompressibleDynamics` model for relevant topologies. Compile `time_step!` and compare fields, also after a `loop`. Need to workaround https://github.com/EnzymeAD/Reactant.jl/issues/2870 which appears in `store_initial_state!`, possibly by temporarily using a kernel to copy the model fields.

dkytezab:
Closes #683. Checking field equality before and after a single timestep.


giordano (commented):
Shouldn't run only on the GPU jobs? Also, it may be easier to move all these tests in a dedicate directory (`reactant/`?) and then in https://github.com/NumericalEarth/Breeze.jl/blob/d978c46862d40a209283993a3b8576607944749d/test/runtests.jl#L13-L20 automatically filter out all the tests starting with `reactant/`.  ***Edit***: I went on and did that.

dkytezab:
I believe that the errors here https://github.com/NumericalEarth/Breeze.jl/actions/runs/25581771362/job/75102082127?pr=687#step:11:536 are due primarily to https://github.com/EnzymeAD/Reactant.jl/issues/2870. Basically, when we compile with Reactant, we alias the initial state rather than storing it as a copy, causing the RK3 substeps to be incorrect. This enters here
https://github.com/NumericalEarth/Breeze.jl/blob/10c0d0b69bf7239e9a58a50da3c05c7423679b79/src/TimeSteppers/ssp_runge_kutta_3.jl#L150-L156
My workaround was instead to use a kernel to copy each of the prognostic fields, which I can implement now for the time being if that sounds good

There may also be an issue with `maybe_prepare_first_time_step!` that I need to investigate further.

giordano:
I'm temporarily checking out https://github.com/EnzymeAD/Reactant.jl/pull/2984 by @Pangoraw to see how it goes here.

giordano:
There are still correctness errors: https://github.com/NumericalEarth/Breeze.jl/actions/runs/27677328615/job/81855741446?pr=687#step:14:1076
```
Error in testset topology=Periodic, Periodic, Bounded:
Test Failed at /__w/Breeze.jl/Breeze.jl/test/reactant/correctness.jl:105
  Expression: report_state("topology=$(label) — after  time_step!", vmodel, rmodel; rtol, atol)

Error in testset topology=Periodic, Bounded,  Bounded:
Test Failed at /__w/Breeze.jl/Breeze.jl/test/reactant/correctness.jl:105
  Expression: report_state("topology=$(label) — after  time_step!", vmodel, rmodel; rtol, atol)
```

dkytezab:
I think there is an issue with `maybe_prepare_first_time_step`. Correct me if I am wrong, but i believe that otherwise the initial tendency is different for the different models.

Pangoraw:
> I think there is an issue with maybe_prepare_first_time_step

should we call first_time_step! in the correctness test so that initialization is the same for Reactant and vanilla models?

giordano:
Where are you looking at?

giordano:
Reactant tests are timing out again, the way they did before #782

giordano:
Given that also #774 is stuck in Reactant tests, I take something is up in post v0.2.266?

.github/workflows/CI.yml
```diff
@@ -123,6 +123,7 @@ jobs:
               # concurrent jobs, some of them reaching a large memory usage.  We reduce the
               # number of jobs to avoid memory issues.
               TEST_ARGS+=(--jobs=$(($(nproc) - 1)))
+              TEST_ARGS+=(--quickfail)
```
giordano: I've been thinking for a while of setting `--quickfail` for all jobs, maybe I'll open a separate PR later to just do that

Pangoraw: I suspect that we could learn more from the failure with it since the timeout seems to prevent the buffered results from being shown.

giordano: In this case were jobs even time out that's really necessary, but in general it's a bit frustrating having to wait long time to find what's failing, which was my motivation for doing this.

I opened #798, elaborating the downside and its mitigation.

Pangoraw: Ok with quickfail, we actually catch the error (not a timeout)

```
2026-06-23T21:53:36.4406195Z   UNKNOWN: <unknown>:0: error: 'math.ctlz' op unsupported op for export to XLA
2026-06-23T21:53:36.4406485Z   <unknown>:0: note: see current operation: %1013 = "math.ctlz"(%1003) : (tensor<8x8x8xi64>) -> tensor<8x8x8xi64>
```

which should be solved once [https://github.com/EnzymeAD/Enzyme-JAX/pull/2578](<https://github.com/EnzymeAD/Enzyme-JAX/pull/2578>) hits a new Reactant_jll

.github/workflows/CI.yml
```diff
@@ -125,6 +125,7 @@ jobs:
...
```
Pangoraw:
```suggestion
```

.github/workflows/CI.yml
wsmoses: I'm more concerned as to who emits ctlz. I'm assuming its not coming from you guys, which implies it may be the defn of some libdevice implementation for sinpi or something. Obviously we shouldn't import that version but sinpi [or whatever is the actual cause] directly

Pangoraw: Pinpointed it to `fma_emulated` used in `ConvertingDivision{Float32}`. It itself calls [`_exponent_finite_nonzero`](https://github.com/JuliaLang/julia/blob/e1b2c72e96df0172243407c5e459de0dbac61cc7/base/math.jl#L954) . Maybe we it could makes sense to use `NativeDivision` for Reactant since it does not have fma?

dkytezab:
This looks good to me? I think we should also add a timestepping loop but that can be a subsequent PR imo, anticipating other correctness issues

test/reactant/correctness.jl
```diff
@@ -0,0 +1,109 @@
...
+using Oceananigans.Grids: Periodic, Bounded
+using Oceananigans.TimeSteppers: first_time_step!
+using Reactant
+using Printf: @printf
+using Test
+using CUDA
+
+if get(ENV, "GITHUB_ACTIONS", "false") == "true"
+    Reactant.MLIR.IR.DUMP_MLIR_ALWAYS[] = true
+    ENV["TMPDIR"] = mkpath(joinpath(@__DIR__, "..", "tmp"))
```
giordano: The directory is created in the workflow file now
```suggestion
```

test/reactant/correctness.jl
```diff
@@ -0,0 +1,109 @@
...
+end
+Reactant.Compiler.SROA_ATTRIBUTOR[] = true
```
giordano: We probably don't need this anymore (it's the default in Reactant)
```suggestion
```

giordano:
> This looks good to me?

Yes!  But we need to wait for https://github.com/EnzymeAD/Reactant.jl/pull/2984 making it into a proper release

giordano:
With https://github.com/EnzymeAD/Reactant.jl/pull/2984 now merged we could resume this, but we need to wait for #923 to be merged first (which is waiting for https://github.com/EnzymeAD/Enzyme-JAX/issues/2999 to be resolved, because we have a massive performance regression at the moment)

PR #688, opened 2026-05-09
[CI] Run benchmarks on macOS with Metal GPU

giordano:
Follow up to #682.  Not sure it's entirely worth it, but it may be useful to ensure there are no spurious Float64 in some of the kernels (which we could also do in the tests, without adding more workloads to the benchmarks)

### DIFF
--- a/.github/workflows/Benchmarks.yml
+++ b/.github/workflows/Benchmarks.yml
@@ -86,7 +86,7 @@ jobs:
             advection: 'WENO5'
             backend: 'reactant'
             topology: 'PBB'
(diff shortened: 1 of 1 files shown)


giordano:
>  but it may be useful to ensure there are no spurious Float64 in some of the kernels (which we could also do in the tests, without adding more workloads to the benchmarks)

I added testing of acoustic substepping on Metal in #689, I'm closing this PR for now because adding yet another macOS job for each PR puts more strain on the CI infrastructure (we have only 5 concurrent macOS runners across the entire organisation)

PR #693, opened 2026-05-12
Reactant benchmarking F64 --> F32

dkytezab:
Dependent on https://github.com/CliMA/Oceananigans.jl/pull/5570


giordano (approved):
If it works, I'm happy with this change (and it'd hopefully make the runs slightly faster)

dkytezab:
AD is 4x faster and forward is about 10x faster for each!

giordano:
Any chance we can do more than one step for the autodiff model now? 🥲

dkytezab:
We can certainly try!

PR #694, opened 2026-05-13
Only load GPU packages on platforms where they're available

giordano:
This avoids annoying logging messages on platforms where the backend packages aren't available (e.g. `Metal.jl` on Linux or `AMDGPU.jl` on macOS)


giordano:
Oof, nevermind, we always need to load Metal.jl in the benchmarking package itself because there are a few functions defined on the backend only.

PR #701, opened 2026-05-14
[CI] Give names to steps in benchmarks job

giordano:
~After compilation, the run is actually fast (0.05 s), so we can run it for all 80 timesteps as is the default.  CC @dkytezab~ ***Edit***: PR repurposed to only give names to the steps, which is something I've been meaning to do for a while

### DIFF
--- a/.github/workflows/Benchmarks.yml
+++ b/.github/workflows/Benchmarks.yml
@@ -81,7 +81,7 @@ jobs:
             advection: 'WENO5'
             backend: 'reactant'
             topology: 'PBB'
-            extra_flags: '--ad --simplified --float_type=Float64 --warmup_steps 1'
+            extra_flags: '--ad --simplified --float_type=Float64 --warmup_steps 1 --time_steps 1'
     defaults:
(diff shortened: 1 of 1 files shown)


dkytezab:
Yes however memory may explode. I had trouble fitting a grid larger than 64x64x32 with a single timestep.

giordano:
Wait, the number of timesteps is part of the compiled function?  I noticed compilation is taking _a lot_ longer than when performing a single time step.

dkytezab:
A single time step is simpler to compile because there is no loop. Checkpointing generates a lot of compilation overhead I believe

giordano:
```
W0000 00:00:1778775119.497787     400 hlo_rematerialization.cc:3233] Can't reduce memory use below 16.75GiB (17983925176 bytes) by rematerialization; only reduced to 51.71GiB (55522772736 bytes), down from 51.76GiB (55581009808 bytes) originally
```
:headstone: 

Ok, yeah, no, this isn't practical.  Sigh.

PR #704, opened 2026-05-14
Moist Exner reference state + splitting-supercell comparison

kaiyuan-cheng:
## Summary

**Goal:** suppress the spurious acoustic / gravity-wave noise that the compressible solver radiates when it starts from a state that does not satisfy its own discrete hydrostatic balance. Two distinct sources of imbalance are fixed:

1. **Reference state imbalance (moist atmospheres).** `ExnerReferenceState` was dry-only, so any moist run sat on top of a reference that did *not* satisfy the moist discrete hydrostatic balance — the slow vertical-momentum tendency was non-zero on a rest moist atmosphere, exciting waves before the simulation even gets going. Unify the dry and moist Exner kernels into a
   single `_compute_exner_reference!` taking `qᵛ` (a `ZeroField` for the dry case). The moist gas constants `Rᵐ, cᵖᵐ, κᵐ` reduce to `Rᵈ, cᵖᵈ, κ` when `qᵛ ≡ 0`, so the dry path is bit-identical and the moist path satisfies the same discrete balance the substepper relies on. `CompressibleDynamics` gains a `reference_vapor_mass_fraction` kwarg; the `ref_spec → ExnerReferenceState` kwargs translation is centralized in a single `_exner_kwargs(ref_spec)` helper shared by `materialize_dynamics` and `boundary_conditions_reference_state`.

2. **Initial-condition imbalance (θ perturbations).** Naively setting `ρ = ρ_reference` while perturbing θ changes `ρθ` → changes pressure via the EOS → radiates an acoustic pulse from the perturbation as soon as `t > 0`. Add `pressure_balanced_density(ρ_bg, θ_bg, θ_init) = ρ_bg · θ_bg / θ_init`; rescaling the reference ρ with this helper preserves `ρθ` (and so pressure) under the θ perturbation, so the warm bubble in the splitting-supercell example enters the run as a buoyancy perturbation only, not also as a pressure perturbation.

3. **Tutorial example.** Rewrite `examples/splitting_supercell.jl` as an anelastic-vs-compressible tutorial that walks through reference state, dynamics, microphysics, advection, `AtmosphereModel`, and `set!` component-by-component for the anelastic core, then introduces only the deltas (split-explicit substepping, moist Exner reference, `pressure_balanced_density` for ρ) for the compressible core. Produces stacked xy (z ≈ 5 km) and xz (y = Ly/2) comparison animations plus a max-vertical-velocity overlay, demonstrating that both cores now evolve from the same well-balanced initial state.


glwagner:
Should we add a closure to the supercell too (eg #671)?

examples/splitting_supercell.jl
```diff
@@ -230,159 +240,330 @@ Colorbar(fig[1, 2], hm, label="θ′ (K)")
...
+    )
+
+    slices_filename = "splitting_supercell_$(label)_slices.jld2"
+    simulation.output_writers[:slices] = JLD2Writer(model, slice_outputs;
+        filename=slices_filename, schedule=TimeInterval(2minutes), overwrite_existing=true)
+
+    CUDA.synchronize()
+    t0 = time_ns()
+    run!(simulation)
+    CUDA.synchronize()
```
glwagner: the timing block may not be needed... the run time can be diagnosed from `sim.run_wall_time`

kaiyuan-cheng:
> Should we add a closure to the supercell too (eg #671)?

I don't think it is necessary. CM1 can produce a splitting supercell nicely without closure.

examples/splitting_supercell.jl
```diff
@@ -230,159 +240,330 @@ Colorbar(fig[1, 2], hm, label="θ′ (K)")
...
+# ``z \approx 5 \, {\rm km}``) and vertical (`xz` at ``y = L_y/2``) slices, and
+# collects the maximum vertical-velocity time series. We use this same driver for
+# both the anelastic and compressible runs.
+
+function run_simulation(model, label)
+    @info "=== Running case: $label ==="
+
+    θ          = liquid_ice_potential_temperature(model)
+    θ_snapshot = deepcopy(θ)            ## snapshot of the initial state on the GPU
+    θ′         = Field(θ - θ_snapshot)  ## lazy field, recomputed each output
```
```suggestion
    θ  = liquid_ice_potential_temperature(model)
    θ′ = Field(θ - deepcopy(θ))  ## lazy field, recomputed each output
```

slightly easier to read I think

examples/splitting_supercell.jl
```diff
@@ -230,159 +240,330 @@ Colorbar(fig[1, 2], hm, label="θ′ (K)")
...
+
+    qᶜˡ = model.microphysical_fields.qᶜˡ
+    qʳ  = model.microphysical_fields.qʳ
+    qᵛ  = model.microphysical_fields.qᵛ
+    u, v, w = model.velocities
+
+    simulation = Simulation(model; Δt=2, stop_time=2hours)
+    conjure_time_step_wizard!(simulation, cfl=0.7)
```
glwagner, line 330: i wonder if we should comment that `cfl` can differ for compressible...

examples/splitting_supercell.jl
```diff
@@ -230,159 +240,326 @@ Colorbar(fig[1, 2], hm, label="θ′ (K)")
...
+
+# Build the model — same constructor, different dynamics:
+
+model_compressible = AtmosphereModel(grid; dynamics = dynamics_compressible,
+                                     microphysics, advection,
+                                     thermodynamic_constants = constants)
+
+# Initial density. Naively setting `ρ` to the reference density while perturbing
+# `θ` would change `ρθ` and seed a spurious acoustic pulse from the warm bubble.
+# Instead we rescale the reference density via [`pressure_balanced_density`](@ref)
```
giordano:
```suggestion
# Instead we rescale the reference density via [`pressure_balanced_density`](@ref Breeze.Thermodynamics.pressure_balanced_density)
```

glwagner:
> > Should we add a closure to the supercell too (eg #671)?
> 
> I don't think it is necessary. CM1 can produce a splitting supercell nicely without closure.

True. I guess the purpose would be to reproduce the intercomparison case, but also it's not a good demonstration of a good model setup. So I agree with leaving the closure out.

src/CompressibleEquations/terrain_compressible_physics.jl
```diff
@@ -362,58 +362,140 @@ using GPUArraysCore: @allowscalar
...
+function terrain_hydrostatic_pressure(z, p₀, θᵣ, qᵛᵣ, pˢᵗ, constants)
+    z == 0 && return p₀
+
+    # Anchor for the k = 1 cell center: integrate the continuous hydrostatic
+    # equation dp/dz = -g ρ from the surface to the local physical height `z`
+    # with a midpoint rule. This only sets the column's pressure level; the
+    # discrete-balance recurrence in `compute_terrain_reference_state!` handles
+    # k ≥ 2. This is the moist analogue of `numerically_integrated_hydrostatic_pressure`
+    # and uses the same fixed `nsteps`.
+    nsteps = 1000
```
glwagner: why don't we use a convergence criterion?

kaiyuan-cheng: I wondered about this too, and traced it back to its source. It turned out that Claude follows https://github.com/NumericalEarth/Breeze.jl/blob/7dfd547d466b7808b773d2a777d5b75ca573af86/src/Thermodynamics/reference_states.jl#L225

glwagner: Probably we should not duplicate this code (if possible), and it might be good to add a proper solver?

kaiyuan-cheng: Agree. I've updated the solver here and revised the solver in `reference_stages.jl`.

benchmarking/Project.toml
giordano, line 1: These changes shouldn't be in this branch, I guess the rebase went wrong.  I'm going to do a git surgery to fix this.

examples/splitting_supercell.jl
ewquon: I'd be curious to see this, too. In my DCMIP2016 TC case, I was able to run with CFL > 1.1. This would make compressible on par with anelastic in terms of walltimes (which are reported in the example).

ewquon (approved):
Looks good to me!

PR #708, opened 2026-05-15
Accept specific-quantity keys in AtmosphereModel forcings

### LINKED ISSUE #703: Streamline forcings by accepting specific-quantity keys
## Motivation

Today, prescribed-tendency forcings in BOMEX, RICO, and similar examples require users to manually convert specific tendencies (K/s, kg/kg/s, J/kg/s) into density-weighted tendencies before handing them to the model. Each forcing involves the same boilerplate:

```julia
ρᵣ = reference_state.density
Fρθ_field = Field{Nothing, Nothing, Center}(grid)
set!(Fρθ_field, z -> ∂t_θ_profile(z))
set!(Fρθ_field, ρᵣ * Fρθ_field)          # manual specific→density
ρθ_forcing = Forcing(Fρθ_field)
forcing = (; ρθ=ρθ_forcing, ...)
```

This is awkward for two reasons:

1. The user must thread `ρᵣ` through their setup even though the model already owns it.
2. The physical reasoning is in specific units, but the API key is density-weighted.

`set!(model; θ=…)` already accepts specific names and converts to density internally (`set_atmosphere_model.jl:86`). Forcings should mirror that ergonomic.

This also paves the way for using Oceananigans' built-in forcings (`Forcing`, `Relaxation`, including FTS support landing in CliMA/Oceananigans.jl#5575) directly in Breeze without users needing to know about Breeze's prognostic-density convention.

## Plan

Path: keep existing Breeze forcings (`SubsidenceForcing`, `GeostrophicForcing`) `ρ`-keyed because they already produce density tendencies internally. Add a new code path for *generic* user-supplied forcings to be keyed under specific names, with Breeze applying the density multiply at kernel time.

### New code

- **`src/Forcings/specific_forcing.jl`** — new `SpecificForcing{F, D, LZ}` wrapping a materialized inner forcing, a density field, and the vertical location of the target field. Kernel callable returns `ρ_at_target_location * inner(i, j, k, grid, clock, fields)`. For `Face`-located targets (e.g. `w`), density is interpolated to face via `ℑzᵃᵃᶠ`. Mirrors the existing pattern of `SubsidenceForcing`/`GeostrophicForcing` (struct + materialize + kernel callable + Adapt + show).

### Dispatch change

- **`src/AtmosphereModels/atmosphere_model.jl`** — extend `atmosphere_model_forcing(::NamedTuple, …)` to:
  - Build a `specific_to_density_name` map from `prognostic_fields` keys plus `moisture_specific_name(microphysics)` (e.g. `:u→:ρu, :v→:ρv, :w→:ρw, :θ→:ρθ, :e→:ρe, :qᵛ→:ρqᵛ`).
  - When a user key is a specific name, wrap the value (or each tuple element) in `SpecificForcing`, materialize, and store under the corresponding ρ-key. ρ-keys keep current behavior.
  - **Error** when a Breeze density-tendency forcing (`SubsidenceForcing`, `GeostrophicForcing`) is supplied under a specific key — these already produce density tendencies, so auto-wrapping would double-count. Message points the user to the corresponding ρ-key.

### Always multiply at kernel time

Uniform code path for anelastic (ρᵣ(z)) and compressible (ρ(x,y,z,t)). One inbounds load + multiply per grid point per RK substep — negligible cost relative to the surrounding tendency arithmetic, and required anyway for compressible or time-dependent inner forcings.

### Examples

- `examples/bomex.jl` — drop the `ρᵣ * Field` blocks for the moisture drying and radiative cooling forcings. Supply as `qᵉ = Forcing(…)`, `e = Forcing(…)`.
- `examples/rico.jl` — same for the large-scale moisture and θ tendencies. Sponge stays `ρw = sponge`.

### Tests and docs

- New tests in the forcings test file covering: specific keys produce the expected ρ-tendency in both anelastic and compressible formulations; `w` specific key works at Face; duplicate `θ`/`ρθ` errors; Breeze forcing under specific key errors; tuple composition.
- A short subsection in the forcings docs explaining the two equivalent ways and when each is convenient.

## Known trade-off

Because Breeze's built-in `SubsidenceForcing` and `GeostrophicForcing` continue to be `ρ`-keyed (they encapsulate the density multiply internally), examples will end up with a mix of key styles, e.g.

```julia
forcing = (; ρθ = subsidence,
             θ  = Forcing(radiative_cooling),
             ρu = (subsidence, geostrophic.ρu),
             ...)
```

This is intentional: the key signals whether the user is supplying a specific tendency (Breeze multiplies by ρ) or a density tendency (Breeze applies as-is). Refactoring `SubsidenceForcing` / `GeostrophicForcing` to return specific tendencies would unify the keys but is out of scope for this branch.

## Out of scope

- `FieldTimeSeriesRelaxation` and the nudging examples — these are being superseded by Oceananigans' built-in `Relaxation` with FTS support (CliMA/Oceananigans.jl#5575).
- Refactoring `SubsidenceForcing` / `GeostrophicForcing` to return specific tendencies (separately considered; not part of this branch).

ewquon:
Closes #703.

## Summary

Breeze forcings can now be supplied under specific (per-unit-mass) prognostic names — `u`, `v`, `w`, `θ`, `e`, `qᵉ`, `qᵛ`, … — and Breeze applies the density factor `ρ` automatically at kernel time. The existing density-weighted keys (`ρu`, `ρθ`, …) continue to work for forcings whose kernels produce density tendencies directly.

- **New `SpecificForcing{F, D, L}` in `src/Forcings/`**: stores a user-supplied specific tendency, the model density, and the target prognostic's `(LX, LY, LZ)` location. The kernel returns `ρ(at target location) * F`, with `ρ` interpolated to the appropriate cell face for `u` (x-Face), `v` (y-Face), and `w` (z-Face) via `ℑxᶠᵃᵃ` / `ℑyᵃᶠᵃ` / `ℑzᵃᵃᶠ`. One code path for anelastic `ρᵣ(z)` and compressible `ρ(x, y, z, t)`.

- **`atmosphere_model_forcing` accepts specific names alongside ρ-prefixed ones**: specific entries are auto-wrapped in `SpecificForcing`. When the same prognostic gets both forms (`θ=…, ρθ=…`), contributions are summed via `MultipleForcings`. User-defined density-tendency forcings supplied under a specific key are rejected by the `is_density_tendency_forcing` trait to prevent double-counting ρ.

- **`SubsidenceForcing` and `GeostrophicForcing` rewritten to return specific tendencies** (the ρ-multiply and horizontal interpolation for `u`/`v` now happen in `SpecificForcing` instead of inside their own kernels). `geostrophic_forcings(uᵍ, vᵍ)` returns `(; u, v)`. Supplying either type under a ρ-prefixed key errors with a migration message — this is the user-visible breaking change.

- **`examples/bomex.jl` and `examples/rico.jl`** drop the manual `ρᵣ * field` boilerplate; all forcings are keyed by specific name.

### Breaking change

Code that previously supplied `SubsidenceForcing` under `ρθ`/`ρu`/`ρqᵉ`/…, or `geostrophic_forcings(...).ρu` / `.ρv`, must move to the specific keys. The construction-time error message describes the migration. The plain ρ-keyed `Forcing(...)` API is unchanged.

## Test plan

- [x] `test/specific_forcing.jl` (new, 28 tests): construction & dispatch, anelastic Center/Face numerical correctness, compressible numerical correctness, mixed θ + ρθ summation, A/B equivalence with the manual `ρᵣ * field` pattern, user-defined-density-tendency-under-specific-key errors, built-in Breeze forcings under ρ-key error with migration message, tuple composition, unknown-key error.
- [x] `test/geostrophic_subsidence_forcings.jl` (updated, 29 tests): all usages migrated to specific keys; new **multi-step equivalence test** asserts that `geostrophic_forcings` produces bit-for-bit identical `ρu`/`ρv` to a hand-rolled `Forcing(ρᵣ * vᵍ)` reference over 10 anelastic time steps; new **multi-step subsidence accumulation test** verifies `ρϕ` evolves linearly under a constant-gradient profile over 5 steps for both `θ` and `qᵛ`.
- [x] `test/forcing_and_boundary_conditions.jl` (66 tests) still pass; one bogus-key test updated since `:u` is now a valid specific alias.
- [ ] BOMEX and RICO examples build / run on CPU.
- [x] CI green.

### DIFF
--- a/src/AtmosphereModels/atmosphere_model.jl
+++ b/src/AtmosphereModels/atmosphere_model.jl
@@ -431,16 +431,35 @@ function atmosphere_model_forcing(user_forcings::NamedTuple, prognostic_fields, 
     return forcings
 end
 
-# Assemble the materialized forcing for one prognostic field, combining any user-supplied
-# ρ-keyed entry with any specific-keyed entry on the same prognostic.
+# Assemble the materialized forcing for one prognostic field. Dispatch on the result of
+# `specific_name_of(density_name)`: a `Symbol` is the specific alias of a ρ-prefixed
+# prognostic; `Nothing` means the prognostic name is already in specific form (e.g. a
+# user tracer like `:c`) and has no separate density alias.
 function assemble_field_forcing(density_name, target_field, user_forcings, model_names, context)
-    specific_name = specific_name_of(density_name)
-    raw_specific_forcing = isnothing(specific_name) ? nothing :
-                           get(user_forcings, specific_name, nothing)
+    return assemble_field_forcing(density_name, specific_name_of(density_name),
+                                  target_field, user_forcings, model_names, context)
+end
+
+# ρ-prefixed prognostic: combine any user-supplied ρ-keyed entry with any specific-keyed
+# entry on the same prognostic, wrapping the specific-keyed entry in SpecificForcing.
+function assemble_field_forcing(density_name, specific_name::Symbol, target_field,
+                                user_forcings, model_names, context)
+    raw_specific_forcing = get(user_forcings, specific_name, nothing)
     wrapped_specific_forcing = wrap_specific_forcing(raw_specific_forcing, density_name)
     raw_density_forcing = get(user_forcings, density_name, nothing)
     combined = combine_forcing_values(raw_density_forcing, wrapped_specific_forcing)
     return materialize_or_default(combined, target_field, density_name, model_names, context)
+end
+
+# Unprefixed prognostic (a user tracer like `:c`): the prognostic name *is* in specific
+# form, so any forcing supplied under that name is a specific tendency and gets wrapped
+# in SpecificForcing — the ρ factor is then applied at kernel time, matching the
+# convention used for ρ-prefixed prognostics' specific aliases.
+function assemble_field_forcing(density_name, ::Nothing, target_field,
+                                user_forcings, model_names, context)
+    raw_forcing = get(user_forcings, density_name, nothing)
+    wrapped_forcing = wrap_specific_forcing(raw_forcing, density_name)
+    return materialize_or_default(wrapped_forcing, target_field, density_name, model_names, context)
 end
 
 # Strip the `ρ` prefix from a density-weighted prognostic name; returns `nothing` for
--- a/src/AtmosphereModels/atmosphere_model.jl
+++ b/src/AtmosphereModels/atmosphere_model.jl
@@ -432,11 +432,11 @@ function atmosphere_model_forcing(user_forcings::NamedTuple, prognostic_fields, 
 end
 
 # Assemble the materialized forcing for one prognostic field. Dispatch on the result of
-# `specific_name_of(density_name)`: a `Symbol` is the specific alias of a ρ-prefixed
+# `specific_name_of(prognostic_name)`: a `Symbol` is the specific alias of a ρ-prefixed
 # prognostic; `Nothing` means the prognostic name is already in specific form (e.g. a
 # user tracer like `:c`) and has no separate density alias.
-function assemble_field_forcing(density_name, target_field, user_forcings, model_names, context)
-    return assemble_field_forcing(density_name, specific_name_of(density_name),
+function assemble_field_forcing(prognostic_name, target_field, user_forcings, model_names, context)
+    return assemble_field_forcing(prognostic_name, specific_name_of(prognostic_name),
                                   target_field, user_forcings, model_names, context)
 end
 
@@ -455,11 +455,11 @@ end
 # form, so any forcing supplied under that name is a specific tendency and gets wrapped
 # in SpecificForcing — the ρ factor is then applied at kernel time, matching the
 # convention used for ρ-prefixed prognostics' specific aliases.
-function assemble_field_forcing(density_name, ::Nothing, target_field,
+function assemble_field_forcing(tracer_name, ::Nothing, target_field,
                                 user_forcings, model_names, context)
-    raw_forcing = get(user_forcings, density_name, nothing)
-    wrapped_forcing = wrap_specific_forcing(raw_forcing, density_name)
-    return materialize_or_default(wrapped_forcing, target_field, density_name, model_names, context)
+    raw_forcing = get(user_forcings, tracer_name, nothing)
+    wrapped_forcing = wrap_specific_forcing(raw_forcing, tracer_name)
+    return materialize_or_default(wrapped_forcing, target_field, tracer_name, model_names, context)
 end
 
 # Strip the `ρ` prefix from a density-weighted prognostic name; returns `nothing` for
--- a/src/AtmosphereModels/atmosphere_model.jl
+++ b/src/AtmosphereModels/atmosphere_model.jl
@@ -422,18 +422,7 @@ function atmosphere_model_forcing(user_forcings::NamedTuple, prognostic_fields, 
     forcing_context = (; coriolis, density, specific_fields)
 
     materialized = Tuple(
-        let density_name = n,
-            specific_name = startswith(string(n), "ρ") ? Symbol(string(n)[nextind(string(n), 1):end]) : nothing
-            ρ_value    = density_name in user_forcing_names ? user_forcings[density_name] : nothing
-            spec_raw   = (specific_name !== nothing && specific_name in user_forcing_names) ?
-                         user_forcings[specific_name] : nothing
-            spec_value = spec_raw === nothing ? nothing :
-                         wrap_specific_forcing(spec_raw, density_name)
-            combined   = combine_forcing_values(ρ_value, spec_value)
-            combined === nothing ?
-                Returns(zero(eltype(f))) :
-                materialize_atmosphere_model_forcing(combined, f, density_name, model_names, forcing_context)
-        end
+        assemble_field_forcing(n, f, user_forcings, model_names, forcing_context)
         for (n, f) in pairs(forcing_fields)
     )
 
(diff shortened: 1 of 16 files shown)


src/Forcings/specific_forcing.jl
```diff
@@ -0,0 +1,86 @@
...
+`ρ(x, y, z, t)` under [`CompressibleDynamics`](@ref Breeze.CompressibleEquations.CompressibleDynamics);
+the same wrapper handles both.
+
+Users typically supply specific forcings directly through specific-named keys
+(`u`, `v`, `w`, `θ`, `e`, `qᵉ`, `qᵛ`, …) in the `forcing` `NamedTuple` passed to
+[`AtmosphereModel`](@ref), and the dispatch wraps each entry in `SpecificForcing`
+automatically. The wrapper can also be constructed directly when finer control is needed.
+
+The inner `forcing` can be anything accepted by Oceananigans' `materialize_forcing`:
+a function `(x, y, z, t)`, a [`Returns`](@ref) callable, a `Field`, an
```
giordano:
```suggestion
a function `(x, y, z, t)`, a `Returns` callable, a `Field`, an
```

src/Forcings/specific_forcing.jl
```diff
@@ -0,0 +1,86 @@
...
+
+for fields located at `Center`, and interpolates ``ρ`` to the target vertical face via
+``ℑzᵃᵃᶠ`` for forcings on `w` (or any `Face`-located target). `ρ` is `ρᵣ(z)` under
+[`AnelasticDynamics`](@ref Breeze.AnelasticEquations.AnelasticDynamics) and the prognostic
+`ρ(x, y, z, t)` under [`CompressibleDynamics`](@ref Breeze.CompressibleEquations.CompressibleDynamics);
+the same wrapper handles both.
+
+Users typically supply specific forcings directly through specific-named keys
+(`u`, `v`, `w`, `θ`, `e`, `qᵉ`, `qᵛ`, …) in the `forcing` `NamedTuple` passed to
+[`AtmosphereModel`](@ref), and the dispatch wraps each entry in `SpecificForcing`
```
giordano:
```suggestion
[`AtmosphereModel`](@ref Breeze.AtmosphereModels.AtmosphereModel), and the dispatch wraps each entry in `SpecificForcing`
```

src/AtmosphereModels/atmosphere_model.jl
```diff
@@ -415,17 +422,52 @@ function atmosphere_model_forcing(user_forcings::NamedTuple, prognostic_fields,
...
-            Returns(zero(eltype(f)))
-            for (n, f) in pairs(forcing_fields)
+        let density_name = n,
+            specific_name = startswith(string(n), "ρ") ? Symbol(string(n)[nextind(string(n), 1):end]) : nothing
+            ρ_value    = density_name in user_forcing_names ? user_forcings[density_name] : nothing
+            spec_raw   = (specific_name !== nothing && specific_name in user_forcing_names) ?
+                         user_forcings[specific_name] : nothing
+            spec_value = spec_raw === nothing ? nothing :
+                         wrap_specific_forcing(spec_raw, density_name)
+            combined   = combine_forcing_values(ρ_value, spec_value)
```
glwagner: I'm having trouble parsing this... using `raw_specific_forcing` and `wrapped_specific_forcing` would help a lot (I didn't understand at first that "spec_raw" was a forcing).

Dispatch rather than if statements might also help for interpreting this, I'll think about that. Generally its better to use fewer if statements whenever possible

src/AtmosphereModels/atmosphere_model.jl
```diff
@@ -415,17 +422,52 @@ function atmosphere_model_forcing(user_forcings::NamedTuple, prognostic_fields,
     forcing_context = (; coriolis, density, specific_fields)
 
     materialized = Tuple(
-        n in keys(user_forcings) ?
-            materialize_atmosphere_model_forcing(user_forcings[n], f, n, model_names, forcing_context) :
-            Returns(zero(eltype(f)))
-            for (n, f) in pairs(forcing_fields)
+        let density_name = n,
```
glwagner: why `let` block?

src/AtmosphereModels/atmosphere_model.jl
glwagner:
```suggestion
            raw_specific_forcing = get(user_forcings, specific_name, nothing)
            wrapped = wrap_specific_forcing(raw_specific_forcing, density_name)
            combined  = combine_forcing_values(wrapped, density_name)
```

along with

```julia
wrap_specific_forcing(::Nothing, density_name) = nothing
```

src/AtmosphereModels/atmosphere_model.jl
```diff
@@ -415,17 +422,52 @@ function atmosphere_model_forcing(user_forcings::NamedTuple, prognostic_fields,
...
+            specific_name = startswith(string(n), "ρ") ? Symbol(string(n)[nextind(string(n), 1):end]) : nothing
+            ρ_value    = density_name in user_forcing_names ? user_forcings[density_name] : nothing
+            spec_raw   = (specific_name !== nothing && specific_name in user_forcing_names) ?
+                         user_forcings[specific_name] : nothing
+            spec_value = spec_raw === nothing ? nothing :
+                         wrap_specific_forcing(spec_raw, density_name)
+            combined   = combine_forcing_values(ρ_value, spec_value)
+            combined === nothing ?
+                Returns(zero(eltype(f))) :
```
glwagner:
```suggestion
                Returns(zero(grid)) :
```

I find this easier to read

ewquon:
Okay, I believe Claude has addressed all of your comments/concerns in [0e5bbd6](https://github.com/NumericalEarth/Breeze.jl/pull/708/commits/0e5bbd6baf6319582e7f1512256da67a0fd4df69), @glwagner

src/AtmosphereModels/atmosphere_model.jl
```diff
@@ -442,14 +431,39 @@ function atmosphere_model_forcing(user_forcings::NamedTuple, prognostic_fields,
     return forcings
 end
 
+# Assemble the materialized forcing for one prognostic field, combining any user-supplied
+# ρ-keyed entry with any specific-keyed entry on the same prognostic.
+function assemble_field_forcing(density_name, target_field, user_forcings, model_names, context)
+    specific_name = specific_name_of(density_name)
+    raw_specific_forcing = isnothing(specific_name) ? nothing :
```
glwagner: why is `specific_name` nothing?

glwagner: sorry dont meant to be difficult 😂 I just don't understand what this code is trying to do by reading it. we get density name. If density name is NOT prefixed by rho --- what do we want to happen?

glwagner (approved):
This looks great

navidcy:
After this PR all forcings are prescribed for u, v, w and the multiplication with ρ if applicable is done under the hood, am I right?

ewquon:
Forcings may be prescribed for `ρu`, `ρv`, `ρw`, `ρc` –OR– `u`, `v`, `w`, `c` in which case the multiplication by ρ is done under the hood. Mixing and matching of density-weighted / specific quantities is possible.

PR #712, opened 2026-05-16
Acoustic substepping on terrain-following coordinates

### LINKED ISSUE #673: Substepping with terrain-following coordinates: implementation plan
@ewquon @kaiyuan-cheng

## TL;DR

Acoustic substepping (`SplitExplicitTimeDiscretization`) doesn't yet compose with terrain-following coordinates. Below is an audit of every substepper kernel that needs an update.

## Background

Two existing capabilities don't yet compose:

- **Acoustic substepping** (`AcousticSubstepper`, `SplitExplicitTimeDiscretization`) — works on `LatitudeLongitudeGrid` and Cartesian with **uniform** vertical discretization. ~100× larger outer Δt than fully-explicit.
- **Terrain-following coordinates** (`TerrainMetrics`, `TerrainCompressibleDynamics`) — works only with `ExplicitTimeStepping`. Implements contravariant Ω̃, two PG stencils (`SlopeOutsideInterpolation` / `SlopeInsideInterpolation`), and uses Oceananigans' generalized `∂xᶠᶜᶜ`/`∂yᶜᶠᶜ` (which include the chain-rule correction on `MutableVerticalDiscretization`).

Combining them requires updating every substepper kernel to:
1. Use **area-weighted** vertical-flux divergences instead of plain `∂zᶜᶜᶜ`.
2. Replace `(ρw)′` with **`(ρΩ̃)′`** wherever the substepper transports mass and θ vertically.
3. Apply the **terrain-corrected horizontal PGF** (linearized) inside `_explicit_horizontal_step!`.
4. Honor the bottom BC `Ω̃(k=1) = 0`, not `w(k=1) = 0`.

(The vertical implicit solve already handles variable Δz; tested on a 33×-stretched column and all eigenvalues are real-positive with LAPACK residual 8e-12. The substep count is bound by the horizontal acoustic CFL — neither is affected by terrain.)

### `_explicit_horizontal_step!` (lines 838–861)
On `MutableVerticalDiscretization`, `∂xᶠᶜᶜ` already does `(∂φ/∂x)_z = (∂φ/∂x)_ζ − slope · ∂z(φ)`. So the linearized horizontal PGF should mostly work; need to validate that `ℑxᶠᵃᵃ(grid, Πᴸ)` and `ℑxᶠᵃᵃ(grid, γRᵐᴸ)` give correct slope-aware face values on the deformed grid. Should also pick a stencil convention (likely `SlopeOutsideInterpolation` since it composes better with the substepper's frozen-pᴸ).

### `_build_predictors_and_vertical_rhs!` (lines 883–945)
Three places use plain `∂zᶜᶜᶜ` and `ℑz` without area weighting:

```julia
ρ′★[k] = ρ′[k] + Δτ·(Gˢρ - ∇ʰ_M) - δτˢ⁻ · ∂zᶜᶜᶜ(grid, ρw′)         # ✗ NOT area-weighted
ρθ′★[k] = ρθ′[k] + Δτ·(Gˢρθ - ∇ʰ_θM) - δτˢ⁻ · ∂zᶜᶜᶜ(theta_face_z_flux, …)
                                                                    # ✗ NOT area-weighted
```

On `LatitudeLongitudeGrid` `Az` is z-independent → `(1/V) δz(Az·X) = ∂zᶜᶜᶜ(X)` exactly, which is why the current code works there. With variable `Az(k)`, this breaks.

**Fix**: replace
```julia
∂zᶜᶜᶜ(grid, ρw′) → (1/V) · δzᵃᵃᶜ(Az_qᶜᶜᶠ, ρw′)
```
and update `theta_face_z_flux` to include `Az_qᶜᶜᶠ`. Also swap `(ρw)′` for `(ρΩ̃)′` in terrain mode.

### Tridiagonal `get_coefficient` (lines 618–668)
Coefficients already use `Δzᶜᶜᶠ(i,j,k)`, `Δzᶜᶜᶜ(i,j,k)`. **Variable Δz verified** on a 33×-stretched column (`scratch/spurious_mode/column_eigenvalue_stretched.jl`): all eigenvalues real-positive, LAPACK cross-check passes. No change needed.

### `_post_solve_recovery!` (lines 968–972)
Same `∂zᶜᶜᶜ(ρw′)` and `∂zᶜᶜᶜ(ℑb_wθ, ρw′, θᴸ)` issue as the predictor. Same fix.

### `_assemble_slow_vertical_momentum_tendency!` (lines 723–746)
Currently computes the frozen z-direction PGF + buoyancy. Need to derive the terrain-aware slow tendency for `(ρΩ̃)′` from the Klemp 2011 contravariant momentum equation.

glwagner:
Closes #673.

Adds terrain-following coordinates and composes them with the acoustic
substepper. The public API is `TerrainFollowingVerticalDiscretization`
(TFVD) with two formulations: `LinearDecay` (Gal-Chen & Somerville 1975)
and `TwoLevelDecay` (Schär et al. 2002 — the "Smooth LEvel VErtical"
coordinate). The legacy `follow_terrain!`/MVD path has been removed —
TFVD is the only terrain-following API.

## Setting up terrain: Witch of Agnesi

```julia
using Oceananigans, Breeze

Nx, Nz   = 400, 200
Lx, Lz   = 200e3, 25e3
h₀, a    = 250.0, 10e3
hill(x, y) = h₀ / (1 + (x / a)^2)

z_faces = TerrainFollowingVerticalDiscretization(
    collect(range(0, Lz, length = Nz + 1));
    formulation = TwoLevelDecay(large_scale_height = Lz / 2,
                                small_scale_height = 2.5e3))

grid = RectilinearGrid(size = (Nx, Nz), halo = (5, 5),
    x = (-Lx/2, Lx/2), z = z_faces,
    topology = (Periodic, Flat, Bounded))

materialize_terrain!(grid, hill)

    reference_potential_temperature = z -> 280 * exp(1e-4 * z / 9.81),
    surface_pressure = 1e5)

model = AtmosphereModel(grid; dynamics, advection = WENO(order = 9))
```

The pressure-gradient stencil (`TerrainMetrics`) is built automatically by
`CompressibleDynamics` from the grid. Default stencil is
`SlopeOutsideInterpolation()`; override with
`CompressibleDynamics(...; slope_stencil = SlopeInsideInterpolation())`.
For `TwoLevelDecay` the `materialize_terrain!` step splits `h` into
smoothed `h₁` (slow decay) and residual `h₂` (fast decay) automatically.

## What this PR does

  - Terrain-corrected horizontal pressure gradient (slope-outside and
    slope-inside stencils). Stencil is selected via `slope_stencil =
    ...` on `CompressibleDynamics`; the `TerrainMetrics` object is built
    automatically from the grid — no user-facing builder call required.
  - Contravariant vertical momentum `ρw̃ = ρw − slope·ρu − slope·ρv` used
    by all vertical transport; surface BC `ρw̃ = 0`.
  - Slow vertical-momentum tendency in perturbation form
    `Gˢρw = -∂z p' - g ρ'` so a well-balanced reference state suppresses
    the `∂z p ≈ -gρ` cancellation error.
  - Per-column discrete-Exner reference state
    (`terrain_reference_pressure`, `terrain_reference_density`).
  - `Oceananigans.Grids.node` override on TFVD so `set!(field, (x,y,z) ->
    f(z))` correctly evaluates `f` at physical altitude (`znode`), not
    reference `ζ` (`rnode`). Without this the IC is out of discrete
    hydrostatic balance and excites a spurious surface column.
  - Removes the legacy `follow_terrain!`/MVD path, `BasicTerrainFollowing`,
    `GridFittedTerrain`, `FaceSampledTerrain`. TFVD is the only path.

## Validation

Schär 2D (`h₀=250 m`, `a=5 km`, `λ=4 km`, `N=0.01 /s`, `U=10 m/s`,
6 h) below sponge, 400×200 vs matched CM1 `psolver=3`:

| metric | value | target | pass |
|---|---|---|---|
| `w` amplitude error (max) | 5.1 % | <5-10 % | ✓ |
| `w` phase error | 1.5e-8 λ | <0.1 | ✓ |
| `w` normalized RMSE | 10.2 % | <10 % | ✓ (edge) |
| `θ'` normalized RMSE | 8.9 % | <10 % | ✓ |
| flat-terrain equivalence | 1-step + 10-step bit-exact | — | ✓ |
| hydrostatic rest over terrain | ~1e-13 residual | — | ✓ |
| no-normal-flow bottom | bottom `w̃ ≡ 0` | — | ✓ |
| metric identities | manufactured-field tests pass | — | ✓ |
| acoustic stability CFL | passes | — | ✓ |
| GPU vs CPU (H100) | diff 1e-11–1e-7 | — | ✓ |
| flat-terrain perf overhead | 0.5 % (target <3 %) | — | ✓ |
| nonflat perf overhead | 8.3 % (target <10 %) | — | ✓ |

Local tests on this branch:

  - `test/terrain_following.jl` — passes (migrated to the auto-build
    API; covers default stencil, override via `slope_stencil`, and the
    pre-built-metrics escape hatch).
  - `acoustic_substepping`, `quality_assurance`, `doctests`,
    `dcmip2016_kessler` — all pass.

## Docs

  - `docs/src/terrain_following_coordinates.md` — rewritten around TFVD:
    continuous theory, discrete operators, well-balanced reference state,
    BCs, the `set!`/`rnode` pitfall and the `node()` override that fixes
    it, plus a worked Schär validation example.
  - `examples/two_dimension_mountain_wave.jl` — migrated to TFVD +
    `TwoLevelDecay` + auto-build (`slope_stencil = SlopeInsideInterpolation()`).

## Known follow-ups

  - Mountain-drag sign convention vs CM1 (diagnostic-tooling issue, not
    a substepper bug).
  - Bell strict-acceptance projection-amplitude on the small periodic-x
    domain at 7200 s (the wave packet wraps; framing issue shared with
    the explicit path).
  - Curvilinear-mesh plotting for vertical slices on TFVD grids (filed
    upstream as [CliMA/Oceananigans.jl#5651](https://github.com/CliMA/Oceananigans.jl/issues/5651)) — independent of this PR.

Part of NUM-151


giordano:
Is this related to #673?

glwagner:
Yes, it closes it.

test/terrain_following.jl
```diff
@@ -184,22 +196,658 @@ using Test
...
+            @test w̃[i, 1, Nz+1] ≈ w[i, 1, Nz+1] atol=1e-10
         end
     end
 
+    @testset "Zero terrain recovers Cartesian vertical transport" begin
+        Nx, Nz = 8, 6
+        Lx, Lz = 10000.0, 5000.0
+
+        z_faces = MutableVerticalDiscretization(collect(range(0, Lz, length=Nz+1)))
+        grid = RectilinearGrid(CPU(); size=(Nx, Nz),
```
giordano: Why this is on `CPU()` and not `default_arch`?  If it is intentional, and not an accident, it'd be good to add a comment at the top of the file to explain the rationale (I see in this entire file `CPU()` is used everywhere, so this may be intentional?).

glwagner: ill switch it over and see if tests pass!

glwagner: Good catch — it was inconsistent with the project's  convention (other test files and the testing-rules doc both use it). Switched all 20 instances to `default_arch` (commit pending), with a small `@isdefined(default_arch) || (default_arch = CPU())` fallback at the top so the file also works when included directly with `julia --project=. test/terrain_following.jl` during development. GPU CPU/GPU equivalence on the same terrain code path is exercised separately by `validation_output/substepper/terrain_gpu_smoke.jl` (1e-11 to 1e-7 CPU/GPU diff on an H100 smoke run earlier this session).

test/terrain_following.jl
```diff
@@ -184,22 +207,660 @@ using Test
...
+
+        set!(model, ρ=ρᵢ, θ=θ₀, u=10, w=1)
+        compute_contravariant_velocity!(model)
+
+        w̃ = model.dynamics.contravariant_vertical_velocity
+        ρw̃ = model.dynamics.contravariant_vertical_momentum
+
+        @test maximum(abs, interior(w̃) .- interior(model.velocities.w)) == 0
+        @test maximum(abs, interior(ρw̃) .- interior(model.momentum.ρw)) == 0
+        @test transport_velocities(model) === model.velocities
```
giordano: This is equivalent to saying they're equal, no?

test/terrain_following.jl
```diff
@@ -184,22 +207,660 @@ using Test
...
+            @test maximum(abs, interior(height_model.dynamics.density) .-
+                               interior(terrain_model.dynamics.density)) <= zero_terrain_tolerance
+            @test maximum(abs, interior(height_ρθ) .-
+                               interior(terrain_ρθ)) <= zero_terrain_tolerance
+            @test maximum(abs, interior(height_model.momentum.ρu) .-
+                               interior(terrain_model.momentum.ρu)) <= zero_terrain_tolerance
+            @test maximum(abs, interior(height_model.momentum.ρv) .-
+                               interior(terrain_model.momentum.ρv)) <= zero_terrain_tolerance
+            @test maximum(abs, interior(height_model.momentum.ρw) .-
+                               interior(terrain_model.momentum.ρw)) <= zero_terrain_tolerance
```
giordano: Use isapprox

test/terrain_following.jl
glwagner: yeah, this is testing a codepath wherein if the terrain is detected to be flat (even though we requested `terrain_metrics`) then we implement an optimization that saves allocating memory for `transport_velocities(model)`, which would otherwise have to be different

test/terrain_following.jl
glwagner: Done — replaced the five lines flagged here with `isapprox(...; atol=zero_terrain_tolerance)`, and also converted the related `maximum(abs, interior(w̃) .- interior(w)) == 0` pair at lines 253-254 to direct `interior(A) == interior(B)` (cleaner; that case is bit-exact equivalence via the flat-terrain fast path, not a tolerance check). Other `<=` patterns in the file are scalar magnitude bounds like `maximum(abs, interior(w)) <= 1e-12` (testing ≈0 of a single field) which I left alone since they're already in the standard idiom — let me know if you want those switched too.

giordano: Did the bot forget to actually push?

validation_output/substepper/complex_mountain_substepper_vs_explicit_field_timeseries_metrics.csv
giordano, line 1: Unrelated file (together with all the others in the directory)

giordano (commented):
There's also a conflict with `main` now, from #718

glwagner:
## Mountain-wave validation animations

Each video below is hosted as a release asset on tag `terrain-pr-712-media-20260601152453` and embedded
inline via HTML5 `<video>` so it plays in the PR view.

### 2D Schär vs CM1 — vertical-velocity field

https://github.com/user-attachments/assets/389b388a-7cb2-4c26-acda-9241175baf61

### 3D bell mountain wave (Smith 1980 wave cone)

https://github.com/user-attachments/assets/cbbb13be-d711-4b6e-85ce-4ecd9a58fccc

### 2D canyon — two-peak interference (Fr = 1.5)

https://github.com/user-attachments/assets/69b4df0f-9b39-4635-aeb3-5a474d1a5192

### 2D Sierra-Nevada cross-section (Fr = 3.9, multi-scale terrain)

https://github.com/user-attachments/assets/be6928d3-e7a7-4e1d-970a-cc5083705723

https://github.com/user-attachments/assets/2476346e-5246-4c86-9924-d31c91eb95c7

### 2D trapped lee waves (two-layer N(z), Schär hill)

https://github.com/user-attachments/assets/d41d8d25-7251-456f-9852-ab54d9bdf25a

test/terrain_following.jl
glwagner: not sure what happened. i'm fixing it now

docs/src/terrain_following_coordinates.md
```diff
@@ -1,212 +1,959 @@
 # [Terrain-following coordinates](@id Terrain-following-section)
 
-Terrain-following coordinates map the irregular physical domain above topography onto a
-regular computational domain. This avoids the need for immersed boundaries or cut cells
-at the surface, simplifying the application of boundary conditions and ensuring smooth
-coordinate surfaces near the ground.
+Terrain-following coordinates map the irregular physical domain above topography
+onto a regular computational domain. The horizontal grid stays uniform; only the
```
ewquon:
```suggestion
onto a regular computational domain. The horizontal grid stays uniform — a design
decision that simplifies the coordinate Jacobian (with ∂ξ/∂x = ∂η/∂y = const) to the
vertical stretching factor σ = ∂z/∂ζ. Only the
```

docs/src/terrain_following_coordinates.md
```diff
@@ -1,212 +1,959 @@
...
-The implementation is built on Oceananigans' `MutableVerticalDiscretization`,
-which provides the column-wise scaling factors ``\sigma(x, y)`` and surface displacement
-``\eta(x, y)`` needed to deform the vertical coordinate. This page describes the
-coordinate transformation, the metric corrections required for the dynamical equations,
-and the user-facing API.
+The price is that derivatives in the computational frame are no longer Cartesian
+derivatives, and the discrete equations have to carry metric correction terms.
+Most of this page is about getting those corrections right — and about an
+initial-condition gotcha that bites every user the first time they set up a
+stratified resting atmosphere over terrain.
```
ewquon:
```suggestion
!!! note "The price"
    Derivatives in the computational frame are no longer Cartesian
    derivatives, and the discrete equations have to carry metric correction terms.
    Most of this page is about getting those corrections right — and about an
    [initial-condition gotcha](#pitfalls-the-ic-gotcha-that-bites-everyone)
    that bites every user the first time they set up a
    stratified resting atmosphere over terrain.
```
Should we more clearly describe the `z` vs `ζ` gotcha?

docs/src/terrain_following_coordinates.md
```diff
@@ -1,212 +1,959 @@
...
+  - The Crank–Nicolson forward weight ``\omega \in [0.5, 1]`` off-centres the
+    implicit vertical solve; the default ``\omega = 0.65`` adds modest
+    damping over centred CN.
+
+The split-explicit machinery is the same as in flat-terrain models; what
+changes is *which* fields go into the kernels (contravariant vs Cartesian
+momentum) and *which* references are subtracted (3D ``p_\text{ref}, \rho_\text{ref}``
+vs 1D background).
+
+## Pitfalls: altitude-dependent initial conditions
```
ewquon:
```suggestion
## [Pitfalls: altitude-dependent initial conditions](#pitfalls-the-ic-gotcha-that-bites-everyone)
```

docs/src/terrain_following_coordinates.md
```diff
@@ -1,212 +1,959 @@
...
+The two patterns to copy verbatim into your own scripts are:
 
-- `transport_velocities(model)`: Returns `(u, v, w)` for standard models or
-  `(u, v, Ω̃)` for terrain-following models.
-- `advecting_momentum(model)`: Returns `(ρu, ρv, ρw)` for z-coordinate models or
-  `(ρu, ρv, ρΩ̃)` for terrain-following models, where `ρΩ̃` is the contravariant
-  vertical momentum.
+  1. `set!(model, ρ = model.dynamics.terrain_reference_density,
+     θ = θ_profile, ...)` — the hydrostatic thermal IC path.
+  2. `_init_terrain_bottom_face_w!` — the kinematic BC, applied once at IC so
```
ewquon: Should `_init_terrain_bottom_face_w!` be user tunable?

If not, why don't we include it in Breeze? Implementing `set!(...; enforce_terrain_kinematic_balance=true)` would be nice, and symmetric with the existing `enforce_mass_conservation` kwarg.

examples/two_dimension_mountain_wave.jl
```diff
@@ -80,78 +81,95 @@
...
-# (set `N² = 1e-4` to match that paper exactly).
+# Note: the standard [Schar2002](@citet) parameters use ``buoyancy_frequency = 0.01 \, {\rm s}^{-1}``
+# (set `buoyancy_frequency_squared = 1e-4` to match that paper exactly).
 
-N² = g^2 / (cᵖᵈ * T₀)       # s⁻² - Brunt–Väisälä frequency squared (≈ 3.2e-4)
-N  = sqrt(N²)               # s⁻¹ - Brunt–Väisälä frequency
-β  = g / (Rᵈ * T₀)          # m⁻¹ - density scale parameter
+buoyancy_frequency_squared = g^2 / (dry_air_heat_capacity * reference_temperature) # s^-2 - Brunt-Vaisala frequency squared
+buoyancy_frequency = sqrt(buoyancy_frequency_squared)                              # s^-1 - Brunt-Vaisala frequency
+density_scale = g / (dry_air_gas_constant_value * reference_temperature)           # m^-1 - density scale parameter
```
ewquon: Any reason we dropped the symbols in favor of explicit variable names? Seems to deviate from our convention

docs/src/terrain_following_coordinates.md
glwagner: Hmm yes, or even better solve this. Let me take a look.

glwagner: I think this is out of date. `set!` should use altitude (`z`), not the computational coordinate. Fixing this.

src/TerrainFollowingDiscretization/terrain_following_vertical_discretization.jl
```diff
@@ -0,0 +1,186 @@
...
+##### σⁿ = ∂z/∂ζ (Jacobian) and znode = z(x,y,ζ) are computed from the
+##### formulation's terrain components + decay profiles. Δz spacings follow as
+##### Δr · σⁿ. Defining σⁿ over (i,j,k) is what makes the Jacobian k-dependent
+##### for TwoLevelDecay without any 3-D σ storage.
+#####
+
+# Match the terrain-following coordinate at the vertical-coordinate slot of the
+# concrete grid types (position 5, as Oceananigans' own `MRG`/`MLLG` aliases do).
+const TFVDRG = Union{RectilinearGrid{<:Any, <:Any, <:Any, <:Any, <:TFVD},
+                     LatitudeLongitudeGrid{<:Any, <:Any, <:Any, <:Any, <:TFVD}}
```
ewquon: The "RG" in this doozy of an acronym is throwing me off — and it doesn't exactly mirror `MRG`/`MLLG` (otherwise it would be `TFVDRG`/`TFVDLLG`, with a separate name for the union).

Would this work?
```suggestion
const TerrainFollowingGrid = AbstractUnderlyingGrid{<:Any, <:Any, <:Any, <:Bounded, <:TFVD}
```

examples/two_dimension_mountain_wave.jl
glwagner: yeah I don't like this either, sorry! I'll fix it.

src/CompressibleEquations/terrain_compressible_physics.jl
```diff
@@ -127,19 +122,277 @@ end
...
+    ζ = rnode(k, grid, Face())
+    return ∂x_h_cc * (1 - ζ / metrics.z_top)
+end
+
+@inline function terrain_slope_y_ccf(i, j, k, grid, metrics)
+    ∂y_h_cc = ℑyᵃᶜᵃ(i, j, 1, grid, metrics.∂y_h)
+    ζ = rnode(k, grid, Face())
+    return ∂y_h_cc * (1 - ζ / metrics.z_top)
+end
+
```
ewquon:
```suggestion
```
This fallback is equivalent to `LinearDecay` and is currently unreachable

src/CompressibleEquations/terrain_compressible_physics.jl
```diff
@@ -127,19 +122,277 @@ end
...
+# On a TerrainFollowingVerticalDiscretization grid the coordinate owns the
+# slope: take (∂z/∂x)_ζ from the grid operator (which carries the formulation's
+# decay — linear for LinearDecay, sinh for TwoLevelDecay) and interpolate the x-face
+# value to (Center, Center) at the z-face. The `metrics` argument is ignored;
```
ewquon: Should we remove the `metrics` arg then?

src/CompressibleEquations/terrain_compressible_physics.jl
```diff
@@ -127,19 +122,277 @@ end
...
+    @inbounds ρw_ccf = ρw[i, j, k]
+
+    return ρw_ccf - slope_x * ρu_ccf - slope_y * ρv_ccf
+end
+
+@inline function acoustic_vertical_momentum_flux(i, j, k, grid,
+                                                 dynamics::TerrainCompressibleDynamics,
+                                                 ρu′, ρv′, ρw̃′)
+    @inbounds return ρw̃′[i, j, k]
+end
```
ewquon: Is this definition and the function signature necessary? This is also defined generally in src/CompressibleEquations/acoustic_substepping.jl as 
```julia
@inline acoustic_vertical_momentum_flux(i, j, k, grid, dynamics, ρu′, ρv′, ρw′) =
    @inbounds ρw′[i, j, k]
```

kaiyuan-cheng (approved):
Implementation look good!

src/CompressibleEquations/terrain_compressible_physics.jl
glwagner: Agreed — that generic fallback was the old MVD/`follow_terrain!` path (linear decay from `metrics.∂x_h`), unreachable since every terrain model carries `TerrainMetrics` built only from a TFVD grid. Removed in 14a1d37.

src/CompressibleEquations/terrain_compressible_physics.jl
glwagner: Yes — done in 14a1d37. The slope comes entirely from the grid `∂z∂x`/`∂z∂y` operator now, so I dropped `metrics` from `terrain_slope_{x,y}_ccf` and followed the chain: every helper that only threaded `metrics` down to the slope (`terrain_vertical_transport_momentum`, `slope_{x,y}_times_∂z[_p′]`, `slope_{x,y}_times_∂z_linearized_pressure`, `_compute_contravariant_velocity!`, `_initialize_terrain_vertical_momentum_perturbation!`) lost the arg too, along with their call sites and kernel launches. Genuine `terrain_metrics.pressure_gradient_stencil` uses are untouched.

src/CompressibleEquations/terrain_compressible_physics.jl
glwagner: Right — it was byte-for-byte identical to the generic in acoustic_substepping.jl (both return `ρw′[i,j,k]`), and the generic already dispatches on `TerrainCompressibleDynamics`. Removed in 14a1d37.

PR #715, opened 2026-05-18
[CI] Clean up old CUDA toolkits on GPU runners

giordano:
This saves some space on the VMs.


giordano:
Let's see how this goes.

giordano:
[That didn't help much](https://github.com/NumericalEarth/Breeze.jl/actions/runs/26107741466/job/76776192598?pr=720#step:2:61), sigh.  I suspect the Docker images are pulled into a different partition than where the CUDA toolkits are in.

PR #718, opened 2026-05-18
Support backward integration in split-explicit acoustic substepping

kaiyuan-cheng:
## Summary

- Use `|Δt|` for the adaptive acoustic substep count in `compute_acoustic_substeps`, so the count stays positive when `Δt < 0`.
- Use `|δτ|` in the `UpperSponge` LHS-diagonal and RHS contributions so the sponge stays dissipative regardless of integration direction (one-sided regularizer; backward integration through a sponge layer is stable but not an exact inverse of the forward step inside the sponge).
- Document backward integration support on `SplitExplicitTimeDiscretization`: the A-stability argument for the off-centered CN vertical solve (``|A|^2 \le 1`` for either sign of ``Δτ``), the sign-self-consistency of horizontal divergence damping, and the sponge irreversibility caveat.

No behavior change for `Δt > 0`.

## Test plan

- [ ] Forward-only regression: existing CompressibleEquations tests still pass (substep count, sponge tridiag).
- [ ] Backward step smoke test: integrate a smooth state forward then backward with `Δt → -Δt` and confirm acoustic substep loop runs without NaNs and recovers the initial state outside any sponge layer to acoustic-truncation accuracy.
- [ ] Sponge-on backward test: confirm stability (no blow-up) when integrating backward through a configured `UpperSponge`, and that the in-sponge state is damped (not amplified).
- [ ] Doctests / docs build pass with the new docstring sections.

### DIFF
--- a/test/acoustic_substepping.jl
+++ b/test/acoustic_substepping.jl
@@ -256,6 +256,15 @@ for arch in arches
             @test N_loose  == ceil(Int, 12 * sqrt(1.4 * 287.0 * 300) / (1.0  * 1000))
             @test N_strict > N_default > N_loose
         end
+
+        @testset "Backward Δt yields same substep count" begin
+            grid = RectilinearGrid(arch; size=(100, 6, 10), halo=(5, 5, 5),
+                                   x=(0, 100kilometers), y=(0, 6kilometers), z=(0, 10kilometers))
+            N_fwd = compute_acoustic_substeps(grid, +12, constants, ν)
+            N_bwd = compute_acoustic_substeps(grid, -12, constants, ν)
+            @test N_fwd ≥ 1
+            @test N_bwd == N_fwd
+        end
     end
 
     @testset "acoustic_cfl plumbed to AcousticSubstepper [$(arch), $(FT)]" for FT in as_test_float_types(arch)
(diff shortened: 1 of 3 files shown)


giordano:
Do we need a simple test to make sure the use case works as intended?

glwagner (approved):
interesting 👀

glwagner:
> Do we need a simple test to make sure the use case works as intended?

I believe @kaiyuan-cheng is working on an application that will provide the ability to exercise and test this feature. However, it may be best to merge this earlier rather than later to reduce conflict with #712 ?

glwagner:
> > Do we need a simple test to make sure the use case works as intended?
> 
> I believe @kaiyuan-cheng is working on an application that will provide the ability to exercise and test this feature. However, it may be best to merge this earlier rather than later to reduce conflict with #712 ?

That said it could be pretty easy to just take one step forward and back and concoct some random notion of "error". That will at least test that it doesn't blow up.

giordano:
> # ⚠️ **Performance Alert** ⚠️


I suspect we are back to the situation "this is so fast that a little jitter is actually significant" for the default configuration with the smallest grid. Since the latest performance improvements in Oceananigans that's gone up and down quite a few times: https://numericalearth.github.io/BreezeBenchmarks/#benchmark-cbl--dynamics--anelastic--microphysics--nothing--float32-

kaiyuan-cheng:
This has been tested, and it appears to work correctly. Should we merge it?
image

glwagner:
should we add the test to CI?

test/acoustic_substepping.jl
```diff
@@ -339,6 +348,62 @@ for arch in arches
...
+        end
+
+        # "Random notion of error" — generous bounds; the round-trip is
+        # not exact but must stay bounded for the small perturbation.
+        ρ_final  = Array(parent(model.dynamics.density))
+        ρu_final = Array(parent(model.momentum.ρu))
+        ρw_final = Array(parent(model.momentum.ρw))
+        @test maximum(abs, ρ_final  .- ρ_init)  < 1
+        @test maximum(abs, ρu_final .- ρu_init) < 1
+        @test maximum(abs, ρw_final .- ρw_init) < 1
```
glwagner: 😂 my own words! maybe a `reltol` thats like 1e-2 would work?

giordano: Yeah, this should use `isapprox`

glwagner: i guess from the figure, 1e-4 might work too

PR #720, opened 2026-05-19
[CI] Run default anelastic benchmarks for longer

giordano:
I'm wondering if this will help with https://github.com/NumericalEarth/Breeze.jl/pull/718#issuecomment-4482873293 (but it'll be hard to tell with a single run)


giordano:
I believe that running more timesteps does achieve the goal of stabilising the performance of these benchmarks, but then this job takes proportionally longer.  I'll play with numbers a little bit to find a compromise between "don't take way too long" and "don't finish too quickly so that we have unstable measurements"

giordano:
Let's see how this goes.

Issue #717, opened 2026-05-18
`compute_velocities!` discards `clock` and `model_fields` when refilling momentum halos, breaking continuous OBC callables


ewquon:
## Summary

`compute_auxiliary_variables!` (in `AtmosphereModels/update_atmosphere_model_state.jl:99-103`) refills the momentum halos *without* passing `model.clock` or `fields(model)`:

```julia
density = dynamics_density(model.dynamics)
fill_halo_regions!(density)
fill_halo_regions!(model.momentum)
```

Compare to `update_state!` in the same file (line 16), which threads them through:

```julia
fill_halo_regions!(prognostic_fields(model), model.clock, fields(model), async=true)
```

For a continuous-callable Open BC (`OpenBoundaryCondition(my_func; parameters = ...)`) on any momentum component, Oceananigans dispatches `getbc` to `getbc(cbf::XBoundaryFunction, j, k, grid, clock, model_fields, args...)`, which requires `clock` and `model_fields`. Without them, dispatch falls through to the generic `getbc(condition, args...) = condition(args...)`, which tries to invoke the `ContinuousBoundaryFunction` directly with `(j, k, grid)` and fails:

```
MethodError: objects of type ContinuousBoundaryFunction{BoundaryAdjacent, Center, Center, LeftBoundary, typeof(ρu_west), @NamedTuple{ρ::Float64, ω::Float64}, Tuple{}, Tuple{}, Tuple{}} are not callable
```

This blocks any time-varying Open BC expressed as a continuous callable on momentum.

## Minimal reproducer

```julia
using Breeze, Oceananigans

grid = RectilinearGrid(size=(16,16,4), x=(0,1000), y=(0,1000), z=(0,200),
                       topology=(Bounded, Bounded, Bounded))
constants = ThermodynamicConstants(eltype(grid))
reference_state = ReferenceState(grid, constants;
                                 surface_pressure=101325.0, potential_temperature=300.0)
dynamics = CompressibleDynamics(SplitExplicitTimeDiscretization();
                                surface_pressure=101325.0,
                                reference_potential_temperature=300.0)

@inline ρu_west(y, z, t, p) = p.ρ * cos(p.ω * t)

ρu_bcs = FieldBoundaryConditions(
    west = OpenBoundaryCondition(ρu_west; parameters = (; ρ = 1.17, ω = 0.01)))

model = AtmosphereModel(grid; dynamics, thermodynamic_constants=constants,
[...]

PR #722, opened 2026-05-20
Propagate momentum open boundary conditions to derived velocities

### LINKED ISSUE #721: `OpenBoundaryCondition` on momentum is silently clobbered by impenetrable-default on derived velocities → NaN
## Summary

When a user sets an `OpenBoundaryCondition` on a momentum component (`:ρu`, `:ρv`, `:ρw`) of an `AtmosphereModel`, the prescribed boundary-face value is silently overwritten by zero in the derived `model.velocities`. The resulting inconsistency between `ρu` (= user value at the wall) and `u` (= 0 at the wall) drives a wall-mode instability that NaNs the run within \~100 iterations on both `AnelasticDynamics` and `CompressibleDynamics` (with `ExplicitTimeStepping`), independent of CFL, IC, or Davies relaxation. Distinct from NumericalEarth/Breeze.jl#716, NumericalEarth/Breeze.jl#717, NumericalEarth/Breeze.jl#719.

## Root cause

`materialize_momentum_and_velocities` (anelastic + compressible) constructs `model.velocities.u/v/w` with `FieldBoundaryConditions()` defaults — these resolve to `ImpenetrableBoundaryCondition` on `Face` sides of `Bounded` topology, regardless of what the user set on `:ρu/ρv/ρw`. After `compute_velocities!` correctly fills `u = ρu/ρ` at boundary faces, the subsequent `fill_halo_regions!(model.velocities)` writes those faces back to 0.

Secondary: `compute_velocities!` launches over `:xyz` (worksize `(Nx,Ny,Nz)`), which excludes the east/north/top boundary faces (i=Nx+1, j=Ny+1, k=Nz+1) entirely — those faces are never written by the kernel and end up holding whatever `fill_halo_regions!` decides (zero, by default).

## MWE (compressible, CPU)

```julia
using Oceananigans, Breeze
using Oceananigans.BoundaryConditions: fill_halo_regions!
using Breeze.CompressibleEquations: CompressibleDynamics, ExplicitTimeStepping
using Breeze.AtmosphereModels: compute_velocities!

grid = RectilinearGrid(CPU(), Float64;
                       size = (16, 16, 4), x = (0, 1000), y = (0, 1000), z = (0, 100),
                       topology = (Bounded, Periodic, Bounded))

dynamics = CompressibleDynamics(ExplicitTimeStepping(); reference_potential_temperature=300.0)
ρu_value = 6.0
ρu_bcs = FieldBoundaryConditions(west = OpenBoundaryCondition(ρu_value),
                                  east = OpenBoundaryCondition(ρu_value))
model = AtmosphereModel(grid; dynamics, formulation=:LiquidIcePotentialTemperature,
                              boundary_conditions=(; ρu=ρu_bcs))

ref = model.dynamics.reference_state
set!(model; θ=300.0, ρ=ref.density, ρu=(x,y,z) -> ρu_value, ρv=0, ρw=0)
fill_halo_regions!(model.momentum, model.clock, fields(model))
compute_velocities!(model)

# u at the boundary face should equal ρu / ρ ≈ 6.0 / 1.171 ≈ 5.12
@show model.momentum.ρu[1, 1, 1]     # 6.0  ✅
@show model.velocities.u[1, 1, 1]    # 0.0  ❌  (expected ≈ 5.12)
@show model.momentum.ρu[17, 1, 1]    # 6.0  ✅
@show model.velocities.u[17, 1, 1]   # 0.0  ❌  (expected ≈ 5.12)
```

West- and east-face velocities both clobbered to 0 despite `ρu` carrying the user's OBC value. Replacing `CompressibleDynamics(ExplicitTimeStepping(); ...)` + `set!(...; ρ=ref.density, ...)` with `AnelasticDynamics(ReferenceState(grid; surface_pressure=101325, potential_temperature=300))` reproduces the same clobber.

## Symptom in user simulations

* Anelastic + OBC + Davies + vortex → NaN
* Anelastic + OBC + Davies + uniform U_bg → NaN
* Compressible + OBC + Davies + isentropic vortex, CFL=0.05 → NaN
* Anelastic + Davies-only (no OBC) + vortex → numerically stable but nonphysical

## Fix

PR forthcoming. Two pieces:

1. Derive velocity BCs from momentum BCs in `materialize_momentum_and_velocities` (Open momentum → `nothing` velocity, so `fill_halo_regions!` doesn't clobber; pass through Periodic/Flux/Value/etc.).
2. Extend `compute_velocities!` to cover the full Face range via `KernelParameters` instead of `:xyz`, including east/north/top boundary faces.

## Environment

Breeze 0.5.0 main (`a617eaf`), Oceananigans 0.107.6, Julia 1.11+, both CPU and CUDA reproduce.

ewquon:
## Summary

Two-part fix:
1. Override velocity BCs to `nothing` on the normal-component face of each `Bounded` direction in `materialize_momentum_and_velocities` (anelastic + compressible), so `fill_halo_regions!(model.velocities)` no longer clobbers the kernel-computed boundary face. Default impenetrable case still works because `OpenBoundaryCondition(0)` on momentum yields `u = ρu/ρ = 0` at the wall.
2. Launch `compute_velocities!` over the full Face range via `KernelParameters`, so east/north/top boundary faces are written each step rather than left stale.

Fixes #721.

## Test plan

New `test/open_boundary_momentum.jl`:

- [x] Anelastic + Open `ρu` ⇒ `u.bcs.{west,east}` is `nothing`; `u` at boundary face ≈ `U_bg` after `compute_velocities!`.
- [x] Compressible — same.
- [x] Anelastic OBC inflow — 20 SSP-RK3 steps stay finite (pre-fix NaN'd within ~100 iters).
- [x] Default impenetrable wall — `u = 0` confirmed, no special-case needed.
- [x] 15/15 pass locally on CPU/Float64.
- [x] CI for GPU/Float32.

### DIFF
--- a/src/AnelasticEquations/anelastic_dynamics.jl
+++ b/src/AnelasticEquations/anelastic_dynamics.jl
@@ -158,8 +158,14 @@ function AtmosphereModels.materialize_momentum_and_velocities(dynamics::Anelasti
     ρw = ZFaceField(grid, boundary_conditions=boundary_conditions.ρw)
     momentum = (; ρu, ρv, ρw)
 
-    velocity_bcs = NamedTuple(name => FieldBoundaryConditions() for name in (:u, :v, :w))
-    velocity_bcs = regularize_field_boundary_conditions(velocity_bcs, grid, (:u, :v, :w))
+    # Velocity is diagnostic (u = ρu/ρ via compute_velocities!). On the normal-component
+    # face in a `Bounded` direction we use `nothing` so `fill_halo_regions!(velocities)`
+    # does not overwrite the kernel-computed boundary face — momentum carries the wall BC.
+    TX, TY, TZ = topology(grid)
+    u_bcs = TX === Bounded ? FieldBoundaryConditions(west=nothing,   east=nothing)  : FieldBoundaryConditions()
+    v_bcs = TY === Bounded ? FieldBoundaryConditions(south=nothing,  north=nothing) : FieldBoundaryConditions()
+    w_bcs = TZ === Bounded ? FieldBoundaryConditions(bottom=nothing, top=nothing)   : FieldBoundaryConditions()
+    velocity_bcs = regularize_field_boundary_conditions((u=u_bcs, v=v_bcs, w=w_bcs), grid, (:u, :v, :w))
     u = XFaceField(grid, boundary_conditions=velocity_bcs.u)
     v = YFaceField(grid, boundary_conditions=velocity_bcs.v)
     w = ZFaceField(grid, boundary_conditions=velocity_bcs.w)
--- a/src/AnelasticEquations/anelastic_dynamics.jl
+++ b/src/AnelasticEquations/anelastic_dynamics.jl
@@ -158,17 +158,14 @@ function AtmosphereModels.materialize_momentum_and_velocities(dynamics::Anelasti
     ρw = ZFaceField(grid, boundary_conditions=boundary_conditions.ρw)
     momentum = (; ρu, ρv, ρw)
 
-    # Velocity is diagnostic (u = ρu/ρ via compute_velocities!). On the normal-component
-    # face in a `Bounded` direction we use `nothing` so `fill_halo_regions!(velocities)`
-    # does not overwrite the kernel-computed boundary face — momentum carries the wall BC.
-    TX, TY, TZ = topology(grid)
-    u_bcs = TX === Bounded ? FieldBoundaryConditions(west=nothing,   east=nothing)  : FieldBoundaryConditions()
-    v_bcs = TY === Bounded ? FieldBoundaryConditions(south=nothing,  north=nothing) : FieldBoundaryConditions()
-    w_bcs = TZ === Bounded ? FieldBoundaryConditions(bottom=nothing, top=nothing)   : FieldBoundaryConditions()
-    velocity_bcs = regularize_field_boundary_conditions((u=u_bcs, v=v_bcs, w=w_bcs), grid, (:u, :v, :w))
-    u = XFaceField(grid, boundary_conditions=velocity_bcs.u)
-    v = YFaceField(grid, boundary_conditions=velocity_bcs.v)
-    w = ZFaceField(grid, boundary_conditions=velocity_bcs.w)
+    # Velocity is diagnostic (u = ρu/ρ via compute_velocities!). Use the auxiliary-field
+    # default BCs (`nothing` on Bounded-Face sides, Periodic on Periodic sides), which
+    # is what XFaceField gives us when constructed with no `boundary_conditions=` kwarg.
+    # `nothing` on Bounded-Face prevents `fill_halo_regions!(velocities)` from clobbering
+    # the kernel-computed boundary face — momentum carries the wall BC.
+    u = XFaceField(grid)
+    v = YFaceField(grid)
+    w = ZFaceField(grid)
     velocities = (; u, v, w)
 
     return momentum, velocities
--- a/src/AtmosphereModels/update_atmosphere_model_state.jl
+++ b/src/AtmosphereModels/update_atmosphere_model_state.jl
@@ -2,7 +2,8 @@ using ..Thermodynamics: Thermodynamics, mixture_gas_constant
 
 using Oceananigans.BoundaryConditions: fill_halo_regions!, compute_x_bcs!, compute_y_bcs!, compute_z_bcs!,
                                        update_boundary_conditions!
-using Oceananigans.Grids: Flat, topology
+using Oceananigans: Face
+using Oceananigans.Grids: topology
 using Oceananigans.ImmersedBoundaries: mask_immersed_field!
 using Oceananigans.TimeSteppers: TimeSteppers
 using Oceananigans.TurbulenceClosures: compute_closure_fields!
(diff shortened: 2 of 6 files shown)


src/AnelasticEquations/anelastic_dynamics.jl
```diff
@@ -158,8 +158,14 @@ function AtmosphereModels.materialize_momentum_and_velocities(dynamics::Anelasti
...
+    # face in a `Bounded` direction we use `nothing` so `fill_halo_regions!(velocities)`
+    # does not overwrite the kernel-computed boundary face — momentum carries the wall BC.
+    TX, TY, TZ = topology(grid)
+    u_bcs = TX === Bounded ? FieldBoundaryConditions(west=nothing,   east=nothing)  : FieldBoundaryConditions()
+    v_bcs = TY === Bounded ? FieldBoundaryConditions(south=nothing,  north=nothing) : FieldBoundaryConditions()
+    w_bcs = TZ === Bounded ? FieldBoundaryConditions(bottom=nothing, top=nothing)   : FieldBoundaryConditions()
+    velocity_bcs = regularize_field_boundary_conditions((u=u_bcs, v=v_bcs, w=w_bcs), grid, (:u, :v, :w))
     u = XFaceField(grid, boundary_conditions=velocity_bcs.u)
     v = YFaceField(grid, boundary_conditions=velocity_bcs.v)
     w = ZFaceField(grid, boundary_conditions=velocity_bcs.w)
```
glwagner:
```suggestion
u = XFaceField(grid)
v = YFaceField(grid)
w = ZFaceField(grid)
```

this is a bit simpler

the issue with above is calling `regularize_field_boundary_conditions` which causes the bcs to be interpreted as prognostic bcs.

The motivation for calling `regularize_field_boundary_conditions` is in fact to build the field with impenetrable bcs. So now that we are not doing that, we should not call `regularize_field_boundary_conditions`.

src/AtmosphereModels/update_atmosphere_model_state.jl
```diff
@@ -82,32 +82,27 @@ function compute_velocities!(model::AtmosphereModel)
...
-            model.dynamics,
-            model.momentum)
+    # Extend the launch to include boundary faces (i=Nx+1, j=Ny+1, k=Nz+1) in
+    # `Bounded` directions so `model.velocities` stays consistent with
+    # `model.momentum` at the wall. See `velocity_boundary_conditions_from_momentum`.
+    Nx, Ny, Nz = size(grid)
+    TX, TY, TZ = topology(grid)
+    ix_u = TX === Bounded ? (1:Nx+1) : (1:Nx)
+    jy_v = TY === Bounded ? (1:Ny+1) : (1:Ny)
+    kz_w = TZ === Bounded ? (1:Nz+1) : (1:Nz)
```
glwagner: Use dispatch?

In general, try to minimize `if`-statements

src/AtmosphereModels/update_atmosphere_model_state.jl
```diff
@@ -82,32 +82,27 @@ function compute_velocities!(model::AtmosphereModel)
...
+
+    launch!(arch, grid, KernelParameters(ix_u, 1:Ny, 1:Nz),
+            _compute_u!, model.velocities.u, grid, model.dynamics, model.momentum.ρu)
+    launch!(arch, grid, KernelParameters(1:Nx, jy_v, 1:Nz),
+            _compute_v!, model.velocities.v, grid, model.dynamics, model.momentum.ρv)
+    launch!(arch, grid, KernelParameters(1:Nx, 1:Ny, kz_w),
+            _compute_w!, model.velocities.w, grid, model.dynamics, model.momentum.ρw)
```
glwagner: is it possible to keep one kernel, but using some interior kernel logic?

I also suggest always launching from 1:N+1 on all sides if possible!

ewquon: Oops, forgot to ask Claude to fuse the kernels

glwagner:
Hit a related issue while wiring this PR into NumericalEarth's [`NestedSimulations`](https://github.com/NumericalEarth/NumericalEarth.jl/tree/eq/era5_breeze_land/src/EarthSystemModels/NestedSimulations) abstraction (a `PrescribedAtmosphere` parent driving a Breeze child via `OpenBoundaryCondition`s on `ρu`, `ρv`).

With **constant-valued** OBCs (`OpenBoundaryCondition(ρ_b * U_bg)`) the PR works as advertised — `u = ρu/ρ` at the boundary is correct after `compute_velocities!`.

With **function-based** OBCs (i.e. `ContinuousBoundaryFunction`-wrapped) the first time step errors with:

```
MethodError: objects of type Oceananigans.BoundaryConditions.ContinuousBoundaryFunction{Face, BoundaryAdjacent, Center, LeftBoundary, …} are not callable
```

Source: in [`compute_velocities!`](https://github.com/NumericalEarth/Breeze.jl/blob/eq/atmosphere-model-open-bc-velocity/src/AtmosphereModels/update_atmosphere_model_state.jl#L80-L82), the momentum halo fill is called without `model.clock, fields(model)`:

```julia
fill_halo_regions!(density)
fill_halo_regions!(model.momentum)
```

When the OBC's condition is a `ContinuousBoundaryFunction`, the specialized `getbc(::XBoundaryFunction, j, k, grid, clock, model_fields, args...)` method needs those extra args. Without them, dispatch falls through to `getbc(condition, args...) = condition(args...)` and Oceananigans tries to call the `ContinuousBoundaryFunction` wrapper directly, which has no `(::CBF)(...)` method.

Suggested fix: thread `model.clock, fields(model)` through to those two `fill_halo_regions!` calls (and probably the trailing `fill_halo_regions!(model.velocities)` too), mirroring how the main `update_state!` halo fill on `prognostic_fields(model)` does it.

Happy to open a follow-up PR if useful.

ewquon:
Yes that issue was already in my queue: https://github.com/NumericalEarth/Breeze.jl/issues/717

src/AtmosphereModels/update_atmosphere_model_state.jl
```diff
@@ -82,32 +76,25 @@ function compute_velocities!(model::AtmosphereModel)
...
-    launch!(arch, grid, :xyz,
+    # Launch over `(1:Nx+1, 1:Ny+1, 1:Nz+1)` for `Bounded`/`Periodic` dims — the N+1
+    # face is the boundary face (Bounded) or a halo cell (Periodic, refilled by the
+    # trailing `fill_halo_regions!(model.velocities)`). `Flat` dims have no Face and
+    # no halo, so stay at `1:N` to avoid out-of-bounds writes.
+    Nx, Ny, Nz = size(grid)
+    TX, TY, TZ = topology(grid)
+    launch!(arch, grid, KernelParameters(1:face_extent(TX, Nx),
+                                         1:face_extent(TY, Ny),
+                                         1:face_extent(TZ, Nz)),
```
glwagner: sorry to be so nitpicky here. I think there is a function in `Oceananigans.Grids` that does this. Something like `length(Face(), TX(), Nx)` etc. 

https://github.com/CliMA/Oceananigans.jl/blob/75133d8b9c6f875b4bd75749cc3dbfc2e746efc6/src/Grids/grid_utils.jl#L54

its ok to define `face_extent` though, more local... maybe a matter of taste

glwagner (approved):

src/AtmosphereModels/update_atmosphere_model_state.jl
ewquon: Np, good to know about this function

PR #723, opened 2026-05-20
Minor notational improvement in acoustic wave example

glwagner:
We may want to use `_i` for "initial" and `_0` for "z=0" as well.

### DIFF
--- a/examples/acoustic_wave.jl
+++ b/examples/acoustic_wave.jl
@@ -273,13 +273,12 @@ set!(ρᵇᵍ, (x, z) -> adiabatic_hydrostatic_density(z, p₀, θ₀, pˢᵗ, c
 set!(ρ_total, ρᵢ)
 
 # The initial wind field is the quantity we differentiate with respect to.
-# Enzyme accumulates ``∂J / ∂u_i`` into the shadow buffer
-# ``du_0``.
+# Enzyme accumulates ``∂J / ∂u_i`` into the shadow buffer ``du_i``.
 
-u₀  = XFaceField(grid_ad)
-du₀ = XFaceField(grid_ad)
-set!(u₀, (x, z) -> Uᵢ(z))
-set!(du₀, 0)
+uᵢ  = XFaceField(grid_ad)
+duᵢ = XFaceField(grid_ad)
+set!(uᵢ, (x, z) -> Uᵢ(z))
+set!(duᵢ, 0)
 
 # The shadow model stores accumulated adjoints for every prognostic field.
 
@@ -314,8 +313,8 @@ Nsteps = (isqrt(Nt - 1) + 1)^2
 # wind field on every backward evaluation.  Without it, AD would differentiate
 # a stale trajectory.
 
-function loss(model, u₀, ρ_total, ρᵇᵍ, θ₀, Δt, nsteps)
-    set!(model; ρ = ρ_total, θ = θ₀, u = u₀)
+function loss(model, uᵢ, ρ_total, ρᵇᵍ, θ₀, Δt, nsteps)
+    set!(model; ρ = ρ_total, θ = θ₀, u = uᵢ)
     @trace mincut=true checkpointing=true track_numbers=false for _ in 1:nsteps
         time_step!(model, Δt)
     end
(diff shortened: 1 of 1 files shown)


dkytezab:
Also we should rename title of the plot so its not '169 delta t' but the actual time value

navidcy (approved):

examples/acoustic_wave.jl
```diff
@@ -271,10 +271,10 @@ model_ad = AtmosphereModel(grid_ad; dynamics = CompressibleDynamics(ExplicitTime
 set!(ρᵇᵍ, (x, z) -> adiabatic_hydrostatic_density(z, p₀, θ₀, pˢᵗ, constants))
 
 ρ_total = CenterField(grid_ad)
-set!(ρ_total, ρ_init)
+set!(ρ_total, ρᵢ)
 
 # The initial wind field is the quantity we differentiate with respect to.
-# Enzyme accumulates ``\partial J / \partial u_0`` into the shadow buffer
+# Enzyme accumulates ``∂J / ∂u_0`` into the shadow buffer
```
glwagner: u_i

navidcy: damn, right!

navidcy: What about du₀?

glwagner: Can we use $du_i$? The `_i` means "initial", which is what the derivative is with respect to.

examples/acoustic_wave.jl
```diff
@@ -356,23 +353,20 @@ end
...
 compiled_grad = Reactant.@compile raise=true raise_first=true sync=true grad_loss(
-    model_ad, dmodel_ad, u₀, du₀,
+    model_ad, dmodel_ad, uᵢ, duᵢ,
     ρ_total, ρᵇᵍ, θ₀, Δt, Nsteps)
 
 @info "Running gradient..."
-du, J = compiled_grad(
-    model_ad, dmodel_ad, u₀, du₀,
-    ρ_total, ρᵇᵍ, θ₀, Δt, Nsteps)
+du, J = compiled_grad(model_ad, dmodel_ad, uᵢ, duᵢ, ρ_total, ρᵇᵍ, θ₀, Δt, Nsteps)
```
navidcy, line 360: @glwagner is the θ₀ really θᵢ here?

navidcy: If I understand correctly that is actually the surface value?

examples/acoustic_wave.jl
```diff
@@ -85,16 +85,16 @@ Uᵢ(z) = U₀ * log((z + ℓ) / ℓ)
...
+δρ = 0.01    # Density perturbation amplitude (kg/m³)
+σ = 20       # Pulse width (m)
 
 gaussian(x, z) = exp(-(x^2 + z^2) / 2σ^2)
 ρ₀ = interior(reference.density, 1, 1, 1)[]
 
-ρ_init(x, z) = adiabatic_hydrostatic_density(z, p₀, θ₀, pˢᵗ, constants) + δρ * gaussian(x, z)
-uᵢ(x, z) = Uᵢ(z) #+ (ℂᵃᶜ / ρ₀) * δρ * gaussian(x, z)
+ρᵢ(x, z) = adiabatic_hydrostatic_density(z, p₀, θ₀, pˢᵗ, constants) + δρ * gaussian(x, z)
+uᵢ(x, z) = Uᵢ(z) # + (ℂᵃᶜ / ρ₀) * δρ * gaussian(x, z)
```
navidcy: why do we have commented code here?
perhaps add the code in a markdown text above and explain when do we want it or not?

navidcy:
Hm... How do we work around this invalid assignment to constant?

```Julia
ERROR: LoadError: LoadError: invalid assignment to constant ##277.uᵢ. This redefinition may be permitted using the `const` keyword.
in expression starting at /__w/Breeze.jl/Breeze.jl/docs/src/literated/acoustic_wave.md:1
when executing the following code block from inputfile `/__w/Breeze.jl/Breeze.jl/examples/acoustic_wave.jl`

```julia
uᵢ  = XFaceField(grid_ad)
duᵢ = XFaceField(grid_ad)
set!(uᵢ, (x, z) -> Uᵢ(z))
set!(duᵢ, 0)
```
```

glwagner:
> Hm... How do we work around this invalid assignment to constant?
> 
> ```julia
> ERROR: LoadError: LoadError: invalid assignment to constant ##277.uᵢ. This redefinition may be permitted using the `const` keyword.
> in expression starting at /__w/Breeze.jl/Breeze.jl/docs/src/literated/acoustic_wave.md:1
> when executing the following code block from inputfile `/__w/Breeze.jl/Breeze.jl/examples/acoustic_wave.jl`
> 
> ```julia
> uᵢ  = XFaceField(grid_ad)
> duᵢ = XFaceField(grid_ad)
> set!(uᵢ, (x, z) -> Uᵢ(z))
> set!(duᵢ, 0)
> ```

The problem is that earlier in the script we define a function `uᵢ`. you cannot build a Field with the same name as a prior defined function, which are const by definition.

navidcy:
Yeah. One of the names should change.
Should we make that function `u_initial(x, z)`? Or use a different name here for the AD vars?

glwagner:
> Yeah. One of the names should change. Should we make that function `u_initial(x, z)`? Or use a different name here for the AD vars?

we sometimes use `ui_func`

navidcy:
Merge?

glwagner:
thank you @navidcy !

PR #724, opened 2026-05-21
Strip nonzero OBC values from acoustic perturbation halos

### LINKED ISSUE #716: Compressible substepper silently misinterprets nonzero OBC values on momentum as perturbations
## Problem

When using `CompressibleDynamics(SplitExplicitTimeDiscretization(); …)` with a user-supplied `OpenBoundaryCondition(value)` on `ρu` / `ρv` / `ρw` where `value` is anything other than `nothing` or zero (e.g. a scalar inflow speed, an array, or an FTS-driven target), the substepper silently treats the value as a *perturbation* of the prognostic momentum from its stage-start linearization, rather than as a full-state Dirichlet boundary value. The result is a runaway acoustic transient at the wall on the first outer step.

## Root cause

At https://github.com/NumericalEarth/Breeze.jl/blob/d86b470ad9b053a787138b911bb63180170d232a/src/CompressibleEquations/acoustic_substepping.jl#L262-L263 the `AcousticSubstepper` constructor inherits user-supplied BCs from the prognostic momentum onto the substepper's perturbation face fields `(ρu)′`, `(ρv)′`, `(ρw)′`:

```julia
bcs_ρu = prognostic_momentum === nothing ? nothing : prognostic_momentum.ρu.boundary_conditions
momentum_perturbation = (u = xface(grid, bcs_ρu), v = yface(grid, bcs_ρv), w = zface(grid, bcs_ρw))
```

The intent is to propagate impenetrability (zero) onto the perturbation field. With a zero / nothing OBC value this is correct. With a nonzero value, `fill_halo_regions!` writes that value into the perturbation field, so at stage entry where `_initialize_perturbation_with_rewind!` makes the interior perturbation zero, the boundary cell jumps to the full-state value while the rest stays zero. The substepper's linearized acoustic dynamics treats that step as a real perturbation and amplifies it.

## Minimal reproducer

```julia
# Stable: scalar OBC=0
ρu_bcs = FieldBoundaryConditions(west=OpenBoundaryCondition(0.0), east=OpenBoundaryCondition(0.0))
# → runs cleanly

# Catastrophic: scalar OBC = ρ·U
ρ_b = …; U = 5.0
ρu_bcs = FieldBoundaryConditions(west=OpenBoundaryCondition(ρ_b*U), east=OpenBoundaryCondition(ρ_b*U))
# → DomainError within first outer step
```

The failure magnitude scales with the OBC value, consistent with the perturbation-amplification mechanism.

The `DomainError` is from [`compressible_time_stepping.jl:145`](src/CompressibleEquations/compressible_time_stepping.jl#L145):

```julia
T_dry = θ^γ * (ρ * Rᵐ / pˢᵗ)^(γ - 1)
```

with γ ≈ 1.4 non-integer. As a separate concern: the `^(γ-1)` call should produce a NaN with a clear stack trace rather than throwing a Julia `DomainError` from low-level Base math — the current behavior makes the crash mode much harder to diagnose than necessary.

## Impact

Anyone trying to drive a compressible Breeze simulation with prescribed lateral inflow will hit this. The mode is silent (no error from the BC machinery — just runaway acoustic amplification), so the misuse is easy to miss until the model crashes mid-run.

## Workaround

Use `OpenBoundaryCondition(0.0)` (or default impenetrability) on the prognostic momentum BCs so the substepper sees zero perturbation at the wall, then apply the prescribed full-state values via a callback at `UpdateStateCallsite` that directly mutates the prognostic data arrays after each halo fill:

```julia
function overwrite_wall!(sim)
    ρu = parent(sim.model.momentum.ρu)
    ρu[Hx+1, Hy+1:Hy+Ny, Hz+1:Hz+Nz] .= prescribed_west
    # likewise east at Hx+Nx+1, ρv south/north, and ρ/ρθ at outermost interior centers
end
add_callback!(sim, overwrite_wall!, IterationInterval(1); callsite = UpdateStateCallsite())
```

## Suggested fix

Extend `AcousticSubstepper` to accept a separate `wall_target_momentum` keyword carrying the prescribed full-state target. The substep loop continues to use the inherited (impenetrability) BCs for its perturbation field, but per-substep boundary updates also overwrite `(ρu)′[wall] = wall_target − ρu_linearization[wall]` so the perturbation channel correctly carries "target minus current linearization." This eliminates the need for the script-level callback workaround.

cc @glwagner

ewquon:
## Summary

Per-side strip in the `AcousticSubstepper` constructor: sides whose BC is an `OpenBoundaryCondition` with a non-`nothing`, non-numerically-zero condition are replaced with `nothing` on the perturbation field, so `fill_halo_regions!` cannot imprint a full-state wall target onto the perturbation halo. Impenetrability defaults, periodic wraps, zero-valued OBCs, and `OBC(nothing)` pass through unchanged. The wall target still re-enters the prognostic state through the prognostic momentum's own BC after `accumulate_momentum_perturbations!` — matching @glwagner's "use `nothing` as the bc" suggestion.

Closes #716.

## Test plan

- [x] New regression testset `Nonzero momentum OBC: stripped on perturbation, stable step` mirrors the issue reproducer on CPU Float64 and GPU{Metal} Float32.
- [x] Full `acoustic_substepping` + `quality_assurance` + `boundary_conditions` + `forcing_and_boundary_conditions` + `substepper_structural` test files pass on CPU. Two Metal Float32 failures (`Backward integration`, `Acoustic divergence damping`) are pre-existing on `main` and unrelated.

### DIFF
--- a/src/CompressibleEquations/acoustic_substepping.jl
+++ b/src/CompressibleEquations/acoustic_substepping.jl
@@ -247,6 +247,29 @@ Adapt.adapt_structure(to, a::AcousticSubstepper) =
 ##### Section 2 — Constructor
 #####
 
+# A condition on an `OpenBoundaryCondition` is "passive" when inheriting it
+# onto the perturbation field would leave the halo at zero: `nothing` (halo
+# fill is a no-op) or a numerically-zero scalar. Arrays, functions, and
+# time-series conditions are treated conservatively as nonzero. Module-level
+# rather than nested in the constructor because multi-method local closures
+# get boxed by Julia 1.12 (caught by the `No Core.Box` quality test).
+is_passive_open_value(::Nothing) = true
+is_passive_open_value(x::Number) = iszero(x)
+is_passive_open_value(_) = false
+
+strip_nonzero_open(bc) = bc
+strip_nonzero_open(bc::BoundaryCondition{<:Open}) =
+    is_passive_open_value(bc.condition) ? bc : nothing
+
+strip_nonzero_open_bcs(::Nothing) = nothing
+strip_nonzero_open_bcs(bcs) =
+    FieldBoundaryConditions(west   = strip_nonzero_open(bcs.west),
+                            east   = strip_nonzero_open(bcs.east),
+                            south  = strip_nonzero_open(bcs.south),
+                            north  = strip_nonzero_open(bcs.north),
+                            bottom = strip_nonzero_open(bcs.bottom),
+                            top    = strip_nonzero_open(bcs.top))
+
 """
 $(TYPEDSIGNATURES)
 
@@ -289,28 +312,10 @@ function AcousticSubstepper(grid, split_explicit::SplitExplicitTimeDiscretizatio
     # impenetrability propagates onto the perturbation momenta, but strip any
     # side carrying a nonzero `OpenBoundaryCondition` to `nothing` so that
     # `fill_halo_regions!` cannot imprint a full-state wall target onto the
-    # perturbation halo (issue #716). A condition is "passive" — leaves the
-    # halo at zero — when it is `nothing` or a numerically-zero scalar; arrays,
-    # functions, and time-series conditions are treated conservatively as
-    # nonzero.
-    is_passive(::Nothing) = true
-    is_passive(x::Number) = iszero(x)
-    is_passive(_) = false
-
-    strip_open(bc) = bc
-    strip_open(bc::BoundaryCondition{<:Open}) = is_passive(bc.condition) ? bc : nothing
-
-    strip_bcs(::Nothing) = nothing
-    strip_bcs(bcs) = FieldBoundaryConditions(west   = strip_open(bcs.west),
-                                             east   = strip_open(bcs.east),
-                                             south  = strip_open(bcs.south),
-                                             north  = strip_open(bcs.north),
-                                             bottom = strip_open(bcs.bottom),
-                                             top    = strip_open(bcs.top))
-
-    bcs_ρu = strip_bcs(prognostic_momentum === nothing ? nothing : prognostic_momentum.ρu.boundary_conditions)
-    bcs_ρv = strip_bcs(prognostic_momentum === nothing ? nothing : prognostic_momentum.ρv.boundary_conditions)
-    bcs_ρw = strip_bcs(prognostic_momentum === nothing ? nothing : prognostic_momentum.ρw.boundary_conditions)
+    # perturbation halo (issue #716).
+    bcs_ρu = strip_nonzero_open_bcs(prognostic_momentum === nothing ? nothing : prognostic_momentum.ρu.boundary_conditions)
+    bcs_ρv = strip_nonzero_open_bcs(prognostic_momentum === nothing ? nothing : prognostic_momentum.ρv.boundary_conditions)
+    bcs_ρw = strip_nonzero_open_bcs(prognostic_momentum === nothing ? nothing : prognostic_momentum.ρw.boundary_conditions)
 
     xface(grid, bcs) = bcs === nothing ? XFaceField(grid) : XFaceField(grid; boundary_conditions = bcs)
     yface(grid, bcs) = bcs === nothing ? YFaceField(grid) : YFaceField(grid; boundary_conditions = bcs)
--- a/src/CompressibleEquations/acoustic_substepping.jl
+++ b/src/CompressibleEquations/acoustic_substepping.jl
@@ -80,8 +80,7 @@ using Oceananigans.Operators:
     Axᶠᶜᶜ, Ayᶜᶠᶜ, Vᶜᶜᶜ
 
 using Oceananigans.Utils: launch!
-using Oceananigans.BoundaryConditions: fill_halo_regions!, BoundaryCondition, Open,
-                                       FieldBoundaryConditions
+using Oceananigans.BoundaryConditions: fill_halo_regions!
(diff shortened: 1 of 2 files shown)


src/CompressibleEquations/acoustic_substepping.jl
```diff
@@ -277,11 +308,14 @@ function AcousticSubstepper(grid, split_explicit::SplitExplicitTimeDiscretizatio
...
-    bcs_ρv = prognostic_momentum === nothing ? nothing : prognostic_momentum.ρv.boundary_conditions
-    bcs_ρw = prognostic_momentum === nothing ? nothing : prognostic_momentum.ρw.boundary_conditions
+    # Perturbation prognostics. Inherit BCs from the prognostic momenta so
+    # impenetrability propagates onto the perturbation momenta, but strip any
+    # side carrying a nonzero `OpenBoundaryCondition` to `nothing` so that
+    # `fill_halo_regions!` cannot imprint a full-state wall target onto the
+    # perturbation halo (issue #716).
+    bcs_ρu = strip_nonzero_open_bcs(prognostic_momentum === nothing ? nothing : prognostic_momentum.ρu.boundary_conditions)
+    bcs_ρv = strip_nonzero_open_bcs(prognostic_momentum === nothing ? nothing : prognostic_momentum.ρv.boundary_conditions)
+    bcs_ρw = strip_nonzero_open_bcs(prognostic_momentum === nothing ? nothing : prognostic_momentum.ρw.boundary_conditions)
```
```suggestion
    bcs_ρu = strip_nonzero_open_bcs(prognostic_momentum.ρu.boundary_conditions)
    bcs_ρv = strip_nonzero_open_bcs(prognostic_momentum.ρv.boundary_conditions)
    bcs_ρw = strip_nonzero_open_bcs(prognostic_momentum.ρw.boundary_conditions)
```

when is `isnothing(prognostic_momentum)`?

test/acoustic_substepping.jl
```diff
@@ -214,6 +214,59 @@ for arch in arches
...
+        ρu_pert_bcs = substepper.momentum_perturbation.u.boundary_conditions
+        @test ρu_pert_bcs.west === nothing
+        @test ρu_pert_bcs.east === nothing
+
+        ref = model.dynamics.reference_state
+        set!(model; θ=300, u=0, qᵗ=0, ρ=ref.density)
+
+        # One forward step must not throw DomainError (the failure mode of #716)
+        # nor produce NaNs.
+        simulation = Simulation(model; Δt=FT(1), stop_iteration=1, verbose=false)
```
glwagner:
```suggestion
        simulation = Simulation(model; Δt=1, stop_iteration=1, verbose=false)
```

src/CompressibleEquations/acoustic_substepping.jl
```diff
@@ -277,11 +308,14 @@ function AcousticSubstepper(grid, split_explicit::SplitExplicitTimeDiscretizatio
...
 
     xface(grid, bcs) = bcs === nothing ? XFaceField(grid) : XFaceField(grid; boundary_conditions = bcs)
```
```suggestion
    xface(grid, bcs) = XFaceField(grid)
```

would this also work?

glwagner: (etc for yface, zface)

src/CompressibleEquations/acoustic_substepping.jl
```diff
@@ -246,6 +247,29 @@ Adapt.adapt_structure(to, a::AcousticSubstepper) =
...
+# fill is a no-op) or a numerically-zero scalar. Arrays, functions, and
+# time-series conditions are treated conservatively as nonzero. Module-level
+# rather than nested in the constructor because multi-method local closures
+# get boxed by Julia 1.12 (caught by the `No Core.Box` quality test).
+is_passive_open_value(::Nothing) = true
+is_passive_open_value(x::Number) = iszero(x)
+is_passive_open_value(_) = false
+
+strip_nonzero_open(bc) = bc
+strip_nonzero_open(bc::BoundaryCondition{<:Open}) =
```
```suggestion
strip_nonzero_open(bc::BoundaryCondition{<:Open, Nothing}) = nothing
strip_nonzero_open(bc::BoundaryCondition{<:Open, Number}) = iszero(bc.condition) ? nothing : bc
strip_nonzero_open(bc::BoundaryCondition{<:Open}) = bc
```

i think you could do stuff like this 😏 but this is minor

glwagner: I am also confused by "0" is a special number here --- can you explain?

src/CompressibleEquations/acoustic_substepping.jl
ewquon: With this simplification, we would no longer propagate user-defined FluxBCs to the perturbation field for the fast update. I think this is physically appropriate, since flux BCs already act on the prognostic state through slow tendencies. The added upshot is that there are also a number of additional simplifications that would come from this path.

src/CompressibleEquations/acoustic_substepping.jl
ewquon: This code is gone now

src/CompressibleEquations/acoustic_substepping.jl
ewquon: This code is gone now

src/CompressibleEquations/acoustic_substepping.jl
glwagner: right. flux BCs are meaningless on non-prognostic fields.

the only boundary condition that matters for velocities is an open boundary condition.

glwagner (approved):

PR #725, opened 2026-05-21
Topology-aware operators for acoustic substepping (+ possible Distributed optimizations)

glwagner:
## Summary

Phase 1 of the acoustic-substepping no-fill plan. Removes per-substep
`fill_halo_regions!` calls from the `CompressibleDynamics`
split-explicit acoustic substep loop and stages the infrastructure
for Phase 2 distributed work.

Key changes:

- **Topology-aware operators in substep kernels.** `_explicit_horizontal_step!`,
  `_build_predictors_and_vertical_rhs!`, and `_thermal_divergence_damping!`
  now use Oceananigans's `δxTᶜᵃᵃ` / `δyTᵃᶜᵃ` plus locally defined
  `∂xTᶠᶜᶜ` / `∂yTᶜᶠᶜ` wrappers, so they read perturbation fields
  through stencils that encode `Periodic` wrap-around and zero-flux at
  `Bounded` walls without halo refills.

- **Fused workspace updates.** The substep-entry `(ρθ)′` snapshot
  used by the Klemp-2018 damping is fused into
  `_build_predictors_and_vertical_rhs!`, killing the per-substep
  `parent .= parent` broadcast. The three parent broadcasts that
  accumulated `(ρu)′, (ρv)′, (ρw)′` into the time-averaged velocity
  slots collapse to one fused kernel.

- **Hoisted kernel configuration.** `acoustic_rk3_substep_loop!`
  configures every repeated kernel once via `configure_kernel`, builds
  stable argument tuples, and wraps a one-shot `convert_to_device` in
  `GC.@preserve` — same pattern as Oceananigans's split-explicit
  free-surface implementation.

- **Launch helpers staged for Phase 2.** New `acoustic_kernel_parameters.jl`
  defines `acoustic_center_parameters` / `acoustic_xface_parameters` /
  ... dispatched on local topology types. Phase 1 uses
  `halo_width = 0` everywhere; Phase 2 will lift that to extend into
  connected halo columns.

Out of scope (deferred): distributed active-halo launches,
`BatchedTridiagonalSolver` `KernelParameters` plumbing, benchmarking
hooks, multi-GPU correctness tests, H100 profiling. See
`acoustic_substepping_no_fill_plan.md` (committed) for the full plan
and `acoustic_substepping_no_fill_plan_review.md` (committed) for the
follow-on review.

## Test plan

- [x] `Pkg.test("Breeze"; test_args=["acoustic_substepping"])` on CPU
      Float64 — all 259 tests pass (full existing acoustic suite
      including MPAS sequencing, vertical tridiag coefficients,
      construction, WS-RK3 model runs without NaN, backward
      integration, IGW stability, dry thermal bubble, balanced rest
      state, divergence damping, upper sponge, `ExnerReferenceState`,
      `SlowTendencyMode`/`HorizontalSlowMode`).
- [ ] Investigate 5 Metal/Float32 failures (backward roundtrip, IGW
      NaN, balanced-state precision); CPU Float64 path is the
      validated reference for Phase 1.
- [ ] Repeat on Linux + CUDA on the H100 box (next session).
- [ ] Confirm no substep-loop `fill_halo_regions!` calls via Nsight
      Systems trace on the H100 (Phase 2 step).


giordano:
Is the link error the same as https://github.com/JuliaGPU/Metal.jl/issues/792?

ewquon:
Heads-up from some M4/Metal testing (I was benchmarking Metal vs a T4 and ran into this): **this branch regresses 2-D Metal compilation of the split-explicit substep, and the 3-D case is still broken.**

The kernel `Breeze.CompressibleEquations.gpu__build_predictors_and_vertical_rhs!` fails at Metal's AIR→metallib **link** stage (`ERROR: Compilation to native code failed`, raised from `Metal/.../compiler/compilation.jl:270`). It is **not** an `InvalidIRError`, and all fields are `Float32` — the AIR is generated but the Metal toolchain rejects the `metallib`.

Verified on Apple M4, Metal.jl 1.9.3, Julia 1.12.6, with the Oceananigans version controlled for:

| Branch | Oceananigans | 2-D `(Periodic, Flat, Bounded)` | 3-D `(Periodic, Periodic, Bounded)` |
|---|---|---|---|
| `main` @ `99549bf` | 0.108.0 | ✅ compiles | ❌ AIR-link fail |
| `main` @ `99549bf` | 0.107.7 | ✅ compiles | ❌ AIR-link fail |
| this branch @ `75053e2` | 0.107.7 | ❌ **regressed** | ❌ AIR-link fail |

The Oceananigans version is ruled out (2-D compiles on `main` at both 0.107.7 and 0.108.0), so the topology-aware-operators rewrite of `_build_predictors_and_vertical_rhs!` is what breaks 2-D, and the 3-D failure pre-dates this branch (it's already present on `main`). Compressible **explicit** compiles + runs on Metal in 3-D, so this is specific to the split-explicit vertical-implicit predictor kernel.

<details>
<summary>Minimal reproducer (Float32; identical setup, only the grid differs)</summary>

```julia
using Oceananigans, Oceananigans.Units
using Oceananigans.TimeSteppers: time_step!
using Breeze
using Breeze: CompressibleDynamics, SplitExplicitTimeDiscretization, dry_air_gas_constant
using Metal: Metal, MetalBackend

Oceananigans.defaults.FloatType = Float32          # Metal has no Float64
arch = GPU(MetalBackend())

constants = ThermodynamicConstants()
g = constants.gravitational_acceleration; Rᵈ = dry_air_gas_constant(constants)
cᵖᵈ = constants.dry_air.heat_capacity; κ = Rᵈ / cᵖᵈ; p₀ = 100000; θ₀ = 300
reference_exner(z)    = 1 - g * z / (cᵖᵈ * θ₀)
reference_pressure(z) = p₀ * reference_exner(z)^(1 / κ)
split() = CompressibleDynamics(SplitExplicitTimeDiscretization(; substeps = 6);
                               surface_pressure = p₀, standard_pressure = p₀,
                               reference_potential_temperature = z -> θ₀)

function build(dim)
    grid = dim == 2 ?
        RectilinearGrid(arch; size = (16, 16),     halo = (5, 5),
                        x = (-8kilometers, 8kilometers), z = (0, 8kilometers),
                        topology = (Periodic, Flat,     Bounded)) :
        RectilinearGrid(arch; size = (16, 16, 16), halo = (5, 5, 5),
                        x = (-8kilometers, 8kilometers), y = (-8kilometers, 8kilometers), z = (0, 8kilometers),
                        topology = (Periodic, Periodic, Bounded))
    model = AtmosphereModel(grid; advection = WENO(), dynamics = split(), timestepper = :AcousticRungeKutta3)
    if dim == 2
        set!(model; θ = (x, z)    -> θ₀ + 10 * max(0, 1 - hypot(x, z - 3kilometers) / 2kilometers),
                    ρ = (x, z)    -> reference_pressure(z) / (Rᵈ * θ₀ * reference_exner(z)), qᵗ = 0)
    else
        set!(model; θ = (x, y, z) -> θ₀ + 10 * max(0, 1 - sqrt(x^2 + y^2 + (z - 3kilometers)^2) / 2kilometers),
                    ρ = (x, y, z) -> reference_pressure(z) / (Rᵈ * θ₀ * reference_exner(z)), qᵗ = 0)
    end
    return model
end

time_step!(build(2), 0.5f0); Metal.synchronize()   # main: OK   | this branch: ERROR
time_step!(build(3), 0.5f0); Metal.synchronize()    # main: ERROR | this branch: ERROR
```
</details>

Two related notes:

- **Metal CI coverage is 2-D only.** `test/acoustic_substepping.jl` runs the split-explicit tiny dry bubble on Metal (good!), but that model is 2-D `(Periodic, Flat, Bounded)`, so neither the 3-D failure nor this 2-D regression is caught. A 3-D split-explicit Metal case would catch both.
- **Re the "spurious Float64 in kernels" point from #688:** confirmed a direct `time_step!(model, Δt::Float64)` errors on Metal with `unsupported use of double value` in `gpu__ssp_rk3_substep!`. `Simulation`/`run!` converts Δt to the grid float type, so normal runs and the test suite are unaffected — only a direct `time_step!` with a `Float64` Δt trips it.

I have the dumped `.ll` / `.air` / `.metallib` for both the `main` 3-D failure and this-branch 2-D failure — happy to attach them here or open a separate tracking issue if you'd prefer to keep it off the PR.

---

**Cross-ref:** this is the Breeze-side trigger for the upstream Metal-compiler bug tracked at JuliaGPU/Metal.jl#792 (which @giordano bisected to #622). Same kernel, same AGX-compiler failure at native-code generation; it just surfaces with a different `NSError` per GPU generation — M1 reports `XPC_ERROR_CONNECTION_INTERRUPTED` (`AGXMetal13_3`, code 2), M4 reports `Compiler encountered an internal error` (`AGXMetalG16G_B0`, code 3). So the 2-D-vs-3-D split here is the same root cause as #792, narrowed: it triggers on the second horizontal dimension and needs no boundary conditions.

glwagner:
We won't be pursuing this

PR #729, opened 2026-05-21
Bump version to 0.5.1 and update Oceananigans range

navidcy:


Project.toml
```diff
@@ -40,7 +40,7 @@ GPUArraysCore = "0.2"
 InteractiveUtils = "<0.0.1, 1"
 JLD2 = "0.5.13, 0.6"
 KernelAbstractions = "0.9"
-Oceananigans = "0.107.4"
+Oceananigans = "0.107.4, 0.108"
 Printf = "1"
```
giordano, line 44: Need to update in all the directories in the workspace to actually test it. Or let dependabot do it for you at https://github.com/NumericalEarth/Breeze.jl/network/updates/27106513/jobs

giordano: Also benchmarking: https://github.com/NumericalEarth/Breeze.jl/blob/b20d657c8b4d2ef0a30f474d152ea17373e702fa/Project.toml#L7

giordano (approved):

navidcy:
Wow. Performance dropped dramatically??

giordano:
Well, the reference point was the huge outlier in https://numericalearth.github.io/BreezeBenchmarks/#benchmark-cbl--ad--dynamics--compressible-explicit--microphysics--nothing--float64-, that's the real issue. No clue what happened there

giordano:
> No clue what happened there

Working hypothesis is that #722 somehow caused extremely large registers spillage in the Reactant benchmark with autodiff (but not the one without autodiff). This caused a significant [performance regression](https://github.com/NumericalEarth/Breeze.jl/pull/722#pullrequestreview-4330872897), but also memory issues which occasionally make the GPU crash. That happened in the benchmark run for https://github.com/NumericalEarth/Breeze.jl/commit/b20d657c8b4d2ef0a30f474d152ea17373e702fa with the result that the timestep was super fast just because it did nothing because the GPU had crashed, thus causing a bogus super fast reference data point this PR was compared with. #732 recovered most of the lost performance, #736 is looking into fully closing the gap.

navidcy:
For one moment I thought it was Oceananigans 0.108
I'll let you merge this whenever you feel it's ripe @giordano!

edit: it's merged, so ignore my last comment :D

PR #736, opened 2026-05-22
Debug AD perf dropoff

dkytezab:


benchmarking/Project.toml
```diff
@@ -29,6 +30,7 @@ CUDA = "5, 6.0"
 CloudMicrophysics = "0.31.4, 0.32, 0.33, 0.34, 0.35"
 Dates = "<0.0.1, 1"
 Enzyme = "0.13.118"
+GPUCompiler = "1.12.0"
```
giordano: You need the tilde to block the minor version, see https://pkgdocs.julialang.org/v1/compatibility/#Version-specifier-format
```suggestion
GPUCompiler = "~1.12.0"
```
Also, it'd be interesting to see if 1.10 makes any difference.

benchmarking/Project.toml
```diff
@@ -28,7 +29,8 @@ ArgParse = "1"
 CUDA = "5, 6.0"
 CloudMicrophysics = "0.31.4, 0.32, 0.33, 0.34, 0.35"
 Dates = "<0.0.1, 1"
-Enzyme = "0.13.118"
+Enzyme = "~0.13.147"
```
giordano: For exact patch version you need
```suggestion
Enzyme = "=0.13.147"
```
otherwise you get anyway the latest 0.13.X

benchmarking/Project.toml
```diff
@@ -11,6 +11,7 @@ CUDA = "052768ef-5323-5732-b1bb-66c8b64840ba"
 CloudMicrophysics = "6a9e3e04-43cd-43ba-94b9-e8782df3c71b"
 Dates = "ade2ca70-3891-5945-98fb-dc099432e06a"
 Enzyme = "7da242da-08ed-463a-9acd-ee780be4f1d9"
+GPUCompiler = "61eb1bfa-7361-4325-ad38-22787b887f55"
```
giordano: Now that we ruled out new GPUCompiler versions as the culprit, we can remove this (and the compat below)

dkytezab:
```
%1024 = stablehlo.reshape %1023 : (tensor<153x256x153x256x153x256xf64>) -> tensor<60088890949632xf64>
```
uhhhhh.... so I'm hitting a similar flavor of issue here https://github.com/EnzymeAD/Enzyme-JAX/pull/2510 on a sharding pass. Seems like something broke further down in the stack? I'm not sure why the dims are being repeated 3 times.

giordano:
That's a new issue, right?

dkytezab:
Yeah just hit this now. But can't find any obviously problematic commits in Enzyme-JAX or Reactant.jl.

giordano:
@dkytezab is this still worth pursuing?  Reactant autodiff performance improved a little bit after https://github.com/NumericalEarth/Breeze.jl/pull/782#issuecomment-4728091093, although probably it didn't completely close the gap caused by #722.

dkytezab:
Yeah, I'm reasonably content enough with the current state to close this atm

PR #739, opened 2026-05-27
Pin GPUCompiler to 1.13.1 and Oceananigans to v0.108.0

dkytezab:
While we resolve some breaking issues with 1.13.2 and Reactant


giordano (approved):

dkytezab:
breaking Oceananigans change :(

PR #689, opened 2026-05-09
Run acousting substepping tests on Metal GPU when functional

giordano:
Follow up to #678.  PR best reviewed by [ignoring whitespace changes](https://github.com/NumericalEarth/Breeze.jl/pull/689/changes?w=1).

PR #740, opened 2026-05-27
Patch for Oceananigans v0.109

navidcy:
Close #749.


dkytezab:
I think we will need to fix the `time_discretization` import as well, i.e. importing it from `Oceananigans.Timesteppers` rather than `Oceananigans.TurbulenceClosures`.

glwagner (approved):

navidcy:
Hm... I see Metal complaining about Float64? Is this new?

https://github.com/NumericalEarth/Breeze.jl/actions/runs/26543355180/job/78189859308?pr=740#step:13:599

I see some Float64 in a Clock? Did Oceananigans 0.109 changed something?

(If somebody can help it's appreciated. I'll get back to this bit later!)

giordano:
> Hm... I see Metal complaining about Float64? Is this new?

Making sure that there are no float64s in float32 kernels is a relatively new check, was introduced in #689. The purpose was precisely to catch these issues.

> If somebody can help it's appreciated

I'd be happy to help, but definitely not today, perhaps tomorrow.

> I see some Float64 in a Clock? Did Oceananigans 0.109 changed something?

@dkytezab?

dkytezab:
I think we need to adapt the last timestep and last stage timestep to the kernel time type, see https://github.com/CliMA/Oceananigans.jl/pull/5634

navidcy:
Still the Mac OS CI fails... :(

navidcy:
It'd be good to push this because it's holding back https://github.com/NumericalEarth/NumericalEarth.jl/pull/305
But I don't know what to do.
I also haven't understood what Oceananigans v0.109 introduced that triggered the Metal CI break.

navidcy:
Seems you fixed it @glwagner; thanks!

navidcy:
Seems like something is NaN-ing in the Docs now?

Somebody else might need to take over this PR? I may not be available much then next couple of weeks -- I'm on holiday.

glwagner:
Status update on the failing CI:

**Fixed** ✅
- macOS Metal Float64-in-kernel errors → convert `Δt` to `eltype(grid)` at every `launch!` site for SSP-RK3, WS-RK3 scalar substep, and the anelastic source / pressure correction.
- GPU `MethodError: no method matching Oceananigans.Architectures.GPU()` → added the Oceananigans dkz-adapt-all branch to `[sources]` of each workspace subproject (`test/`, `docs/`, `examples/`, `benchmarking/`) so the v0.109 branch resolves under all envs.
- GPU `MethodError: Cannot convert ::Float64 to ::Dates.DateTime` → upstream Oceananigans `dkz-adapt-all` fix in [CliMA/Oceananigans.jl@ab5cd25](https://github.com/CliMA/Oceananigans.jl/commit/ab5cd2553) routes `Adapt.adapt_structure(::Clock)`'s `last_Δt`/`last_stage_Δt` through `time_step_type(KT)` so DateTime clocks no longer try to `convert(::Type{DateTime}, ::Float64)`.

**Current GPU summary** — 2868/2933 pass. All non-Reactant tests green; only Reactant compilation tests still fail (see below).

**Still failing** 
- GPU `reactant/{centered,weno}_compilation` + `reactant/advection_diffusion_ad`: `Build` testsets pass, but `Raise backward` produces non-finite loss / zero or NaN AD gradient (FD baseline is finite). On main these tests pass; on this branch the model compiles but the compiled gradient is bad. I tried passing `Δt` through unchanged under `ReactantState` (`BreezeReactantExt` override of `kernel_time_step`) — that did not change the result, so the regression is somewhere else in the Oceananigans 0.109 set (most likely the Reactant ext interaction with new Clock semantics).
- `build-docs`: `DomainError with -Inf` raised from `(1 - g·z/(cᵖᵈ·T₀))^(cᵖᵈ/Rᵈ)` in `adiabatic_hydrostatic_pressure` while doing `set!(model; ρ=ref.density)` in `inertia_gravity_wave.jl`. Passes on main.
- One Reactant gradient benchmark: XLA `gpu_hlo_schedule` reports input/output byte size in the TB range — same pre-existing Reactant memory issue, not from this PR.

Happy to dig into the Reactant AD or the docs `DomainError` next if you'd like, but they look like upstream issues to bounce to @dkytezab / @giordano.

navidcy:
The NaN in the docs could very well be upstream since it wasn't there for Oceananigans v0.108.

glwagner:
> The NaN in the docs could very well be upstream since it wasn't there for Oceananigans v0.108.

I don't understand the theory because the change for 109 is merely moving a function from one module to another?

docs/Project.toml
```diff
@@ -20,6 +20,7 @@ Reactant = "3c362404-f566-11ee-1572-e11a4b42c853"
 
 [sources]
 Breeze = {path = ".."}
+Oceananigans = {url = "https://github.com/CliMA/Oceananigans.jl", rev="dkz-adapt-all"}
```
glwagner: @dkytezab is this necessary for the PR to pass? Can we merge this PR without this pin?

glwagner:
Update: traced the docs `DomainError` to an upstream Oceananigans interpolate bug — fix pushed.

**Root cause.** The example sets density from the (reduced) reference state:

```julia
set!(models[name]; θ=θᵢ, u=U, qᵗ=0, ρ=ref.density)   # ref.density is (Nothing, Nothing, Center)
```

After PR `CliMA/Oceananigans.jl#5586` (in v0.108.1+), `set!(::Field, ::Field)` auto-routes through `interpolate!` when sizes/locations don't match. The kernel computes `at_node = flatten_node(...)` from `to_grid` — on the `(Bounded, Flat, Bounded)` grid that's a 2-tuple `(x, z)`. It then calls

```julia
_fractional_indices((x, z), grid, nothing, nothing, ℓz)
```

Three methods *can* match that signature; Julia picked the most specific one — the 1-tuple `(z,)` branch — because Julia's function-arg destructuring is permissive and `(z,)` happily binds to the first element of a longer tuple. So `fractional_z_index` got called with `x_value` instead of `z_value`, returned a `kk` outside the source's z range, and the kernel wrote garbage memory (NaN past the buffer end) into `model.dynamics.density`. That NaN propagated through `ρθ` and crashed `θ^γ` in `_compute_temperature_and_pressure!` with `DomainError`.

**Fix.** Pushed to `CliMA/Oceananigans.jl@281a3d7f9` on the `dkz-adapt-all` branch this PR already pins via `[sources]`:

1. Explicit `::NTuple{N, Any}` arity on every `_fractional_indices` method so the dispatcher can't silently bind a longer tuple to a shorter pattern.
2. Three 2-tuple "pruning" methods mirroring the existing 3-tuple recursions at lines 228/231/234 — for the case where the source has more `Nothing` locations than the destination has `Flat` dims.

Verified locally with `examples/inertia_gravity_wave.jl`: `set!(model; θ=θᵢ, u=U, qᵗ=0, ρ=ref.density)` now succeeds, and all four model configs (Anelastic, Boussinesq, Compressible, Substepper) advance past iteration 0. Pushed an empty Breeze commit (`b7d3ba32`) to re-trigger CI so the docs job picks up the new `dkz-adapt-all` head.

docs/Project.toml
giordano: I believe that fixed [Float64s leaking inside Float32 kernels](https://github.com/NumericalEarth/Breeze.jl/pull/740#issuecomment-4559407109)

glwagner: Do we have tests that fail if this occurs? Just curious if this is nice to squash or necesssary.

glwagner: Note, I also pushed a very necessary patch to that branch (after I made the comment), so as it stands we do indeed have to merge that first.

giordano: > Do we have tests that fail if this occurs?

Yes, [acoustic_substepping tests were failing](https://github.com/NumericalEarth/Breeze.jl/actions/runs/26543355180/job/78189859308?pr=740#step:13:595)

giordano:
We probably want to tag Oceananigans v0.109.1 after https://github.com/CliMA/Oceananigans.jl/pull/5634 is merged, and then require v0.109.1 here, because we really need the changes in that PR.

giordano:
@glwagner I'm confused by 01dea6f31505f278389aa24df9ff0cffbc943a77: all tests were already passing on the commit before (b7d3ba32ec99ffd82d5be8321f52fba8a897e041), so I don't understand what problem that commit is trying to fix.  Or is that an issue you observed locally?

glwagner:
> @glwagner I'm confused by [01dea6f](https://github.com/NumericalEarth/Breeze.jl/commit/01dea6f31505f278389aa24df9ff0cffbc943a77): all tests were already passing on the commit before ([b7d3ba3](https://github.com/NumericalEarth/Breeze.jl/commit/b7d3ba32ec99ffd82d5be8321f52fba8a897e041)), so I don't understand what problem that commit is trying to fix. Or is that an issue you observed locally?

I didn't want to commit that. However it may be necessary to get Breeze substepping running on Metal. I'm not sure we have such a test now? I can remove the commit though. Better for another PR (probably should also add Metal tests if we want things to work there)

glwagner:
Ok sorry I was confused. We have Metal tests. Claude thinks there is a difference between the runner, and my local machine and that the tests won't pass on my machine. I'm just going to revert the commit for now.

giordano:
I did see some differences between running the tests locally and on CI in the past (the GPU on CI is virtualised, I think), but I haven't looked at this PR on my MacBook, agreed it may still be worth to resolve these differences in a follow up PR.

glwagner:
> I did see some differences between running the tests locally and on CI in the past (the GPU on CI is virtualised, I think), but I haven't looked at this PR on my MacBook, agreed it may still be worth to resolve these differences in a follow up PR.

I would suspect the issues predate this PR since they seem unrelated. Also I don't like the fix on that commit (sorry for committing... I was using Claude and I said "don't commit", but it committed anyways). I don't want to split the kernels just to make Metal work. We'd need a better reason than that. Maybe there is another solution.

navidcy:
> I was using Claude and I said "don't commit", but it committed anyways

rude of Claude

docs/Project.toml
navidcy: Shall I try dropping the pin to the branch to see what happens?

dkytezab: Once [this PR](https://github.com/CliMA/Oceananigans.jl/pull/5634) is merged and `0.109.1` is released, we will unpin

navidcy:
As far as I am concerned , this is all good to go if CI passes

navidcy:
Thanks! I was on my phone and missed that source!

navidcy:
If CI passes we merge this? Is there anything else?

Project.toml
navidcy, line 1: Bump patch release?

PR #742, opened 2026-05-27
Thread clock/model_fields through momentum halo fills (#717)

### LINKED ISSUE #717: `compute_velocities!` discards `clock` and `model_fields` when refilling momentum halos, breaking continuous OBC callables

ewquon:
## What this enables

Time-dependent **Open boundary conditions on momentum** (`ρu`, `ρv`, `ρw`) now work on `AtmosphereModel` — both forms:

- `OpenBoundaryCondition(fts::FieldTimeSeries)` — array/series-backed targets
- `OpenBoundaryCondition(f; parameters=...)` — continuous callables `f(y, z, t, p)`

These are the building blocks for lateral boundary conditions driven by a parent model or a prescribed analytical target. Before this change, both crashed the moment `update_state!` ran (see below), forcing callers to hand-roll a bare boundary-face array refreshed every iteration to sidestep the broken dispatch. With this fix the FTS or callable can be handed straight to `OpenBoundaryCondition`.

## The bug (#717)

`compute_velocities!` and the `AcousticRungeKutta3` substep-loop state recovery refilled the density and momentum halos with bare `fill_halo_regions!(field)` calls, dropping `model.clock` and `fields(model)`. For a time-dependent Open BC on momentum, Oceananigans' `getbc` then has no signature that can evaluate the time argument and falls through to a path that fails:

- continuous callables → `getbc(condition, args...) = condition(args...)` → `ContinuousBoundaryFunction ... is not callable`
- `FieldTimeSeries` → `getbc(::AbstractArray, i, j, ...)` → `BoundsError` on `condition[i, j]` (indexing a 4-D array with 2 indices)

## The fix

Thread `model.clock, fields(model)` through both halo-fill sites, matching the prognostic halo fill already in `update_state!`:

- `src/AtmosphereModels/update_atmosphere_model_state.jl` — `compute_velocities!`
- `src/CompressibleEquations/acoustic_substepping.jl` — substep-loop state recovery

Caveat noted inline: velocities are still stale at these points (recomputed immediately after), so a momentum BC with velocity `field_dependencies` would observe last-stage values. Time/space-dependent BCs (the common case) are unaffected.

## Tests

Added two regression tests to `test/forcing_and_boundary_conditions.jl`, both on a `CompressibleDynamics` model stepped through the acoustic substepper:

- continuous-callable Open BC on `ρu` — verifies `set!`/`time_step!` no longer throw and the west boundary face carries the prescribed `cos(ωt)` value;
- `FieldTimeSeries` Open BC on `ρu` — verifies the boundary face equals slice 1 at t=0 and linearly interpolates to 1.5 at t=5.

Full `Pkg.test()` is green except the pre-existing Metal/Float32 `acoustic_substepping` flakes ("Backward integration" and "Acoustic divergence damping"), which fail identically on `main`.

Fixes #717.


glwagner (approved):
Great. I think there is a helper function that is somethin glike `bc_args(model) = (model.clock, fields(model))` this too, but I can't remember its name right now

ewquon:
Found it — you're thinking of `boundary_condition_args(model)` in `Oceananigans.Models` (`src/Models/boundary_condition_operation.jl`):

```julia
boundary_condition_args(model::Union{NonhydrostaticModel, HydrostaticFreeSurfaceModel}) = (model.clock, fields(model))
```

Exactly the `(model.clock, fields(model))` bundle. One catch: it's restricted to that `Union` (and not exported), so `AtmosphereModel` isn't covered — same situation as `OceananigansModels`/`update_model_field_time_series!`. Using it here would mean extending it Breeze-side:

```julia
Oceananigans.Models.boundary_condition_args(model::AtmosphereModel) = (model.clock, fields(model))
```

I left it out of this PR for scope reasons: Breeze currently uses the explicit `model.clock, fields(model)` form everywhere (including `update_state!` two lines above the site fixed here, and the anelastic stepping path), so adopting the helper cleanly means either generalizing it upstream or a Breeze-wide sweep of all those call sites — bigger than this bugfix. Happy to do that as a follow-up if you'd like the codebase to standardize on `boundary_condition_args`.

glwagner:
> Found it — you're thinking of `boundary_condition_args(model)` in `Oceananigans.Models` (`src/Models/boundary_condition_operation.jl`):
> 
> ```julia
> boundary_condition_args(model::Union{NonhydrostaticModel, HydrostaticFreeSurfaceModel}) = (model.clock, fields(model))
> ```
> 
> Exactly the `(model.clock, fields(model))` bundle. One catch: it's restricted to that `Union` (and not exported), so `AtmosphereModel` isn't covered — same situation as `OceananigansModels`/`update_model_field_time_series!`. Using it here would mean extending it Breeze-side:
> 
> ```julia
> Oceananigans.Models.boundary_condition_args(model::AtmosphereModel) = (model.clock, fields(model))
> ```
> 
> I left it out of this PR for scope reasons: Breeze currently uses the explicit `model.clock, fields(model)` form everywhere (including `update_state!` two lines above the site fixed here, and the anelastic stepping path), so adopting the helper cleanly means either generalizing it upstream or a Breeze-wide sweep of all those call sites — bigger than this bugfix. Happy to do that as a follow-up if you'd like the codebase to standardize on `boundary_condition_args`.

Up to you. Might be less work to put in now, but I see the point about being prim and proper with scope.

ewquon:
Who are you calling "prim and proper"?

Looks like we already have `boundary_condition_args` defined in Breeze, it was just sitting there.

glwagner:
> Who are you calling "prim and proper"?
> 
> Looks like we already have `boundary_condition_args` defined in Breeze, it was just sitting there.

I was speaking to Eliot-Claude not Eliot-human of course

ewquon:
**CI failures are environmental, not from this PR.** Reproduced on a Tesla T4 devbox resolving the same broken stack (GPUCompiler 1.13.2, Oceananigans 0.108.2, Reactant 0.2.262):

| CI failure | Devbox finding |
|---|---|
| `build-docs` NaN | **Reproduced.** `inertia_gravity_wave.jl` NaNs at **iteration 0** → `DomainError` in `compute_temperature_and_pressure!`. Periodic/Flat/Bounded `ExplicitTimeStepping`, **no momentum OBCs** — this PR's path is not on the stack. |
| macOS Metal acoustic failures | **Inverse-confirmed.** `acoustic_substepping` on **CUDA passes 151/151** on the devbox. The failures are Metal/Float32-backend-specific numerics. |
| aws CUDA suite | CI log isolates failures to `reactant/{advection_diffusion_ad, centered_compilation, weno_compilation}` → XLA OOM (1.5 TiB allocs). The AtmosphereModel path (`acoustic_substepping`) passes on the same CUDA hardware. |
| Reactant benchmarks | Same XLA OOM signature; matches the known regression #736 / #739 is pinning away from. |

Independent control: **#739 (a deps-pin-only PR) fails the identical set of jobs, plus ubuntu CPU — which this PR *passes*.** A pure-deps PR cannot be failing because of `compute_velocities!`.

Expected to clear once #739 lands and this branch is rebased.

test/forcing_and_boundary_conditions.jl
```diff
@@ -126,6 +126,76 @@ end
...
+                           x=(0, 1000), y=(0, 1000), z=(0, 200),
+                           topology=(Bounded, Bounded, Bounded))
+    dynamics = CompressibleDynamics(SplitExplicitTimeDiscretization();
+                                    reference_potential_temperature=FT(300),
+                                    surface_pressure=FT(1e5))
+
+    # 2-D (y, z) boundary slice for a west OBC on ρu (Face, Center, Center).
+    # Slice values 1, 2, 3 at times 0, 10, 20 so the boundary value linearly
+    # interpolates to 1.5 at t = 5.
+    times = [FT(0), FT(10), FT(20)]
```
giordano: Just slightly simpler to type, without repeating `FT` multiple times
```suggestion
    times = FT[0, 10, 20]
```

test/forcing_and_boundary_conditions.jl
```diff
@@ -126,6 +126,76 @@ end
...
+    model = AtmosphereModel(grid; dynamics, boundary_conditions=(; ρu=ρu_bcs))
+    set!(model; θ=FT(300), ρ=FT(1.17))
+
+    # t = 0: west boundary face equals the first slice.
+    @test @allowscalar(model.momentum.ρu[1, 1, 1]) ≈ FT(1) atol=sqrt(eps(FT))
+
+    # t = 5: halfway between slices 1 and 2 → linear interpolation gives 1.5.
+    time_step!(model, FT(5))
+    @test model.clock.iteration == 1
+    @test @allowscalar(model.momentum.ρu[1, 1, 1]) ≈ FT(1.5) atol=sqrt(eps(FT))
```
giordano: When comparing to one, an absolute tolerance of `sqrt(eps(FT))` is equivalent to the default relative tolerance, no?
```suggestion
    @test @allowscalar(model.momentum.ρu[1, 1, 1]) ≈ FT(1)

    # t = 5: halfway between slices 1 and 2 → linear interpolation gives 1.5.
    time_step!(model, FT(5))
    @test model.clock.iteration == 1
    @test @allowscalar(model.momentum.ρu[1, 1, 1]) ≈ FT(1.5)
```

test/forcing_and_boundary_conditions.jl
```diff
@@ -126,6 +126,76 @@ end
...
+    set!(model; θ=FT(300), ρ=FT(1.17))
+
+    # West boundary face (i=1) carries the prescribed value at the current time.
+    bc_value(t) = FT(1.17) * cos(FT(0.01) * t)
+    @test @allowscalar(model.momentum.ρu[1, 1, 1]) ≈ bc_value(model.clock.time) atol=sqrt(eps(FT))
+
+    # Stepping re-evaluates the BC at the new time without error.
+    time_step!(model, FT(10))
+    @test model.clock.iteration == 1
+    @test @allowscalar(model.momentum.ρu[1, 1, 1]) ≈ bc_value(model.clock.time) atol=sqrt(eps(FT))
```
giordano: I ***really*** prefer not to use absolute tolerance, and instead set the relative tolerance (if [the default one](https://docs.julialang.org/en/v1/base/math/#Base.isapprox) is not enough) _if_ needed, unless the reference number is exactly zero.  The machine precision [`eps`](https://docs.julialang.org/en/v1/base/base/#Base.eps-Tuple{Type{%3C:AbstractFloat}}) is about a "scale", but the absolute tolerance is scale-less by definition.

ewquon: Agree.

PR #744, opened 2026-05-28
Pin oceananigans v0.108.0

navidcy:
Temporary fix until #740 is ready.


giordano:
This should be fixed in the registry, not here

giordano:
https://github.com/JuliaRegistries/General/blob/9d693d21bcdf921d32674afe5f958a495a0e2c4c/B/Breeze/Compat.toml#L50 should be
```toml
Oceananigans = "0.107.4 - 0.108.0"
```

navidcy:
Thanks; https://github.com/JuliaRegistries/General/pull/156488 also https://github.com/JuliaRegistries/General/pull/156490

giordano:
Actually, we still need this until #740 is done.  But the change in the General registry was still necessary because that affects relevant _past_ versions, which this PR alone wouldn't affect.  Will reopen this PR after #739 is merged.

giordano:
Well, we need both of them together, I'll apply the changes here to that PR.

navidcy:
Okie!

PR #746, opened 2026-05-28
Fire model callsites and advance FieldTimeSeries in AtmosphereModel (#719)

### LINKED ISSUE #719: AtmosphereModel never fires `UpdateStateCallsite` or `TendencyCallsite` callbacks
## Summary

`Breeze.AtmosphereModels.update_state!` (in `AtmosphereModels/update_atmosphere_model_state.jl:12-27`) accepts a `callbacks` argument but never iterates over it. The same holds for `compute_tendencies!`. As a result, any user callback registered with `callsite = UpdateStateCallsite()` or `callsite = TendencyCallsite()` is silently dropped for `AtmosphereModel`s.

Compare to `Oceananigans.Models.NonhydrostaticModels`, which honors both callsites:

* `update_nonhydrostatic_model_state.jl:48` — `callback.callsite isa UpdateStateCallsite && callback(model)`
* `compute_nonhydrostatic_tendencies.jl:37` — `callback.callsite isa TendencyCallsite && callback(model)`

## Why this matters

The only callsite Breeze currently honors is `TimeStepCallsite` (fired from `Simulations/run.jl:256`). A callback there runs *after* `time_step!` completes — so anything it writes to a boundary face is overwritten by the next iteration's first `fill_halo_regions!` (Open BCs clobber the boundary face with `getbc`). `UpdateStateCallsite` is Oceananigans' designated site for "set state after halos are filled, before tendencies are computed," and is the only callsite at which a user can override BC-filled values for the upcoming tendency calculation. Without it, boundary overwrites (e.g., in the specified zone of a Davies lateral BC) and similar field-surgery patterns are impossible in Breeze.

## Reproducer

```julia
using Breeze, Oceananigans
using Oceananigans: UpdateStateCallsite, TimeStepCallsite

grid = RectilinearGrid(size=(8,8,4), x=(0,1000), y=(0,1000), z=(0,200),
                       topology=(Bounded, Bounded, Bounded))
constants = ThermodynamicConstants(eltype(grid))
reference_state = ReferenceState(grid, constants;
                                 surface_pressure=101325.0, potential_temperature=300.0)
                                surface_pressure=101325.0,
                                reference_potential_temperature=300.0)
model = AtmosphereModel(grid; dynamics, thermodynamic_constants=constants)
set!(model; θ=300.0, ρ=reference_state.density)

fired_update = Ref(0)
fired_step   = Ref(0)
update_cb(sim) = (fired_update[] += 1; nothing)
step_cb(sim)   = (fired_step[]   += 1; nothing)

sim = Simulation(model; Δt=0.01, stop_iteration=3)
add_callback!(sim, update_cb, IterationInterval(1); callsite = UpdateStateCallsite())
add_callback!(sim, step_cb,   IterationInterval(1); callsite = TimeStepCallsite())
run!(sim)

@show fired_update[]  # observed: 0  (expected: ≥ 3)
@show fired_step[]    # observed: 4  (sanity check — TimeStepCallsite works)
```

## Suggested fix

In `update_atmosphere_model_state.jl:12-27`, after `compute_auxiliary_variables!(model)` and `update_boundary_conditions!`, iterate the callbacks tuple:

```julia
for callback in callbacks
    callback.callsite isa UpdateStateCallsite && callback(model)
end
```

Same pattern for `TendencyCallsite` in `compute_tendencies!`. Mirror the dispatch sites in NonhydrostaticModel.

The `callbacks` argument is already correctly wired from `Simulations/run.jl:246-247` through the timesteppers (e.g. `acoustic_runge_kutta_3.jl:193` passes it to `update_state!`); only the iteration is missing.

## Workaround

`OpenBoundaryCondition(array)` with a 2D cache refreshed via a callback at `IterationInterval(1)` (default `TimeStepCallsite`). The cache update arrives in time for the next iteration's `fill_halo_regions!`, which then writes the refreshed value to the boundary face. Pairs with the workaround in NumericalEarth/Breeze.jl#717.

## Context

Encountered while building a "spec_zone overwrite" pattern for compressible lateral boundary conditions during analytical vortex transit validation. Together with NumericalEarth/Breeze.jl#717 (continuous-callable OBCs broken) and NumericalEarth/Breeze.jl#716 (substepper perturbation BC inheritance), these are three Breeze-side blockers for clean compressible LBC validation.

ewquon:
## What this enables

Two related capabilities on `AtmosphereModel` that were silently broken:

- **Callbacks registered at `UpdateStateCallsite` / `TendencyCallsite` now fire.** Previously, `add_callback!(...; callsite = UpdateStateCallsite())` was a no-op on `AtmosphereModel` because the per-stage dispatch loops were missing from `update_state!` and `compute_tendencies!`.
- **`FieldTimeSeries` embedded in BCs or forcings now advance their time pointer through `update_state!`.** Previously the call fell through to the no-op `update_model_field_time_series!(::AbstractModel, ::Clock)` fallback because `AtmosphereModel` is not in Oceananigans' `OceananigansModels` Union, so any FTS in a BC or forcing was frozen at its initial time slice.

Together these enable chunked / on-disk FTS-backed Open BCs and callback-based mid-step interventions (e.g. wall-value overwrites between RK stages).

## The fix

`src/AtmosphereModels/update_atmosphere_model_state.jl`:

- Add `UpdateStateCallsite` and `TendencyCallsite` dispatch loops to `update_state!` and `compute_tendencies!`, mirroring `NonhydrostaticModel` (`update_nonhydrostatic_model_state.jl:47-49`, `compute_nonhydrostatic_tendencies.jl:36-38`).
- Extend `Oceananigans.Models.update_model_field_time_series!(::AtmosphereModel, ::Clock)`, walking `(fields(model), model.forcing)` exactly like the `OceananigansModels` method. Called at the top of `update_state!` so halo fill / tendency computation see the current time slice.

The `update_model_field_time_series!` gap was flagged by @glwagner in his review of #719.

## Tests

New `test/atmosphere_model_callbacks.jl` (10 tests):

- `UpdateStateCallsite`, `TendencyCallsite`, `TimeStepCallsite` firing counters; the two per-stage callsites fire equal counts.
- `compute_tendencies!(model)` (no callbacks) remains callable.
- `update_model_field_time_series!` dispatches to the `AtmosphereModel` method (not the no-op fallback) and discovers an embedded FTS through the model walk.
- A chunked `InMemory(2)` `FieldTimeSeries` Open BC on `ρu` of a compressible substepper model: stepping past the in-memory window advances `backend.start` — the load-bearing #719 signal that wouldn't move without the extension.

Fixes #719.

### DIFF
--- a/src/AtmosphereModels/update_atmosphere_model_state.jl
+++ b/src/AtmosphereModels/update_atmosphere_model_state.jl
@@ -1,5 +1,6 @@
 using ..Thermodynamics: Thermodynamics, mixture_gas_constant
 
+using Oceananigans: UpdateStateCallsite, TendencyCallsite
 using Oceananigans.BoundaryConditions: fill_halo_regions!, compute_x_bcs!, compute_y_bcs!, compute_z_bcs!,
                                        update_boundary_conditions!
 using Oceananigans: Face
@@ -21,7 +22,12 @@ function TimeSteppers.update_state!(model::AtmosphereModel, callbacks=[]; comput
     update_radiation!(model.radiation, model)
     compute_forcings!(model)
     microphysics_model_update!(model.microphysics, model)
-    compute_tendencies && compute_tendencies!(model)
+
+    for callback in callbacks
+        callback.callsite isa UpdateStateCallsite && callback(model)
+    end
+
+    compute_tendencies && compute_tendencies!(model, callbacks)
 
     tracer_specific_to_density!(model) # convert specific tracer distribution to tracer density
 
@@ -246,7 +252,7 @@ end
     @inbounds temperature[i, j, k] = T
 end
 
-function compute_tendencies!(model::AtmosphereModel)
+function compute_tendencies!(model::AtmosphereModel, callbacks=[])
     grid = model.grid
     arch = grid.architecture
 
@@ -333,6 +339,10 @@ function compute_tendencies!(model::AtmosphereModel)
     #####
 
     compute_dynamics_tendency!(model)
+
+    for callback in callbacks
+        callback.callsite isa TendencyCallsite && callback(model)
+    end
 
     return nothing
 end
--- a/src/AtmosphereModels/update_atmosphere_model_state.jl
+++ b/src/AtmosphereModels/update_atmosphere_model_state.jl
@@ -1,20 +1,40 @@
 using ..Thermodynamics: Thermodynamics, mixture_gas_constant
 
-using Oceananigans: UpdateStateCallsite, TendencyCallsite
+using Oceananigans: Face, UpdateStateCallsite, TendencyCallsite
 using Oceananigans.BoundaryConditions: fill_halo_regions!, compute_x_bcs!, compute_y_bcs!, compute_z_bcs!,
                                        update_boundary_conditions!
-using Oceananigans: Face
+using Oceananigans.Fields: flattened_unique_values
 using Oceananigans.Grids: topology
 using Oceananigans.ImmersedBoundaries: mask_immersed_field!
 using Oceananigans.Models: boundary_condition_args
-using Oceananigans.TimeSteppers: TimeSteppers
+using Oceananigans.OutputReaders: update_field_time_series!, extract_field_time_series
+using Oceananigans.TimeSteppers: TimeSteppers, Clock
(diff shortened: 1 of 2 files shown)


glwagner (approved):
excellent

PR #747, opened 2026-05-28
Per-substep open-boundary enforcement for split-explicit substepper

### LINKED ISSUE #738: Split-explicit substepper loses mass at open boundaries
## Summary

The split-explicit (`AcousticRungeKutta3` / `SplitExplicitTimeDiscretization`) timestepper loses several percent of density when an open lateral boundary carries a transient inflow, while the explicit timestepper holds density flat under identical conditions. Root cause: the acoustic substep loop treats the perturbation fields' open boundaries as **reflecting** (impenetrable normal momentum + zero-gradient scalars) and freezes the boundary mass flux in the stage-entry slow tendency for the whole RK stage. WRF and ERF instead refresh the specified lateral boundary (momentum, mass, θ) **on every acoustic substep**.

## Reproduction

2.5-D Lamb–Oseen vortex transiting a `(Bounded, Bounded, Bounded)` domain through an open west boundary (OBC on `ρu`/`ρv`, `ValueBoundaryCondition` on `ρ`/`ρθ`, Davies fringe). `Nx=Ny=128, Nz=4`, WENO5, `u_θ_peak=20 m/s`, `|U_bg|=2 m/s`. Same IC, grid, advection, and BCs for both timesteppers; only the time discretization differs.

| ρ_min | t = 0 | t = 3000 s |
|---|---|---|
| **explicit**, transit | 1.145 | 1.138 (flat) |
| **substep**, transit | 1.145 | **1.080 (−5.7%)** |
| **explicit**, stationary (vortex parked at center) | 1.1385 | 1.1386 (flat) |
| **substep**, stationary | 1.1386 | 1.1323 (−0.55%) |

The drift is entirely substep-specific (both explicit cells are flat) and is dominated by the boundary transit: parking the vortex at domain center (no boundary inflow) cuts it ~10× (−5.7% → −0.55%). The residual −0.55% is a separate, smaller interior effect (the acoustic split's handling of the IC's discrete cyclostrophic imbalance); it is also substep-specific but independent of the boundary mechanism.

## Root cause

In `src/CompressibleEquations/acoustic_substepping.jl`, the substep continuity update (`_build_predictors_and_vertical_rhs!`) advances the density perturbation as
```
ρ′★ = ρ′ + Δτ·(Gˢρ − div_xy((ρu)′,(ρv)′)) − δτˢ⁻·∂z((ρw)′)
```
- `Gˢρ = −div(full momentum)` is the slow tendency, computed once per RK stage (`compressible_density_tendency.jl`) and **frozen** across all substeps. It carries the boundary inflow only at its stage-entry value.
- The perturbation momenta `(ρu)′,(ρv)′` use topology-default **impenetrability** at the open boundary (introduced in #724 / b20d657 to stop the #716 wall-jump crash), so they contribute **zero** boundary flux to the continuity update and **reflect** interior-generated acoustic waves. The perturbation scalars `ρ′,(ρθ)′` are filled with the default (zero-gradient) halo, which likewise reflects the pressure wave.

So for an RK stage spanning ~30 acoustic substeps, the boundary is a frozen flux + a reflecting wall. During a transient inflow (vortex core crossing) the reflected/under-accounted acoustic mass accumulates as a density bias.

## How WRF, ERF, and MPAS handle this (source-verified)

All three update the **specified lateral boundary every acoustic substep**, consistently across momentum + mass + θ, and apply the **relaxation zone at the slow (RK) level** — Breeze already matches the latter (Davies fringe in the slow tendency) but not the former.

- **ERF** (`erf-model/ERF`): each fast step calls `apply_bcs(..., fast_only=true)` (`Source/TimeIntegration/ERF_TI_substep_fun.H`), which does `FillIntermediatePatch` for **ρ, ρθ, and the momenta/velocities** from the physical boundaries (`Source/TimeIntegration/ERF_TI_utils.H`). Relaxation zone: `realbdy_compute_interior_ghost_rhs` in `ERF_TI_slow_rhs_pre.H` (slow `slow_dt`).
- **WRF ARW** (`wrf-model/WRF`, `dyn_em/solve_em.F`): inside `small_steps : DO iteration = 1, number_of_small_timesteps`, after `advance_uv` it calls `spec_bdyupdate(u_2,…,dts_rk)` / `spec_bdyupdate(v_2,…)`, and after `advance_mu_t` it calls `spec_bdyupdate(t_2,…)` / `spec_bdyupdate(mu_2,…)` — the specified boundary on **momentum, mass (μ), and θ is advanced every acoustic small step** at the small-step size `dts_rk`. `relax_bdy_dry` is called once at `rk_step == 1`.
- **MPAS-A** (`MPAS-Dev/MPAS-Model`, `src/core_atmosphere/dynamics/mpas_atm_time_integration.F`): the acoustic loop `do small_step = 1, number_sub_steps(rk_step)` calls `atm_advance_acoustic_step`. Inside `atm_advance_acoustic_step_work`, the **perturbation-form** spec-zone branch advances `rho_pp += dts*tend_rho`, `rtheta_pp += dts*tend_rt`, `rw_p += dts*tend_rw` every substep, and the momentum update `ru_p += dts*(tend_ru − (1 − specZoneMaskEdge)*pgrad)` switches off the interior pressure gradient in the spec zone, advancing by the boundary tendency instead.

MPAS is the closest analog to Breeze (it also evolves acoustic **perturbations** `rho_pp, rtheta_pp, ru_p`): in the spec zone it advances the perturbation by the prescribed boundary tendency rather than the reflecting interior acoustic update.

## Fix attempts (prototyped, Nx=128 horizontal LO transit, stop=3000 s)

Baseline for comparison: explicit holds ρ_min flat (~1.138); substep drifts 1.145 → 1.080 (−5.7%). A working fix must restore mass conservation **and** leave the entering vortex intact (baseline substep tracks it to ~100 m centroid error, u_θ ≈ 20).

| attempt | where it acts | result |
|---|---|---|
| **Zero the outermost interior cell** `ρ′[1],(ρθ)′[1]` each substep | interior cell | stable; mass fixed (ρ_min 1.142) but **erases the entering vortex** (u_θ 20→16, centroid err 102→**878 m**) — pins the cell to the *base* (no inflow anomaly) |
| **Homogeneous `ValueBoundaryCondition(0)` on `ρ′,(ρθ)′` halos** | halo | **unstable** — growing `w` mode, NaN by ~iter 100 (negative-reflection/pressure-release Dirichlet self-amplifies in the acoustic substep) |
| **In-kernel mask, keeping `Gˢρ`** (drop only the fast `∇ʰ_M` in spec-zone cells) | predictor kernel | **instant NaN** (iter 1–2). Breeze's slow tendency `Gˢρ = −div(ρu)` carries the *full inflow-momentum divergence*, which must be balanced by the fast `∇ʰ_M`; masking the fast term alone leaves it unbalanced → collapse |
| **In-kernel driving tendency** (replace `Gˢρ−∇ʰ_M` in spec-zone cells with a drive toward the prescribed wall value) | predictor kernel | stable ~4 iters then **NaN**. The mass predictor is replaced but the cell stays coupled to the rest of the substep — post-solve recovery re-adds `−δτ·∂z(ρw′)` and the vertical `w`-solve reacts to the driven `ρ′★` — so mass is on the driving tendency while momentum/`w` in that cell remain on acoustic dynamics → mismatch compounds → blowup |
| **Post-recovery cell-pin** — after the full substep, set `ρ′[1] = (ρᴸ[halo]−ρᴸ[1])/2` so the recovered density equals the prescribed wall value `v` | after the substep | **works**: stable, mass 1.145 → 1.139 (**−0.55%**, matches explicit), vortex preserved (centroid err **62 m**, u_θ ≈ 19). One blemish: the abrupt per-substep overwrite injects a divergence that raises `max|w|` to ~0.6 |

**Key lessons.** (i) The boundary cell must be pinned to the *prescribed inflow value* `v`, not to the base — pinning to base erases the vortex. The base halo already encodes `v` (`ρᴸ[halo] = 2v − ρᴸ[1]`, since `update_state!` applied the prognostic `ValueBC` at stage entry). (ii) A partial in-kernel replacement (mass only) destabilizes because the acoustic substep is coupled (predictor → `w`-solve → recovery); the spec-zone treatment must be consistent across **all** coupled fields, as MPAS does (it replaces `tend_rho`, `tend_rtheta`, `tend_ru`, and zeroes `tend_rw` together). (iii) Acting **after** the full coupled substep (the post-recovery cell-pin) sidesteps the internal-consistency problem, at the cost of an injected-divergence `max|w|`.

## Proposed fix

Two viable directions, both per-substep, both negligible cost (boundary-plane, no extra kernel launches):

1. **Canonical (MPAS-faithful), in-kernel.** In spec-zone cells/faces replace the dynamics tendency with the boundary driving tendency toward the prescribed wall value, applied **consistently across the coupled fields** — predictor (`ρ′,(ρθ)′`), post-solve recovery (mask the `∂z(ρw′)` correction there), the vertical `w`-solve (`tend_rw → 0`), and the normal momentum (already pinned to the prescribed value by the impenetrable halo + base). This is the literal MPAS `atm_advance_acoustic_step` + `atm_bdy_adjust_dynamics_speczone_tend` treatment; the difficulty is purely the multi-kernel consistency above.

2. **Pragmatic, post-recovery.** Pin the outermost open-boundary cell of `ρ′,(ρθ)′` to the prescribed wall value `v` after each substep (the working "post-recovery cell-pin" row above). Simplest and already verified to fix mass + preserve the vortex; the remaining work is reducing `max|w|` (e.g. a gradual relaxation toward `v` over the stage instead of an abrupt per-substep overwrite).

Recommendation: the post-recovery cell-pin is the minimal working fix today; the in-kernel form is the cleaner long-term solution once the coupled-field consistency is handled.

## Related

- #716 (substepper inherits nonzero OBC as perturbation amplitude — runaway crash)
- #724 / b20d657 (stripped OBC values from perturbation halos → made the perturbation boundary impenetrable; fixed the crash but introduced this conservation gap)
- #708, #722 (specific-forcing dispatch; momentum-OBC propagation)

Diagnosis and fix attempts performed on a prototype branch (env-gated). An `OpenBoundaryCondition` + `PerturbationAdvection` radiation BC on the perturbation momentum alone was also tried and is **not** the fix (it NaNs — radiating one field of the coupled acoustic characteristic while the others reflect is inconsistent).

ewquon:
Closes #738.

## Summary

The acoustic substep loop's perturbation scalars `ρ′, (ρθ)′` carry the default `NoFluxBoundaryCondition` on `Bounded` dims, whose halo fill mirrors the interior (`c[halo] = c[1]` — zero-gradient). At the open boundary face this makes `∂ₓ p′ = (p′[1] − p′[0])/Δx = 0` every substep, since `p′ ∝ γRᵐᴸ·Πᴸ·ρθ′` is built from `ρθ′`. The boundary-face momentum update `(ρu)′ += Δτ·(Gⁿρu − ∂x_pᴸ − ∂x_p′)` still picks up the slow tendency `Gⁿρu` and the frozen base pressure gradient `∂x_pᴸ` (both fixed at stage entry), but the **acoustic-feedback term** `∂x_p′` — the only contribution that responds to within-substep acoustic activity — is zero at the face. The interior acoustic system is therefore decoupled from the boundary: outgoing acoustic waves see no PGF response at the open face and reflect, while the boundary mass flux is carried only by the frozen stage-entry slow tendency `Gˢρ`. This biases the discrete mass balance under transient open-boundary inflow; for example, a canonical vortex-transit study hit −5% to −27% density drift on the substepper while explicit held density constant.

Equivalently: at the boundary face/cell the substepping degenerates. With `∂x_p′ = 0` and the other horizontal tendencies frozen, `(ρu)′ += Δτ·(frozen)` applied `n_sub` times is identical to `(ρu)′ += Δt·(frozen)` for `Δt = n_sub·Δτ` — i.e. the outer step. The same holds for `ρ′,(ρθ)′` at the outermost cell, whose horizontal θ-flux divergence vanishes once the boundary face momentum is decoupled. Whatever apparent stability the substepper exhibited at the boundary came entirely from the implicit column solve (degenerate for our `Nz = 4`, `Δx ≈ Δz` validation setup) and the frozen-tendency advance — not from acoustic substepping.

This PR fixes the drift by relaxing the outermost open-boundary cell of `ρ′, (ρθ)′` toward the prescribed wall value `v` after each substep's post-solve recovery — breaking the `p′[0] = p′[1]` mirror so the boundary face sees a non-zero `∂ₓp′` and outgoing acoustic momentum can exit. Since `update_state!` applied the prognostic `ValueBoundaryCondition` to the base at stage entry, `ρᴸ[halo] = 2v − ρᴸ[cell]`, so the target perturbation is `(ρᴸ[halo] − ρᴸ[cell]) / 2` — read directly from the base field, no extra BC plumbing. The relaxation factor α is exposed as `SplitExplicitTimeDiscretization(; open_boundary_relaxation = 0.5)`. The enforcement is unconditional and a no-op on any side without an active `OpenBoundaryCondition` (periodic, walls, `OpenBoundaryCondition(nothing)` all skip it), so existing setups pay no cost.

## Background — how we landed here

The investigation traced the canonical treatment to WRF (`spec_bdyupdate` inside `small_steps`), ERF (`apply_bcs(fast_only=true)`), and MPAS (perturbation-form spec-zone branch in `atm_advance_acoustic_step_work`) — all three enforce the specified lateral boundary on every acoustic substep. Several in-kernel ports of that pattern destabilized (instant NaN: keeping `Gˢρ` while masking the fast term leaves the spec cell unbalanced; replacing only the mass predictor with a driving tendency NaNs after a few iters because post-solve recovery and the vertical `w`-solve still couple the cell into the acoustic dynamics).

What works — and matches an existing NWP code closely — is FV3-LAM (`fv_regional_bc.F90`, called inside the `do it=1, n_split` acoustic loop): each substep blends interior values toward externally supplied boundary data via `array = (1 − w)·array + w·blend_value`, with an exponential weight `w = exp(−(0.5 + 10·d))` across `nrows_blend` rows. **This PR is the one-cell limit of FV3's blend**: a single boundary row, with α playing the role of FV3's outermost-row weight (≈ 0.6). The simpler form is acceptable here because the Davies fringe already provides a wider relaxation zone at the slow-tendency level; this PR only adds the missing per-substep enforcement in the outermost cell.

Acting after the full coupled substep (rather than inside the predictor) sidesteps the multi-kernel consistency problem the in-kernel approaches hit — the entire acoustic update completes consistently, then the boundary cell is reset.

## Validation

Five 2-D vortex-transit scenarios on a bounded square domain with time-varying `OpenBoundaryCondition`s on the boundary-normal momenta (ρu, ρv), `ValueBoundaryCondition`s on density (ρ) and density-weighted potential temperature (ρθ), and Davies relaxation zones adjacent to the lateral boundaries; we use a default substep BC weight of α=0.5. All five share `|U_bg|` = 2 m/s. Scenarios: In the "horizontal" and "skim" scenarios, the vortex enters the domain on a single boundary; the "skim" scenarios introduce vortices whose trajectories coincide with the inner edge of a relaxation zone; the "diagonal" scenarios introduce vortices at domain corners that traverse the domain diagonally.

`max|w|` is normalized by `|U_bg|` so a value of 1 means the spurious vertical motion is as large as the background flow:

| scenario | drift no fix | drift with fix | max\|w\|/\|U_bg\| no fix | max\|w\|/\|U_bg\| with fix |
|---|---|---|---|---|
| horizontal       | −8%       | −0.01%     | 0.425  | 0.105 |
| **skim_low**     | **−27%**  | **−0.01%** | 0.007  | 0.038 |
| skim_high        | −9%       | −0.01%     | 0.0015 | 0.055 |
| diagonal_sw_ne   | −10%      | −0.02%     | 0.100  | 0.160 |
| diagonal_se_nw   | −8%       | −0.02%     | 0.0025 | 0.130 |

Vortex amplitude is ≥99% retained in all cases. The fix improves the worst-case `max|w|/|U_bg|` (horizontal, 0.43 → 0.11). A few quieter scenarios see modest increases; the deep-interior `max|w|` in both the no-fix and fix substep runs (10⁻² to 10⁻¹ m/s) sits ~1000× above explicit's noise floor (~10⁻⁵ m/s) regardless of this PR. This gap is largely a **design-regime mismatch**: the split-explicit substepper is designed for production NWP grids with `Δx, Δy >> Δz_min`, where the implicit vertical solve sidesteps a tight vertical-acoustic CFL and Klemp horizontal divergence damping + CN off-centering are tuned for `Δx ~ km` scales. Our 2.5D validation runs at `Δx ≈ 78 m`, `Δz = 50 m` (aspect ratio ≈ 1.6) with `Nz = 4` — outside that regime. The cyclostrophic-residual acoustic noise the substepper carries with the vortex is at scales the production-tuned damping doesn't target, and is independent of the mass-drift bug this PR addresses (which is regime-independent).

### 3-D confirmation — baroclinic-wave limited-area model

Beyond the 2-D vortex, the enforcement was exercised on a realistic 3-D problem. A global DCMIP2016 baroclinic wave (`CompressibleDynamics` + `AcousticRungeKutta3` + `SphericalCoriolis` on a `LatitudeLongitudeGrid`) was run as a parent, and its prognostic state over a sub-window was sampled to drive a *coincident* limited-area model (LAM) through `FieldTimeSeries`-backed open lateral boundaries (ρu/ρv `OpenBoundaryCondition`, ρ/ρθ `ValueBoundaryCondition`) plus a Davies relaxation zone. Two findings:

- The per-substep enforcement is **grid-agnostic** — it works unchanged on `LatitudeLongitudeGrid` (the kernels dispatch on BC classification + index ranges, not grid type).
- The LAM **preserves the parent solution through the full baroclinic life cycle** (days 6–30, spanning explosive cyclogenesis and wave breaking): deep-interior potential-temperature RMSE stays bounded at ~0.5–0.8 K (peaking ~1.4 K during the most violent breaking, then recovering), with no density drift. A LAM initialized instead from the undeveloped background jet spins the wave up purely from the boundary data and converges to the same ~0.6 K.

## Tests

Three testsets added to `test/acoustic_substepping.jl` (16/16 pass on CPU/Float64):

1. `open_boundary_relaxation` kwarg propagation + range validation.
2. No-op without active open BCs (periodic and all-walls bounded).
3. Outermost cell tracks prescribed ρ after a few substeps (end-to-end).

## Test plan

- [x] New testsets pass locally.
- [x] No-op path verified — no behavior change for periodic, walls, or `OpenBoundaryCondition(nothing)`.
- [x] Existing #716 regression test unaffected (the relaxation acts on Center scalar perturbations; the perturbation momentum's impenetrable halo is untouched).
- [x] CI green.

## Known limitation / follow-up — boundary momentum

This PR fixes the density (mass) drift. A separate, pre-existing **boundary-momentum** imbalance remains — intrinsic to the acoustic-substepping open-boundary treatment, and the more precise characterization of the substep `max|w|` gap noted above. A first-step diagnostic (LAM initialized to a coincident global solution, advanced one step at the global Δt, differenced against the global model) shows:

- The deep interior matches the global solution to roundoff; the difference is boundary-localized and almost entirely in the **momenta** at the boundary-adjacent cell (e.g. `ρv` ≈ 30× its field scale), with `ρ`/`ρθ` only ~1%.
- The **explicit** timestepper shows no such imbalance at the same (vertically stable) Δt, and the substep injection is ~Δt-independent — so it is specific to the per-substep boundary treatment, not the OBC, grid, or Δt.
- It is **not removable by tuning α**: the first-step momentum imbalance grows monotonically with α (the `ρ′/(ρθ)′` enforcement's boundary pressure perturbation is what kicks the adjacent momentum through the pressure gradient) — α is a trade-off knob, not a mitigation.
- A momentum analog of this enforcement (relaxing the boundary-face `(ρu)′/(ρv)′` toward 0) was prototyped and is a **no-op** — the boundary-face momentum perturbation is already ≈0, and the spurious signal is at the first *interior* cell — so it is intentionally **excluded** from this PR.

In practice this transient is controlled by the Davies / flow-relaxation zone (the standard operational treatment), which is why limited-area runs using these BCs are stable. Note that this PR's density-drift fix works by relaxing `ρ′/(ρθ)′` to re-couple the boundary acoustics — it works *around* the frozen `Gˢρ` rather than correcting that tendency — and the direct momentum analog of that relaxation is the no-op described above. A **distinct, unverified follow-up direction** is therefore to correct the **frozen stage-entry slow momentum tendency** `Gⁿρu` at the boundary-*adjacent* cell (a different variable and location than the boundary-value relaxation), which is where the first-step diagnostic localizes the overshoot.

### DIFF
--- a/src/CompressibleEquations/acoustic_substepping.jl
+++ b/src/CompressibleEquations/acoustic_substepping.jl
@@ -1310,21 +1310,22 @@ end
 
 @inline is_active_open_bc(bc) = (bc isa BoundaryCondition{<:Open}) && !(bc.condition isa Nothing)
 
-# Relax a Center perturbation toward the prescribed wall value in the outermost
-# open-boundary cell row: `target = v − ρᴸ[cell] = (base[halo] − base[cell]) / 2`.
-@kernel function _relax_open_boundary_x_cell!(pert, base, i_cell, i_halo, α)
+# Relax ρ′ and (ρθ)′ at the outermost open-boundary cell toward the prescribed
+# wall value in a single kernel: target = v − cᴸ[iᴮ] = (cᴸ[iᴴ] − cᴸ[iᴮ]) / 2.
+# `iᴮ` is the outermost interior cell index, `iᴴ` the adjacent halo cell index.
+@kernel function _relax_open_boundary_x!(ρ′, ρθ′, ρᴸ, ρθᴸ, iᴮ, iᴴ, α)
     j, k = @index(Global, NTuple)
     @inbounds begin
-        target = (base[i_halo, j, k] - base[i_cell, j, k]) / 2
-        pert[i_cell, j, k] += α * (target - pert[i_cell, j, k])
+        ρ′[iᴮ, j, k]  += α * ((ρᴸ[iᴴ, j, k]  - ρᴸ[iᴮ, j, k])  / 2 - ρ′[iᴮ, j, k])
+        ρθ′[iᴮ, j, k] += α * ((ρθᴸ[iᴴ, j, k] - ρθᴸ[iᴮ, j, k]) / 2 - ρθ′[iᴮ, j, k])
     end
 end
 
-@kernel function _relax_open_boundary_y_cell!(pert, base, j_cell, j_halo, α)
+@kernel function _relax_open_boundary_y!(ρ′, ρθ′, ρᴸ, ρθᴸ, jᴮ, jᴴ, α)
     i, k = @index(Global, NTuple)
     @inbounds begin
-        target = (base[i, j_halo, k] - base[i, j_cell, k]) / 2
-        pert[i, j_cell, k] += α * (target - pert[i, j_cell, k])
+        ρ′[i, jᴮ, k]  += α * ((ρᴸ[i, jᴴ, k]  - ρᴸ[i, jᴮ, k])  / 2 - ρ′[i, jᴮ, k])
+        ρθ′[i, jᴮ, k] += α * ((ρθᴸ[i, jᴴ, k] - ρθᴸ[i, jᴮ, k]) / 2 - ρθ′[i, jᴮ, k])
     end
 end
 
@@ -1338,20 +1339,16 @@ function apply_open_boundary_relaxation!(substepper, model, grid, arch)
     ρᴸ  = model.dynamics.density
     ρθᴸ = thermodynamic_density(model.formulation)
     if is_active_open_bc(bcs_u.west)
-        launch!(arch, grid, :yz, _relax_open_boundary_x_cell!, ρ′,  ρᴸ,  1, 0, α)
-        launch!(arch, grid, :yz, _relax_open_boundary_x_cell!, ρθ′, ρθᴸ, 1, 0, α)
+        launch!(arch, grid, :yz, _relax_open_boundary_x!, ρ′, ρθ′, ρᴸ, ρθᴸ, 1, 0, α)
     end
     if is_active_open_bc(bcs_u.east)
-        launch!(arch, grid, :yz, _relax_open_boundary_x_cell!, ρ′,  ρᴸ,  Nx, Nx + 1, α)
-        launch!(arch, grid, :yz, _relax_open_boundary_x_cell!, ρθ′, ρθᴸ, Nx, Nx + 1, α)
+        launch!(arch, grid, :yz, _relax_open_boundary_x!, ρ′, ρθ′, ρᴸ, ρθᴸ, Nx, Nx + 1, α)
     end
     if is_active_open_bc(bcs_v.south)
-        launch!(arch, grid, :xz, _relax_open_boundary_y_cell!, ρ′,  ρᴸ,  1, 0, α)
-        launch!(arch, grid, :xz, _relax_open_boundary_y_cell!, ρθ′, ρθᴸ, 1, 0, α)
+        launch!(arch, grid, :xz, _relax_open_boundary_y!, ρ′, ρθ′, ρᴸ, ρθᴸ, 1, 0, α)
     end
     if is_active_open_bc(bcs_v.north)
-        launch!(arch, grid, :xz, _relax_open_boundary_y_cell!, ρ′,  ρᴸ,  Ny, Ny + 1, α)
-        launch!(arch, grid, :xz, _relax_open_boundary_y_cell!, ρθ′, ρθᴸ, Ny, Ny + 1, α)
+        launch!(arch, grid, :xz, _relax_open_boundary_y!, ρ′, ρθ′, ρᴸ, ρθᴸ, Ny, Ny + 1, α)
     end
     return nothing
 end
--- a/src/CompressibleEquations/acoustic_substepping.jl
+++ b/src/CompressibleEquations/acoustic_substepping.jl
@@ -81,7 +81,7 @@ using Oceananigans.Operators:
     Axᶠᶜᶜ, Ayᶜᶠᶜ, Vᶜᶜᶜ
 
 using Oceananigans.Utils: launch!
-using Oceananigans.BoundaryConditions: fill_halo_regions!
+using Oceananigans.BoundaryConditions: fill_halo_regions!, BoundaryCondition, Open
 
 using Oceananigans.Grids: Flat, Center, peripheral_node,
                           topology,
@@ -173,6 +173,9 @@ struct AcousticSubstepper{N, FT, D, AD, US, CF, MP, TAV, GT, TS}
     damping :: D
     substep_distribution :: AD
     sponge :: US
+    # Per-substep relaxation factor `α` for the outermost open-boundary cell
+    # of `ρ′,(ρθ)′`. See `apply_open_boundary_relaxation!` and issue #738.
+    open_boundary_relaxation :: FT
 
     # Linearization basic state ``Uᴸ`` — Πᴸ and θᴸ derived from the live
     # `model.dynamics.pressure`, `model.dynamics.density`, and the
@@ -226,6 +229,7 @@ Adapt.adapt_structure(to, a::AcousticSubstepper) =
                        adapt(to, a.damping),
                        a.substep_distribution,
                        adapt(to, a.sponge),
+                       a.open_boundary_relaxation,
                        adapt(to, a.linearization_exner),
                        adapt(to, a.linearization_potential_temperature),
                        adapt(to, a.linearization_gamma_R_mixture),
@@ -276,6 +280,7 @@ function AcousticSubstepper(grid, split_explicit::SplitExplicitTimeDiscretizatio
     damping = split_explicit.damping
     sponge = split_explicit.sponge
     substep_distribution = split_explicit.substep_distribution
+    open_boundary_relaxation = convert(FT, split_explicit.open_boundary_relaxation)
 
     # Linearization basic state — Πᴸ, θᴸ derived from live model fields.
     linearization_exner                           = CenterField(grid)
@@ -313,7 +318,7 @@ function AcousticSubstepper(grid, split_explicit::SplitExplicitTimeDiscretizatio
                                                tridiagonal_direction = ZDirection())
 
     return AcousticSubstepper(Ns, acoustic_cfl, ω, damping, substep_distribution,
-                              sponge,
+                              sponge, open_boundary_relaxation,
                               linearization_exner,
                               linearization_potential_temperature,
                               linearization_gamma_R_mixture,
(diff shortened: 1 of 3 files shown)


ewquon:
@kaiyuan-cheng can you confirm that FV3 does boundary relaxation in the fast timesteps (inside `fv_regional_bc.F90` as noted above)? Thanks!

kaiyuan-cheng:
> @kaiyuan-cheng can you confirm that FV3 does boundary relaxation in the fast timesteps (inside `fv_regional_bc.F90` as noted above)? Thanks!

Yes, it uses blending every dynamics timestep.
https://github.com/NOAA-GFDL/GFDL_atmos_cubed_sphere/blob/6f658bd00ffa52158c1c0f32244b6dd59a0360d7/model/dyn_core.F90#L466

src/CompressibleEquations/acoustic_substepping.jl
```diff
@@ -173,6 +173,9 @@ struct AcousticSubstepper{N, FT, D, AD, US, CF, MP, TAV, GT, TS}
     damping :: D
     substep_distribution :: AD
     sponge :: US
+    # Per-substep relaxation factor `α` for the outermost open-boundary cell
+    # of `ρ′,(ρθ)′`. See `apply_open_boundary_relaxation!` and issue #738.
+    open_boundary_relaxation :: FT
```
glwagner, line 178: Another possibility is to implement this as a `scheme` of `Open`:

https://github.com/CliMA/Oceananigans.jl/blob/901b2c046d04557112fa8edfc53ce6174014f251/src/BoundaryConditions/boundary_condition_classifications.jl#L70-L72

For example this would permit things like: different relaxation on different fields, or modular development of new schemes other than this straightforward relaxation.

I would call the scheme something like `AcousticRelaxation`.

src/CompressibleEquations/acoustic_substepping.jl
```diff
@@ -1281,6 +1286,76 @@ end
...
+        launch!(arch, grid, :yz, _relax_open_boundary_x_cell!, ρ′,  ρᴸ,  Nx, Nx + 1, α)
+        launch!(arch, grid, :yz, _relax_open_boundary_x_cell!, ρθ′, ρθᴸ, Nx, Nx + 1, α)
+    end
+    if is_active_open_bc(bcs_v.south)
+        launch!(arch, grid, :xz, _relax_open_boundary_y_cell!, ρ′,  ρᴸ,  1, 0, α)
+        launch!(arch, grid, :xz, _relax_open_boundary_y_cell!, ρθ′, ρθᴸ, 1, 0, α)
+    end
+    if is_active_open_bc(bcs_v.north)
+        launch!(arch, grid, :xz, _relax_open_boundary_y_cell!, ρ′,  ρᴸ,  Ny, Ny + 1, α)
+        launch!(arch, grid, :xz, _relax_open_boundary_y_cell!, ρθ′, ρθᴸ, Ny, Ny + 1, α)
```
glwagner: Probably best to combine these two launches into one kernel (best to minimize the number of kernels)

src/CompressibleEquations/acoustic_substepping.jl
```diff
@@ -1281,6 +1286,76 @@ end
...
+# how hard the cell is pulled each substep. The relaxation is a no-op on any
+# side whose prognostic-momentum BC is not an active `OpenBoundaryCondition`
+# (periodic, walls, and `OpenBoundaryCondition(nothing)` all skip it), so the
+# enforcement has zero cost when no open lateral BC is present.
+
+@inline is_active_open_bc(bc) = (bc isa BoundaryCondition{<:Open}) && !(bc.condition isa Nothing)
+
+# Relax a Center perturbation toward the prescribed wall value in the outermost
+# open-boundary cell row: `target = v − ρᴸ[cell] = (base[halo] − base[cell]) / 2`.
+@kernel function _relax_open_boundary_x_cell!(pert, base, i_cell, i_halo, α)
```
```suggestion
@kernel function _relax_open_boundary_x_cell!(c′, cᴸ, i_cell, i_halo, α)
```

notation like this might be easier to read (also maybe best to come up with notation for `_cell` and `_halo` to avoid mixing math and english)

src/CompressibleEquations/acoustic_substepping.jl
glwagner: You could then follow the style of `PerturbationAdvection` and develop an alias + constructor to support syntax like:

```julia
obc = AcousticRelaxationBoundaryCondition(condition; scheme_kw...)
```

ewquon: This would require a small Oceananigans change (a few LOC), but I think it would make sense to implement
```julia
struct Value{MS} <: AbstractBoundaryConditionClassification
    scheme::MS
end
```
This allows users to opt into using the new `AcousticRelaxation` scheme but not have to change from a `ValueBoundaryCondition` (e.g., on ρθ) to an `OpenBoundaryCondition`, which strictly speaking, should only apply to boundary-normal components of velocity. Thoughts?

glwagner: Ah I see the issue, because we need to apply a scheme to tracers too. 

Since its a deeper change perhaps we should defer this discussion to an issue and go with the hard-coded interface you've developed for this PR?

ewquon: I'm fine with moving forward and opening an issue for future enhancement.

glwagner: lets do it

ewquon:
@glwagner — I made the requested fixes and opened a follow-on issue

ewquon:
Full system test: baroclinic wave. A global 30-day simulation provides the parent solution. LAM calculations with this PR were started from day 6, after waves had been triggered by the instability. The LAM solution was initialized in two ways: directly from the parent and to quiescent background conditions. Dashed lines indicate the interior edge of the Davies relaxation zone.

Notes:
* Large-scale flow features are maintained throughout the duration of the simulation.
* The momentum residual issue (described in the "Known limitation / follow up" section fo the PR body) is clearly seen in the first row of inflow cells. The same issue is also seen adjacent to the outflow boundary, resulting in feedback to the interior of the domain.
* Somewhat surprisingly, the LAM initialized to background flow does not completely "forget" its IC, resulting in subtle differences between the two LAM simulations.
 
https://github.com/user-attachments/assets/02214512-d74f-40b7-8655-3feecdfcd6b9

glwagner (approved):

PR #752, opened 2026-05-30
Fix moist temperature inversion in compressible equation of state

kaiyuan-cheng:
## Summary

The fully compressible core inverts liquid-ice potential temperature → temperature
inconsistently when condensate is present, under-heating moist parcels and giving
a too-weak moist updraft in the splitting supercell. The anelastic core inverts the
same θₗᵢ definition exactly, so the two cores agree while dry and diverge at
condensation onset.

## Root cause

`temperature_and_pressure` (`src/CompressibleEquations/compressible_time_stepping.jl`)
computed the dry-adiabatic temperature from density,

    T_dry = θₗᵢ^γ (ρ Rᵐ / pˢᵗ)^(γ-1)

then added the latent-heat correction `(ℒˡ qˡ + ℒⁱ qⁱ)/cᵖᵐ`. The dry inversion
evaluates the Exner function at the *dry* pressure `ρ Rᵐ T_dry` instead of the
actual pressure `ρ Rᵐ T`, so condensate-laden cells are under-heated by
≈ `(γ-1)·ℒqˡ/cᵖᵐ` (≈4.5 K at 5 g/kg, ≈7 K at 8 g/kg). The cooler, denser air has
roughly half the buoyancy.

A round-trip check (form θₗᵢ from a known `(T, p, q)` via Breeze's canonical
definition, then recover `T`) confirms the anelastic core recovers `T` exactly
while the old compressible formula is off by several K once condensate appears.

## Fix

Solve the consistent coupled `(T, p)` system. With `p = ρ Rᵐ T` and
`θₗᵢ = (T − L)(pˢᵗ/p)^κ`, the temperature is the fixed point `T = D·T^κ + L`,
`D ≡ θₗᵢ(ρ Rᵐ/pˢᵗ)^κ`, iterated to convergence (contraction rate ≈ κ ≈ 0.28).
Dry/rest cells (`L = 0`) return the dry-adiabatic temperature bit-for-bit, so the
exact discrete rest atmosphere is preserved. The terrain compressible path shares
the same kernel and is fixed too.

## Verification

Verified with a **saturation-adjustment** variant of the splitting-supercell setup,
which isolates the effect cleanly (no rain / sedimentation / evaporation): the
compressible core's max|w| at t = 15 min was **2.6× weaker** than the anelastic
core (**11.4 vs 28.6 m/s**) and rises to **27.7 m/s** with the fix, matching the
anelastic core. The example's **Kessler** scheme exhibits the same bug; the fix
brings the two cores substantially closer there too, with a smaller residual in
the explosive phase where both cores overshoot independently of this fix.

- `test/substepper_rest_state.jl` (11/11) and `test/substepper_structural.jl`
  (22/22) still pass — the rest state is unchanged.
- New regression test "Compressible moist temperature inversion consistency"
  asserts the diagnosed temperature equals the canonical θₗᵢ inversion for a
  saturated compressible state (off by several K before this fix).

### DIFF
--- a/src/CompressibleEquations/compressible_dynamics.jl
+++ b/src/CompressibleEquations/compressible_dynamics.jl
@@ -19,6 +19,8 @@ Fields
 - `terrain_metrics`: [`TerrainMetrics`](@ref) for terrain-following coordinates (or `nothing`)
 - `Ω̃`, `ρΩ̃`: contravariant vertical velocity / momentum diagnostic fields (or `nothing` when no terrain metrics)
 - `terrain_reference_pressure`, `terrain_reference_density`: 3D reference pressure / density for the terrain pressure gradient force (or `nothing`)
+- `temperature_tolerance`, `temperature_maxiter`: relative convergence tolerance on the moist
+  equation-of-state temperature iteration step `|ΔT|/T`, and the iteration cap on that fixed-point loop
 
 The `time_discretization` determines how tendencies are computed and which
 time-stepper is used:
@@ -37,6 +39,8 @@ struct CompressibleDynamics{TD, D, P, FT, RS, TM, CV, CM, TRP, TRD}
     contravariant_vertical_momentum :: CM      # ρΩ̃ diagnostic field (or Nothing)
     terrain_reference_pressure :: TRP          # 3D reference pressure for terrain PG (or Nothing)
     terrain_reference_density :: TRD           # 3D reference density for terrain buoyancy (or Nothing)
+    temperature_tolerance :: FT                # relative convergence tol on |ΔT|/T for the moist EOS inversion
+    temperature_maxiter :: Int                 # iteration cap for the moist EOS temperature fixed-point loop
 end
 
 """
@@ -60,13 +64,19 @@ Keyword Arguments
   hydrostatically-balanced reference state used in base-state subtraction. Can be a constant `θ₀`
   or a function `θ(z)`. Default: `nothing` (no base-state correction).
   When provided, an [`ExnerReferenceState`](@ref) is built during materialization.
+- `temperature_tolerance`: relative convergence tolerance on the moist EOS temperature
+  iteration step `|ΔT|/T` (default: `1e-8`)
+- `temperature_maxiter`: maximum number of moist EOS temperature fixed-point iterations
+  (default: `8`, matching the prior hardcoded loop length)
 """
 function CompressibleDynamics(time_discretization::TD = ExplicitTimeStepping();
                               standard_pressure = 1e5,
                               surface_pressure = 101325.0,
                               reference_potential_temperature = nothing,
                               reference_temperature = nothing,
-                              terrain_metrics = nothing) where TD
+                              terrain_metrics = nothing,
+                              temperature_tolerance = 1e-8,
+                              temperature_maxiter = 8) where TD
 
     FT = promote_type(typeof(standard_pressure), typeof(surface_pressure))
     pˢᵗ = convert(FT, standard_pressure)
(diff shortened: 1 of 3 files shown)


src/CompressibleEquations/compressible_time_stepping.jl
```diff
@@ -133,17 +133,36 @@ end
...
+    # a buoyancy inconsistent with the anelastic core (which inverts the same θ
+    # definition exactly). Solving the coupled (T, p) system gives the fixed point
+    #   T = D T^κ + L ,   D ≡ θ (ρ Rᵐ / pˢᵗ)^κ ,
+    # which we iterate to convergence (contraction rate ≈ κ ≈ 0.28). For dry/rest
+    # cells (L = 0) the result is `T_dry` bit-for-bit, preserving the exact
+    # discrete rest atmosphere.
+    L = (ℒˡᵣ * qˡ + ℒⁱᵣ * qⁱ) / cᵖᵐ
+    D = θ * (ρ * Rᵐ / pˢᵗ)^κ
+    T = T_dry + L
+    for _ in 1:8
```
glwagner: Let's stop the business of hard coding solvers!

kaiyuan-cheng: Agreed. Admittedly, I haven't taken a close look at the implementation details yet. That's why this is still a draft. The end result looks good though.

glwagner: we can put some guidance in `AGENTS.md` -- what do you think?

kaiyuan-cheng: Sounds good to me.
It looks like there's another hardcoded iteration
https://github.com/NumericalEarth/Breeze.jl/blob/7b2a209e89ddc2cb05af2db4e45a4c6474550279/src/Thermodynamics/reference_states.jl#L556

@glwagner Do we have an existing solution for generalizing the task like this?

glwagner: No, there is nothing formal. Saturation adjustment uses `tolerance` and `maxiter` properties:

https://github.com/NumericalEarth/Breeze.jl/blob/7b2a209e89ddc2cb05af2db4e45a4c6474550279/src/Microphysics/saturation_adjustment.jl#L208-L214

NumericalEarth has a similar thing for a fixed point solver for surface fluxes.

There are GPU compatible root solvers in the julia ecosystem we could use. But I think the most important thing is just to expose the solver parameters to the user.

src/CompressibleEquations/compressible_dynamics.jl
```diff
@@ -37,6 +39,8 @@ struct CompressibleDynamics{TD, D, P, FT, RS, TM, CV, CM, TRP, TRD}
     contravariant_vertical_momentum :: CM      # ρΩ̃ diagnostic field (or Nothing)
     terrain_reference_pressure :: TRP          # 3D reference pressure for terrain PG (or Nothing)
     terrain_reference_density :: TRD           # 3D reference density for terrain buoyancy (or Nothing)
+    temperature_tolerance :: FT                # relative convergence tol on |ΔT|/T for the moist EOS inversion
+    temperature_maxiter :: Int                 # iteration cap for the moist EOS temperature fixed-point loop
```
glwagner, line 43: I think we should develop a type for this --- we need a property like

```julia
equation_of_state_inversion
```

or something similar

Then design `struct` that capture this. For example, we can support existing behavior `DryInversion` and a new case

```julia
struct FixedPointIteration
    tolerance
    maxiter
end
```

(sketching this out)

ewquon:
Superceded by #767

PR #762, opened 2026-06-01
Apply operator-split microphysics once per time step

kaiyuan-cheng:
`DCMIP2016KesslerMicrophysics` updates state directly inside
`microphysics_model_update!` (it bypasses the tendency interface). That hook ran
once per RK stage from `update_state!` **plus** once from a separate explicit
per-step call in both time-steppers — so the full-`Δt` operator-split update was
applied **4× per step**, with the extra calls reading a stale diagnostic
`θ`/pressure.

Over-applying the irreversible precipitation cascade (autoconversion, accretion,
sedimentation) strips condensate from the updraft too fast, removing the
water-loading brake and over-intensifying the storm. In
`examples/splitting_supercell.jl` the anelastic peak `max w` drops from ~117 m/s
to ~79 m/s with the fix (compressible ~108 → ~74), both cores in close agreement.

## Fix

- Gate the operator-split update behind a new `apply_microphysics_model_update`
  keyword on `update_state!` (default `false`); the SSP and acoustic RK3 steppers
  set it `true` on the final post-RK `update_state!` only — applied once per step,
  after `compute_auxiliary_variables!` refreshes the state it reads.
- Remove the misplaced explicit per-step hooks and their now-unused imports.

The tendency-interface microphysics (`compute_microphysical_tendencies!`) is
unchanged and still runs every stage; `set!`/init use the default `false`, and
`ParcelModels` (which calls the hook directly) is unaffected.

Validated against `examples/splitting_supercell.jl` (numbers above).


src/AtmosphereModels/update_atmosphere_model_state.jl
```diff
@@ -30,7 +30,24 @@ function Oceananigans.Models.update_model_field_time_series!(model::AtmosphereMo
...
+# calls) do NOT advance the operator-split update. The time-steppers pass
+# `apply_microphysics_model_update=true` exactly once per step — on the final post-RK
+# `update_state!`, after `compute_auxiliary_variables!` has refreshed the diagnostic
+# `θ`, pressure, and moisture that the update reads. This guarantees a scheme like
+# `DCMIP2016KesslerMicrophysics` (which bypasses the tendency interface and mutates
+# state directly) is applied exactly once per step, on a consistent state. Because that
+# update writes prognostic fields in the interior only, halos and the diagnostic
+# variables are recomputed afterwards (via `finalize_microphysics_model_update!`) so
+# the tendencies below — which the next step's first RK substep consumes — reflect the
+# post-update state rather than stale pre-update halos and diagnostics.
```
glwagner: I think this comment shoudl be simplified and should not refer to specific microphysics schemes

src/AtmosphereModels/update_atmosphere_model_state.jl
```diff
@@ -41,7 +58,12 @@ function TimeSteppers.update_state!(model::AtmosphereModel, callbacks=[]; comput
     update_boundary_conditions!(prognostic_fields(model), model)
     update_radiation!(model.radiation, model)
     compute_forcings!(model)
-    microphysics_model_update!(model.microphysics, model)
+    apply_microphysics_model_update && microphysics_model_update!(model.microphysics, model)
```
glwagner: would another strategy be to move `microphysics_model_update!` out of `update_state!` and into `time_step!` directly?

I think we want to think of `update_state!` as "idempotent", ie it can be called as many times as needed.

Otherwise there are complications, such as this kwarg, and also for checkpointing.

test/dcmip2016_kessler.jl
```diff
@@ -362,7 +362,10 @@ end
 
     set!(model.formulation.potential_temperature_density, reshape(ρ_prof .* θˡⁱ_init, 1, 1, Nz))
     model.clock.last_Δt = Δt
-    update_state!(model)
+    # Apply the operator-split Kessler update once, as the time-steppers do on the
+    # final post-RK `update_state!`; the default `apply_microphysics_model_update=false`
+    # leaves the prognostic state untouched.
+    update_state!(model; apply_microphysics_model_update=true)
```
glwagner:
```suggestion
    microphysics_model_update!(model.microphysics, model)
    update_state!(model)
```

not sure if this is another pattern that might work, but it could be simpler.

glwagner: I think it means that we need to fill halo regions inside the microphysics update. But `update_state!` can be left unchanged

kaiyuan-cheng:
@glwagner `update_state!` is idempotent again. Dropped the `apply_microphysics_model_update` kwarg, the finalize hook, and `recompute_auxiliary_state!`. The RK3 steppers now call `update_state!` then `microphysics_model_update!` once per step, and the Kessler update restores state itself via `update_state!`.

glwagner (approved):

PR #764, opened 2026-06-02
Add `adiabatic_balance!` for compressible and anelastic dynamics

### LINKED ISSUE #741: FV3's adiabatic balancing for cold-start simulation
## Purpose

Analyses (ERA5, GFS) provide the atmospheric fields but **no balanced nonhydrostatic state** — `w` is cold-started at zero. FV3's adiabatic initialization `na_init` spins them up into balance with the atmospheric fields via a short symmetric adiabatic integration.

## Algorithm

Let `D(±Δt)` be one sign-aware adiabatic dynamics step (no physics/IAU/diffusion).
For `m = 1…N_init`:

```
D(+Δt); D(-Δt); nudge      # half-cycle A:  0 → +Δt → 0
D(-Δt); D(+Δt); nudge      # half-cycle B:  0 → -Δt → 0
```

Nudge of the atmospheric fields toward the IC (`x₀`) to prevent them from drifting too far from the initial conditions, with `wₜ=2`, `xₜ=1/3`:

$$x \leftarrow x_t\,x_\text{dyn} + (1-x_t)\,x_0 = \tfrac13 x_\text{dyn} + \tfrac23 x_0$$

**`w` is never nudged** — whatever balance the excursion imprints on them survives. That is the point: slow fields anchored, NH fields free to find
their balance.

kaiyuan-cheng:
## Summary

Adds `adiabatic_balance!(model; Δt, cycles = 1, weight = 2)` — FV3-style `na_init` adiabatic initialization for `AtmosphereModel`.

It spins up a balanced vertical momentum `ρw` (and the nonhydrostatic pressure balance) consistent with the model's analysis initial state. Each of `cycles` iterations runs a symmetric forward/backward dynamics excursion — which lets `ρw` develop — then nudges the *initial fields* back toward their `t = 0` snapshot via `x ← (x + weight·x₀)/(1 + weight)` (default `weight = 2` → ⅓ dynamics + ⅔ IC). `ρw` is never snapshotted or nudged, so the balance the excursion imprints on it is kept; the clock is reset to `t = 0` on exit.

- **Dynamics-agnostic.** Dispatches on `AtmosphereModel`; the nudged set is a per-dynamics `initial_fields` method:
  - `CompressibleModel` → `(ρ, ρu, ρv, ρθ, ρqᵉ)`
  - `AnelasticModel` → `(ρu, ρv, ρθ, ρqᵉ)` — drops `ρ`, which is the fixed anelastic reference `ρᵣ(z)`, not prognostic.

  Adding another dynamics is one new `initial_fields` method.
- **Adiabatic-by-contract.** The caller must pass a model built with `microphysics = nothing`, `sponge = nothing`, and no forcing (these run inside `update_state!`/`time_step!` and would corrupt the spin-up). BCs are not modified — use time-invariant boundaries so the ±Δt excursion stays nearly reversible.
- Lives in top-level `src/adiabatic_balance.jl`, exported from `Breeze`.

## Test plan

- [x] `test/adiabatic_balance.jl` (16 tests, CPU, run locally): nudge algebra (blend exact, `ρw` untouched); compressible rest-state invariance + clock reset; `ρw` spin-up (seeded `ρw` shrinks across a cycle); **anelastic** rest-state invariance — which also exercises the backward `time_step!(model, -Δt)` on the anelastic stepper.
- [ ] CI: full suite, Float32, GPU.

Part of NUM-168


kaiyuan-cheng:
close #741

src/adiabatic_initialization.jl
```diff
@@ -0,0 +1,98 @@
...
+Spin up a balanced vertical momentum `ρw` (and the nonhydrostatic pressure
+balance) consistent with a `model`'s initial (analysis) state, via FV3
+adiabatic initialization (`na_init`).
+
+Analyses (ERA5, GFS, …) supply the density, momentum, and thermodynamic state but
+cold-start the vertical velocity `w` at zero (hydrostatic), so the vertical /
+nonhydrostatic state is out of balance with the rest. For each of `cycles`
+iterations the routine runs a symmetric forward/backward dynamics excursion —
+which lets `ρw` develop — and then nudges the *initial fields* back toward their
+`t = 0` snapshot by the weighted mean
```
ewquon:
```suggestion
nonhydrostatic state is out of balance with the rest. Each cycle of `cycles`
entails two half-cycles: a symmetric forward/backward dynamics excursion
followed by a symmetric backward/forward excursion, both at the same `Δt`.
Following each excursion — which lets `ρw` develop — the routine nudges
the *initial fields* back toward their `t = 0` snapshot by the weighted mean
```

src/adiabatic_initialization.jl
```diff
@@ -0,0 +1,98 @@
...
+
+    x ← (x + weight·x₀) / (1 + weight)
+
+(default `weight = 2` → ⅓ dynamics + ⅔ snapshot). The nudge keeps the
+initialization from drifting too far from the initial conditions. The vertical momentum `ρw` is
+never snapshotted or nudged, so the balance the excursion imprints on it is
+exactly what is kept: with the initial fields held to the analysis and the
+vertical field free, `ρw` relaxes into balance with them.
+
+The initial fields depend on the dynamics:
```
ewquon:
```suggestion
The initial fields that are modified depend on the dynamics:
```

glwagner:
should it be `initialize_adiabatically!`    ?

One idea is, "functions are verbs" and "objects are nouns". We don't universally obey this, but it can provide clarity to use the imperative for a `!` function

I am also not sure this should truly be thought of as "initialization". Yes, it is useful for initialization, but it is more correctly or mundanely some kind of state transformation. What exactly does this function do? I guess it finds the closest balanced state from some unbalanced condition (where "balance" refers to slow vs fast balances in the equations of motion)

src/adiabatic_initialization.jl
```diff
@@ -0,0 +1,98 @@
...
+        # Half-cycle B: 0 → -Δt → 0, then nudge.
+        time_step!(model, -Δt)
+        time_step!(model, +Δt)
+        nudge_initial_fields!(model, snapshot, weight)
+        update_state!(model)
+    end
+
+    # Production integration begins from t = 0.
+    model.clock.time = zero(model.clock.time)
+    model.clock.iteration = 0
```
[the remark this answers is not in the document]
ewquon: Might be better to unit test this than to explicit test? After cycling, I think `model.clock.stage == 1` and `!isinf(model.clock.last_Δt)`

ewquon (approved):
Looks good, only a few minor suggestions from me. I don't feel very strongly about the guardrails that codex is trying to build in. One guardrail that might be helpful is a check for whether or not the BCs and forcing are fixed and empty, respectively, per your assumptions.

ewquon:
To @glwagner's point, I might call this a fixed-point iteration (with under-relaxation)

kaiyuan-cheng:
@ewquon @glwagner Yes, `adiabatic_initialization` names the method, but not the purpose. As the purpose is to do non-hydrostatic adjustment, how about `initialize_nonhydrostatic_balance`?

glwagner:
I think we should avoid the term "initialize" because the connection with initialization is only incidental

ewquon:
spitballing here
* `adiabatic_balance!`
* `balance_vertical_momentum!`
* `balance_vertical_dynamics!`

glwagner:
I like `balance_vertical_dynamics` or `vertical_momentum` if it really does only affect the vertical momentum, that is nice and precise.

kaiyuan-cheng:
> I like `balance_vertical_dynamics` or `vertical_momentum` if it really does only affect the vertical momentum, that is nice and precise.

While the goal is to spin up vertical momentum, thermodynamic fields and horizontal momentums also get adjusted. For this reason, I think `adiabatic_balance` is more precise.

glwagner:
@kaiyuan-cheng would you be ok with a change to `balance_adiabatically!` so its clear this is a verb? The idea is to read this way:

```julia
balance_adiabatically!(model)
```

in contrast for a name like "adaibatic_balance" I would expect this syntax

```julia
balanced_state = adiabatic_balance(model)
```

in other words, such a function would compute the balanced state and return it. 

No big deal either way, just my thought.

PR #766, opened 2026-06-04
Add Oceananigans compat for 0.110

navidcy:
Updated Oceananigans version range in Project.toml


navidcy:
hm.... on second thought, there might be breaking changes with Open Boundary conditions that need to be addressed and then possibly compat for 0.109 dropped? @glwagner?

giordano (approved):
Looks like it worked? Or are we missing tests?

navidcy:
It wasn't enforcing v0.110 properly; when done all hell broke loose :) 

Closing in favour of #770

PR #767, opened 2026-06-04
Density-based θˡⁱ state: fix compressible saturation & temperature inversion (closes #765, supersedes #752)

### LINKED ISSUE #765: Compressible θˡⁱ thermodynamic state uses a pressure coordinate — one root cause behind both the temperature-inversion inconsistency (#752 / #625-6) and a saturation adjustment evaluated at the wrong pressure
/cc @kaiyuan-cheng — this touches the thermodynamics core and changes the direction of the in-flight [#752](<https://github.com/NumericalEarth/Breeze.jl/issues/752>), so I'd value your review of the framing and the proposed fix before anyone implements it.

> **Context.** This surfaced while building a downstream compressible-θˡⁱ moist configuration, but the bug is entirely in **committed** Breeze thermodynamics — both symptoms reproduce with the two self-contained scripts below (pure `Breeze.Thermodynamics` / `Breeze.Microphysics`, no model and no external code). File:line refs are on `eq/substep-open-bc-enforcement`; the relevant lines are unchanged from `main`.

## TL;DR

The compressible `LiquidIcePotentialTemperatureFormulation` represents its thermodynamic state in a **pressure** coordinate (`LiquidIcePotentialTemperatureState` carries `reference_pressure`, with density *derived*), even though the compressible core's true prognostic coordinate is **density** `ρ`. That single mismatch produces **two** distinct problems:

1. **Temperature-inversion inconsistency** (NumericalEarth/Breeze.jl#625 item 6, draft PR [#752](<https://github.com/NumericalEarth/Breeze.jl/issues/752>)): two `T` diagnostics differ by `≈ (Rᵐ/cᵖᵐ)·ΔL` whenever condensate is present.
2. **Saturation adjustment evaluated at the wrong pressure**: `SaturationAdjustment` evaluates `qsat` (and the Exner inversion) at the state's `reference_pressure`, held fixed through its secant. On the anelastic core that *is* the fixed EOS pressure `p̄(z)`, so it's exact; on the compressible core the cell's pressure is `p = ρRᵐT` at the prognostic `ρ`, so saturation is computed against a pressure inconsistent with the cell's density. (Both paths conserve θˡⁱ — this is *not* an energy / cₚ-vs-cᵥ issue.)

These need **different** fixes — PR [#752](<https://github.com/NumericalEarth/Breeze.jl/issues/752>) (iterate the inversion) does nothing for symptom 2, and a density-consistent saturation solve does nothing for the inversion seen by other schemes — but they share **one root**. A single change (a density-coordinate compressible state + a density-consistent `adjust_thermodynamic_state`) closes both for *every* moist scheme.

---

## The shared root cause

The compressible θˡⁱ state is built in `src/PotentialTemperatureFormulations/potential_temperature_formulation.jl:88-91`:

```julia
pᵣ = dynamics_pressure(dynamics)[i,j,k]          # the stored pressure
LiquidIcePotentialTemperatureState(θ, q, pˢᵗ, pᵣ)
```

`LiquidIcePotentialTemperatureState` (`src/Thermodynamics/dynamic_states.jl:22-27`) carries `reference_pressure`, and everything is derived from it:

* `temperature(𝒰)` is the **Exner / constant-pressure** inversion `Πθ + ΔL`, `Π = (pᵣ/pˢᵗ)^κ` (`dynamic_states.jl:46-58`).
* `density(𝒰) = pᵣ/(RᵐT)` — density is *diagnosed from* `pᵣ` (`dynamic_states.jl:143-148`), not carried.

This is correct for the **anelastic** core, where `dynamics_pressure` is the reference pressure (the true independent variable). It is wrong for the **compressible** core, where `ρ` is prognostic and `p` floats. The two symptoms are the two consequences.

---

## Symptom 1 — temperature inversion (recap of [#752](<https://github.com/NumericalEarth/Breeze.jl/issues/752>) / NumericalEarth/Breeze.jl#625-6)

The compressible dynamics compute `T` density-based but non-iterated (`src/CompressibleEquations/compressible_time_stepping.jl`, \~145-149):

```
T_A = θ^γ (ρRᵐ/pˢᵗ)^(γ-1) + ΔL ;   p = ρRᵐ T_A      # stored pressure
```

while everything else reports the Exner form at that stored `p` (`src/AtmosphereModels/update_atmosphere_model_state.jl:271`):

```
T_B = (p/pˢᵗ)^κ θ + ΔL
```

`T_A` and `T_B` are the first two Picard iterates of the self-consistent fixed point `T* = (ρRᵐT*/pˢᵗ)^κ θ + ΔL`, so they differ by `≈ κ·ΔL` whenever `qˡ+qⁱ > 0` (both biased cold vs `T*`).

**Reproducer 1** (pure `Breeze.Thermodynamics`):

```julia
using Breeze.Thermodynamics: ThermodynamicConstants, MoistureMassFractions,
    LiquidIcePotentialTemperatureState, temperature,
    mixture_gas_constant, mixture_heat_capacity

constants = ThermodynamicConstants(Float64); pˢᵗ = 1.0e5
θ, ρ, qᵛ, qˡ, qⁱ = 300.0, 1.0, 0.020, 0.005, 0.0
q  = MoistureMassFractions(qᵛ, qˡ, qⁱ)
Rᵐ, cᵖᵐ = mixture_gas_constant(q, constants), mixture_heat_capacity(q, constants)
γ  = cᵖᵐ / (cᵖᵐ - Rᵐ)
ΔL = (constants.liquid.reference_latent_heat*qˡ + constants.ice.reference_latent_heat*qⁱ)/cᵖᵐ

T_A = θ^γ * (ρ*Rᵐ/pˢᵗ)^(γ-1) + ΔL                                   # compressible dynamics
p_A = ρ * Rᵐ * T_A
T_B = temperature(LiquidIcePotentialTemperatureState(θ, q, pˢᵗ, p_A), constants)  # everywhere else
@show ΔL, T_A, T_B, T_B - T_A, (Rᵐ/cᵖᵐ)*ΔL
```

```
ΔL = 12.05 K ;  T_A = 295.99 K ;  T_B = 299.29 K
T_B − T_A = 3.31 K ≈ κ·ΔL = 3.36 K           # control (qˡ=qⁱ=0): 0.0 K
```

≈ **0.7 K per g/kg** of condensate. PR [#752](<https://github.com/NumericalEarth/Breeze.jl/issues/752>) fixes this by iterating the inversion to `T*` — necessary but not sufficient (symptom 2 is untouched).

---

## Symptom 2 — saturation adjustment is evaluated at the reference pressure, not the cell's density

`SaturationAdjustment` is Breeze-native (`src/Microphysics/saturation_adjustment.jl`; **not** CloudMicrophysics — CliMA's lives in `Thermodynamics.jl` and is internal-energy-based). Its inner step (`saturation_adjustment.jl:148-154`) holds the state's pressure **fixed**:

```julia
pᵣ  = 𝒰₀.reference_pressure
qᵛ⁺ = adjustment_saturation_specific_humidity(T, pᵣ, qᵗ, …)   # pressure-based qsat
# secant on  r(T) = T − temperature(𝒰₁)
```

So the equilibrium is found at the state's `reference_pressure`, held fixed through the secant. (θˡⁱ is held fixed too — `with_moisture` preserves `𝒰.potential_temperature` — so this conserves θˡⁱ regardless; it is *not* a cₚ-vs-cᵥ / enthalpy-vs-internal-energy distinction.) The only question is *which* pressure that is:

* **Anelastic:** `dynamics_pressure(::AnelasticDynamics) = reference_state.pressure = p̄(z)` (`src/AnelasticEquations/anelastic_dynamics.jl:108`), the hydrostatic reference, which *is* the thermodynamic pressure the EOS uses — the dynamic perturbation `p′` is excluded (`total_pressure = p̄ + p′` is kept separate). It is fixed in time, so evaluating `qsat` there is exact and self-consistent.
* **Compressible:** `dynamics_pressure` is the diagnosed `p = ρRᵐT` at the prognostic density `ρ`. As the secant varies `T`, the consistent pressure `ρRᵐT` moves with it, but `SaturationAdjustment` keeps the stale stored value — so `qsat` (and the Exner factor) are evaluated at a pressure that no longer matches the cell's density.

The consequence is directly observable: with `reference_pressure` pinned, the diagnosed density of the adjusted state *drifts* away from the cell's `ρ` — i.e. the equilibrium is found against the stale reference pressure, not the cell's density.

**Reproducer 2** (pure `Breeze`, committed code only) — saturate a supersaturated parcel and check whether the diagnosed `ρ` is preserved:

```julia
using Breeze.Thermodynamics: ThermodynamicConstants, MoistureMassFractions,
    LiquidIcePotentialTemperatureState, temperature, mixture_gas_constant
using Breeze.Microphysics: SaturationAdjustment, adjust_thermodynamic_state, WarmPhaseEquilibrium

constants = ThermodynamicConstants(Float64); pˢᵗ = 1.0e5
sa = SaturationAdjustment(Float64; equilibrium = WarmPhaseEquilibrium())

θ, p, qᵗ = 300.0, 9.0e4, 0.020                       # supersaturated parcel, all vapor
𝒰₀ = LiquidIcePotentialTemperatureState(θ, MoistureMassFractions(qᵗ), pˢᵗ, p)
ρ(𝒰) = p / (mixture_gas_constant(𝒰.moisture_mass_fractions, constants) * temperature(𝒰, constants))

𝒰₁ = adjust_thermodynamic_state(𝒰₀, sa, constants)  # equilibrium at the fixed reference pressure
T₀, T₁ = temperature(𝒰₀, constants), temperature(𝒰₁, constants)
@show T₀, T₁, 𝒰₁.moisture_mass_fractions.liquid     # latent warming + condensate
@show ρ(𝒰₀), ρ(𝒰₁), ρ(𝒰₁)/ρ(𝒰₀) - 1                # ≠ 0  ⇒  equilibrium at the stale p, not the cell's ρ
```

Output (run on the committed code, `--project=Breeze.jl/examples`):

```
T₀ = 291.148 K ;  T₁ = 295.118 K     # 1.61 g/kg condensed (qᵛ 0.0200 → 0.0184)
ρ(𝒰₀) = 1.0641 ;  ρ(𝒰₁) = 1.0525 ;  ρ₁/ρ₀ − 1 = −1.09 %
```

The diagnosed density drifts by `O(ΔL/T)` (here −1.1 %) because `T` rose while `p` was pinned. On a compressible cell `ρ` is the prognostic and fixed, so the converged vapor sits on `qsat(T, p_stale)` rather than `qsat(T, ρ)` — the equilibrium is not at the cell's own saturation point. (A downstream moist TC configuration measured a \~2–5 K saturated-state error from this; that figure is from work not in this repo and is not load-bearing — Reproducer 2 establishes the inconsistency from committed code alone.)

Because *every* moist scheme routes its partition through `SaturationAdjustment` (the 1M/2M `BulkMicrophysics`, and `ZeroMomentCloudMicrophysics` in `ext/BreezeCloudMicrophysicsExt/zero_moment_microphysics.jl`, all via `maybe_adjust_thermodynamic_state`, `update_atmosphere_model_state.jl:266`), **all of them inherit this on the compressible core.**

---

## Why anelastic is unaffected

For the anelastic core, `dynamics_pressure = reference_state.pressure = p̄(z)` — the fixed EOS pressure — so (1) the Exner inversion is exact and (2) `qsat` is already evaluated at the right pressure. The proposed fix is therefore strictly additive and compressible-θˡⁱ-only; the anelastic path, `StaticEnergyState`, and the existing anelastic examples are untouched.

---

## Proposed fix — give the compressible θˡⁱ state a density coordinate

One change closes both symptoms, mirroring the existing two-state pattern in `dynamic_states.jl`:

1. **New state** `CompressibleLiquidIceState{FT}(θ, q, pˢᵗ, ρ)` in `src/Thermodynamics/dynamic_states.jl` (density coordinate instead of `reference_pressure`):
   * `density(𝒰) = 𝒰.density` (trivial).
   * `temperature(𝒰)` = **iterated constant-density θˡⁱ→T inversion** to `T*` (closed-form `θ^γ(ρRᵐ/pˢᵗ)^(γ-1)` when no condensate, so no iteration in dry/vapor-only cells). **This is the** [#752](<https://github.com/NumericalEarth/Breeze.jl/issues/752>) **fix, located in the state.**
   * `saturation_specific_humidity(𝒰)` needs no override — the generic method (`dynamic_states.jl:12-16`) already calls `saturation_specific_humidity(T, density(𝒰), …)`, the density-based `qsat` (`vapor_saturation.jl:182`).
   * plus `with_moisture`, `with_temperature`, `total_specific_moisture`, `is_absolute_zero`.
2. **Density-consistent SA** — a specialized `adjust_thermodynamic_state(𝒰::CompressibleLiquidIceState, ::SaturationAdjustment, constants)` that solves a single secant on the constant-density θˡⁱ-conservation residual (holding θˡⁱ fixed, using density-based `qsat` and `p = ρRᵐT`). A single secant — not the generic one calling an *iterated* `temperature` — avoids nested iteration.
3. **Dispatch the constructor** — `diagnose_thermodynamic_state(…, ::LiquidIcePotentialTemperatureFormulation, dynamics::CompressibleDynamics, q)` returns `CompressibleLiquidIceState(θ, q, pˢᵗ, dynamics_density(dynamics)[i,j,k])`; the existing (anelastic) method stays.
4. **Retire formula A** — `compressible_time_stepping.jl` uses the state's `temperature` for the stored `T` and `p = ρRᵐT`, so dynamics and diagnostic share one inversion → `T_A ≡ T_B ≡ T*`.

### What it buys

* The `κ·ΔL` temperature inconsistency disappears (single iterated inversion) — **subsumes** [#752](<https://github.com/NumericalEarth/Breeze.jl/issues/752>) **/** NumericalEarth/Breeze.jl#625**-6**.
* `SaturationAdjustment` evaluates `qsat` at the cell's actual density for **every** scheme on the compressible core (SA, 1M, 2M, `ZeroMomentCloudMicrophysics`).
* A downstream constant-density rain-out scheme (not in this repo) can then drop its bespoke workaround and reduce to `ZeroMomentCloudMicrophysics(qc_0=0, τ_precip≈Δt)` on the corrected core.

## Relationship to in-flight work

* **PR** [#752](<https://github.com/NumericalEarth/Breeze.jl/issues/752>) (iterate `T = D·T^κ + L`): correct for symptom 1, but a point fix in `compressible_time_stepping.jl` that leaves symptom 2. Under this proposal [#752](<https://github.com/NumericalEarth/Breeze.jl/issues/752>)'s iteration moves into `CompressibleLiquidIceState.temperature` and serves every consumer.
* NumericalEarth/Breeze.jl#625 **item 6**: checkbox is premature — still buggy on `main` (formula A unchanged).

## Validation

* Promote Reproducer 1 to a round-trip regression: `with_temperature(𝒰, T)` → `temperature(𝒰)` returns `T`, and `T_A == T_B == T*` (fails today by `≈ κΔL`).
* Promote Reproducer 2 to assert `SaturationAdjustment` conserves density on the compressible state (`ρ(𝒰₁) == ρ(𝒰₀)`) and lands on the density-based saturation curve (`qᵛ == saturation_specific_humidity(T₁, ρ, …)`).
* A condensate-bearing θˡⁱ round-trip in both states.

---

**Question for review (@kaiyuan-cheng):** does a new `CompressibleLiquidIceState` (density coordinate) feel right as the home for both the iterated inversion and the density-consistent saturation, vs. keeping [#752](<https://github.com/NumericalEarth/Breeze.jl/issues/752>)'s point fix separate and handling the saturation elsewhere? The trade-off is a per-call fixed-point in `temperature` for condensate-bearing cells (short-circuited when dry).

ewquon:
**Closes #765; supersedes #752.**

On the compressible core, temperature is a state function of `(ρ, θˡⁱ, q)`, but it was diagnosed in
two inconsistent ways once condensate appeared:

1. **In the dynamics** (`temperature_and_pressure`, which sets `model.temperature` and `p = ρRᵐT`):
   a *non-iterated* inversion `T = θ^γ (ρRᵐ/pˢᵗ)^(γ-1) + (ℒˡqˡ + ℒⁱqⁱ)/cᵖᵐ` — the dry-adiabatic
   temperature plus a linear latent-heat shift.
2. **In `update_state!`** (the saturation-adjusted state's `temperature`, which sets the
   vapor/condensate partition): the Exner inversion `T = (p/pˢᵗ)^κ θ + (ℒˡqˡ + ℒⁱqⁱ)/cᵖᵐ`.

These are the first two Picard iterates of the true fixed point, so with condensate they disagree by
≈ (Rᵐ/cᵖᵐ)·ΔL (~0.7 K per g/kg of condensate) — the dynamics and the microphysics end up carrying
*different* temperatures. (This is the inversion symptom #752 fixed.) Compounding it,
`SaturationAdjustment` evaluated `qsat` against a *stale reference pressure* rather than the
prognostic density, so the equilibrium did not lie on the saturation curve at the cell's own ρ.

Both stem from one root: the θˡⁱ state was **pressure-based** — closed on a reference pressure
(`LiquidIcePotentialTemperatureState.reference_pressure`) — on a core whose prognostic is density.

## Change

- New **`LiquidIceDensityState`** — the *density-based* counterpart of
  `LiquidIcePotentialTemperatureState` (the distinction is the closing variable, density vs
  pressure — independent of the dynamical core; thanks @glwagner for the naming). Its `temperature`
  solves the *single* constant-density inversion `T = (ρRᵐT/pˢᵗ)^κ θ + ΔL/cᵖᵐ` (Newton, to
  convergence); `saturation_specific_humidity` and `exner_function` then evaluate at the actual ρ
  (with p = ρRᵐT).
- Density-consistent `adjust_thermodynamic_state(::LiquidIceDensityState, ::SaturationAdjustment)` —
  a secant on the constant-density θˡⁱ-conservation residual (`qsat` at ρ, p = ρRᵐT, θˡⁱ held fixed).
- `diagnose_thermodynamic_state(…, ::CompressibleDynamics, q)` builds it, and the compressible
  `temperature_and_pressure` uses the *same* inversion — so the dynamics temperature and the
  saturation-adjusted temperature are now the one converged value, and the inconsistency above is
  gone. **The anelastic (reference-pressure) path is untouched.**
- The inversion's convergence is configurable via `temperature_tolerance` / `temperature_maxiter` on
  `CompressibleDynamics` (defaults `1e-8` / `8` — **#752's knobs preserved**). They are carried onto
  the state so both the dynamics inversion and the saturation/diagnostic inversion use identical
  controls — a loose tolerance cannot make the two temperatures disagree again.

## Supersedes #752

#752's inversion fix moves into `LiquidIceDensityState.temperature` (its tolerance/maxiter knobs are
preserved here); this additionally fixes the saturation-against-a-stale-pressure error that #752 does
not touch.

## Tests

`test/compressible_saturation_adjustment.jl` (531 assertions): inversion self-consistency and
round-trip, honoring of the solver controls (`temperature_maxiter = 0` reproduces the old
non-iterated inversion), density-consistent saturation adjustment (vapor on `qsat(T, ρ)`, θˡⁱ
conserved), and a moist compressible `update_state!` regression (vapor sits on `qsat(T, ρ)` in
condensing cells, which also requires the dynamics and saturation temperatures to agree).

## Validation

SA-variant of the Klemp splitting supercell, `max|w|` at t = 15 min:

| config | `max|w|` |
|---|---|
| anelastic + SA (reference) | 19.8 m/s |
| compressible + SA, `temperature_maxiter = 0` (old non-iterated inversion) | **9.7 m/s** |
| compressible + SA, default (this fix) | **26.9 m/s** |

The old inversion under-heats moist parcels → roughly half the buoyancy → a too-weak updraft; the
fix recovers it by ×2.8 (9.7 → 26.9), reproducing #752's bug→fix recovery (11.4 → 27.7 vs anelastic
28.6 at full resolution). Reduced-resolution CPU here; a full-resolution GPU reproduction is
available on request.

cc @kaiyuan-cheng @glwagner

### DIFF
--- a/examples/acoustic_wave.jl
+++ b/examples/acoustic_wave.jl
@@ -50,7 +50,12 @@ Lx, Lz = 1000, 200  # (m)
 grid = RectilinearGrid(size = (Nx, Nz), x = (-Lx/2, Lx/2), z = (0, Lz),
                        topology = (Periodic, Flat, Bounded))
 
-model = AtmosphereModel(grid; dynamics = CompressibleDynamics(ExplicitTimeStepping()))
+# `temperature_tolerance = 0` selects the fixed-trip (unrolled) θˡⁱ→T inversion. The differentiable
+# pass at the end of this example takes a reverse-mode gradient through the compressible time step
+# with Reactant/Enzyme, and the default tolerance-based `while` inversion compiles to an XLA `while`
+# op that does not differentiate cheaply (NumericalEarth/Breeze.jl#767). The forward and adjoint
+# models use identical dynamics, so we make that choice here too.
+model = AtmosphereModel(grid; dynamics = CompressibleDynamics(ExplicitTimeStepping(); temperature_tolerance = 0))
 
 # ## Background state
 #
@@ -256,7 +261,7 @@ grid_ad = RectilinearGrid(ReactantState(); size = (Nx, Nz),
                           x = (-Lx/2, Lx/2), z = (0, Lz),
                           topology = (Periodic, Flat, Bounded))
 
-model_ad = AtmosphereModel(grid_ad; dynamics = CompressibleDynamics(ExplicitTimeStepping()))
+model_ad = AtmosphereModel(grid_ad; dynamics = CompressibleDynamics(ExplicitTimeStepping(); temperature_tolerance = 0)) # fixed-trip EOS inversion so Enzyme can differentiate it (see forward model above)
 
 # ### Fixed and varying fields
 #
--- a/examples/acoustic_wave.jl
+++ b/examples/acoustic_wave.jl
@@ -50,12 +50,14 @@ Lx, Lz = 1000, 200  # (m)
 grid = RectilinearGrid(size = (Nx, Nz), x = (-Lx/2, Lx/2), z = (0, Lz),
                        topology = (Periodic, Flat, Bounded))
 
-# `temperature_tolerance = 0` selects the fixed-trip (unrolled) θˡⁱ→T inversion. The differentiable
-# pass at the end of this example takes a reverse-mode gradient through the compressible time step
-# with Reactant/Enzyme, and the default tolerance-based `while` inversion compiles to an XLA `while`
-# op that does not differentiate cheaply (NumericalEarth/Breeze.jl#767). The forward and adjoint
-# models use identical dynamics, so we make that choice here too.
-model = AtmosphereModel(grid; dynamics = CompressibleDynamics(ExplicitTimeStepping(); temperature_tolerance = 0))
+# This example is dry, so the θˡⁱ→T inversion has an exact closed form and needs no Newton steps.
+# `temperature_tolerance = 0` selects the fixed-trip (unrolled) inversion and `temperature_maxiter = 2`
+# keeps the trip count tiny: the differentiable pass at the end of this example takes a reverse-mode
+# gradient through the compressible time step with Reactant/Enzyme, and the default tolerance-based
+# `while` inversion compiles to an XLA `while` op that does not differentiate cheaply, while a high
+# iteration count needlessly inflates the traced graph (NumericalEarth/Breeze.jl#767). The forward
+# and adjoint models use identical dynamics.
+model = AtmosphereModel(grid; dynamics = CompressibleDynamics(ExplicitTimeStepping(); temperature_tolerance = 0, temperature_maxiter = 2))
 
 # ## Background state
 #
@@ -261,7 +263,7 @@ grid_ad = RectilinearGrid(ReactantState(); size = (Nx, Nz),
                           x = (-Lx/2, Lx/2), z = (0, Lz),
                           topology = (Periodic, Flat, Bounded))
 
-model_ad = AtmosphereModel(grid_ad; dynamics = CompressibleDynamics(ExplicitTimeStepping(); temperature_tolerance = 0)) # fixed-trip EOS inversion so Enzyme can differentiate it (see forward model above)
+model_ad = AtmosphereModel(grid_ad; dynamics = CompressibleDynamics(ExplicitTimeStepping(); temperature_tolerance = 0, temperature_maxiter = 2)) # fixed-trip, low-iteration EOS inversion so Enzyme can differentiate it cheaply (see forward model)
 
 # ### Fixed and varying fields
 #
--- a/src/CompressibleEquations/CompressibleEquations.jl
+++ b/src/CompressibleEquations/CompressibleEquations.jl
@@ -61,7 +61,7 @@ using Oceananigans.BoundaryConditions: fill_halo_regions!
 using Oceananigans.Operators: divᶜᶜᶜ
 using Oceananigans.Utils: prettysummary, launch!
 
-using Breeze.Thermodynamics: mixture_gas_constant, mixture_heat_capacity, dry_air_gas_constant,
+using Breeze.Thermodynamics: mixture_gas_constant, dry_air_gas_constant,
                              vapor_gas_constant, ExnerReferenceState, temperature, LiquidIceDensityState
 
 using Breeze.AtmosphereModels: AtmosphereModels, AtmosphereModel, grid_moisture_fractions, dynamics_density, standard_pressure, thermodynamic_density, thermodynamic_density_name, specific_prognostic_moisture
--- a/src/CompressibleEquations/CompressibleEquations.jl
+++ b/src/CompressibleEquations/CompressibleEquations.jl
@@ -62,7 +62,7 @@ using Oceananigans.Operators: divᶜᶜᶜ
 using Oceananigans.Utils: prettysummary, launch!
 
 using Breeze.Thermodynamics: mixture_gas_constant, mixture_heat_capacity, dry_air_gas_constant,
-                             vapor_gas_constant, ExnerReferenceState
+                             vapor_gas_constant, ExnerReferenceState, temperature, LiquidIceDensityState
 
 using Breeze.AtmosphereModels: AtmosphereModels, AtmosphereModel, grid_moisture_fractions, dynamics_density, standard_pressure, thermodynamic_density, thermodynamic_density_name, specific_prognostic_moisture
 using Breeze.PotentialTemperatureFormulations: LiquidIcePotentialTemperatureFormulation
--- a/src/CompressibleEquations/compressible_dynamics.jl
+++ b/src/CompressibleEquations/compressible_dynamics.jl
@@ -19,6 +19,8 @@ Fields
 - `terrain_metrics`: [`TerrainMetrics`](@ref) for terrain-following coordinates (or `nothing`)
 - `Ω̃`, `ρΩ̃`: contravariant vertical velocity / momentum diagnostic fields (or `nothing` when no terrain metrics)
 - `terrain_reference_pressure`, `terrain_reference_density`: 3D reference pressure / density for the terrain pressure gradient force (or `nothing`)
+- `temperature_tolerance`, `temperature_maxiter`: relative convergence tolerance on the moist
+  equation-of-state temperature inversion step `|ΔT|/T`, and the iteration cap on that solve
 
 The `time_discretization` determines how tendencies are computed and which
 time-stepper is used:
@@ -37,6 +39,8 @@ struct CompressibleDynamics{TD, D, P, FT, RS, TM, CV, CM, TRP, TRD}
     contravariant_vertical_momentum :: CM      # ρΩ̃ diagnostic field (or Nothing)
     terrain_reference_pressure :: TRP          # 3D reference pressure for terrain PG (or Nothing)
     terrain_reference_density :: TRD           # 3D reference density for terrain buoyancy (or Nothing)
+    temperature_tolerance :: FT                # relative convergence tol |ΔT|/T for the moist EOS θˡⁱ→T inversion
+    temperature_maxiter :: Int                 # iteration cap for the moist EOS temperature inversion
 end
 
 """
@@ -60,17 +64,25 @@ Keyword Arguments
   hydrostatically-balanced reference state used in base-state subtraction. Can be a constant `θ₀`
   or a function `θ(z)`. Default: `nothing` (no base-state correction).
   When provided, an [`ExnerReferenceState`](@ref) is built during materialization.
+- `temperature_tolerance`: relative convergence tolerance on the moist EOS temperature
+  inversion step `|ΔT|/T` (default: `1e-8`)
+- `temperature_maxiter`: maximum number of moist EOS temperature inversion iterations
+  (default: `8`)
 """
 function CompressibleDynamics(time_discretization::TD = ExplicitTimeStepping();
                               standard_pressure = 1e5,
                               surface_pressure = 101325.0,
                               reference_potential_temperature = nothing,
                               reference_temperature = nothing,
-                              terrain_metrics = nothing) where TD
+                              terrain_metrics = nothing,
+                              temperature_tolerance = 1e-8,
+                              temperature_maxiter = 8) where TD
 
     FT = promote_type(typeof(standard_pressure), typeof(surface_pressure))
     pˢᵗ = convert(FT, standard_pressure)
     p₀ = convert(FT, surface_pressure)
+    temperature_tolerance = convert(FT, temperature_tolerance)
+    temperature_maxiter = Int(temperature_maxiter)
     # Store reference spec temporarily; ExnerReferenceState is built in materialize_dynamics.
     # If reference_temperature is given, store it as a NamedTuple to distinguish from θ₀.
     ref_spec = if reference_temperature !== nothing
@@ -82,7 +94,8 @@ function CompressibleDynamics(time_discretization::TD = ExplicitTimeStepping();
     # are built later in materialize_dynamics.
     return CompressibleDynamics(time_discretization, nothing, nothing, pˢᵗ, p₀, ref_spec,
                                 terrain_metrics,
-                                nothing, nothing, nothing, nothing)
+                                nothing, nothing, nothing, nothing,
+                                temperature_tolerance, temperature_maxiter)
 end
 
 Adapt.adapt_structure(to, dynamics::CompressibleDynamics) =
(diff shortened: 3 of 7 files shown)


kaiyuan-cheng (approved):

glwagner:
What is the performance hit from this PR?

For Reactant, we may need to implement a `FixedIterations` type solver in addition to the Newton solver.

The introduction of a solver into the algorithm has some significant downsides that must be approached with great care. I think an alternative is to use a different prognostic variable, correct?

ewquon:
`temperature_maxiter = 8` is the default but I dropped it to 2 to reduce the test cost for the acoustic wave test (because the `while` loop doesn't compile). We could just set it to 2 overall. 

https://github.com/NumericalEarth/Breeze.jl/pull/767/changes#diff-7e253f7f2f79cb8847e588cfb3f2e0228471fc0037edafbf74acc7b10d50be7eR54

I think regardless of the prognostic variable, the lesson learned was that we need to be careful and consistent in how we recover T.

glwagner:
The need for a solver is dictated by the intersection of the dynamics and the formulation right? Eg for a static energy formulation and compressible dynamics, we do not need a solver here?

glwagner:
In terms of fixed iterations, its not enough for reactant to just lower the max iterations. We have to completely remove the state-dependent nature of it. This is why it's useful to abstract the solver, so we can have something like `solver::FixedIterations(2)` that has no convergence criteria at all. We can put this in in a follow up PR.

glwagner:
The philosophical question is whether solver parameters belong in 1) the dynamics; 2) the formulation or 3) an independent property of `AtmosphereModel`. 

I think it could make sense to put the solver in the formulation. The idea being, we probably will only ever have two dynamics (anelastic and compressible), but we might have quite a few formulations (potential temperature, liquid ice potential temperature, static energy, total energy, entropy, temperature ?) Thus, it makes sense to associate the solver with the formulation which may or may not require it. (Thinking aloud)

ewquon:
I don't think an `AtmosphereModel` property makes sense. The interesting bit is that the solver depends on both the dynamics and the formulation that are chosen. I think associating solver parameters with the formulation is a good idea, analogous to the existing `SaturationAdjustment` struct.

glwagner:
> I don't think an `AtmosphereModel` property makes sense. The interesting bit is that the solver depends on both the dynamics and the formulation that are chosen. I think associating solver parameters with the formulation is a good idea, analogous to the existing `SaturationAdjustment` struct.

I think this could show that the formulation and dynamics are partly overlapping (they both refer to how we write the equations).

PR #770, opened 2026-06-05
Updates for Oceananigans v0.110

navidcy:
Uses `NormalFlowBoundaryCondition` instead of `OpenBoundaryCondition` after https://github.com/CliMA/Oceananigans.jl/pull/5648


giordano (approved):
I may have mentioned this before, but it's probably easier to start from dependabot PRs (like #768, which was opened before this PR). If a Dependabot PR isn't open yet, you can trigger it at https://github.com/NumericalEarth/Breeze.jl/network/updates/27106513/jobs

PR #771, opened 2026-06-05
Add `InstantaneousPrecipitation` microphysics, remove `ZeroMomentCloudMicrophysics`

ewquon:
## Summary

Adds `InstantaneousPrecipitation` — the Reed–Jablonowski large-scale condensation (LSC) scheme behind the DCMIP-2016 TC validation (coming in separate PR) — as a dedicated microphysics scheme built on the density-based saturation-adjustment machinery from #767, and **removes `ZeroMomentCloudMicrophysics`**, which it supersedes.

The once-per-step kernel condenses supersaturation at constant density (θˡⁱ held fixed), then rains out **all** condensate while **retaining** the released latent heat (θˡⁱ is bumped to its condensate-free value, so the diagnosed temperature stays warm). Both steps are now a 6-line composition of #767 primitives — `adjust_thermodynamic_state(::LiquidIceDensityState, ::SaturationAdjustment)` for the condensation and `with_temperature` for the rain-out — replacing LSC's previous bespoke ~60-line constant-density secant, which #767 made redundant.

Named for parity with the sibling `DCMIP2016KesslerMicrophysics` (file `dcmip2016_large_scale_condensation.jl`).

## Why a dedicated scheme (and why `ZeroMomentCloudMicrophysics` is removed)

`ZeroMomentCloudMicrophysics(qc_0 = 0, τ ≈ Δt)` looks equivalent on the surface, but it is **not** on the θˡⁱ core: ZMCM sinks condensate water from `ρqᵉ`, but **nothing sources `ρθˡⁱ`** — there is no `microphysical_tendency(_, Val{:ρθˡⁱ})` anywhere, and the θˡⁱ tendency has no microphysical term. So ZMCM would remove the rain's water but **lose its latent warming → no storm intensification**. LSC's defining behavior — instantaneous rain-out **with retained latent warming** — requires a fused kernel that writes `ρθˡⁱ` directly (the same pattern the Kessler scheme uses).

With `InstantaneousPrecipitation` covering the instant-rain-out use case, ZMCM no longer earns its keep: its only distinguishing physics was the rate-based removal (`τ_precip`/`qc_0`/`S_0`), and its θˡⁱ-warming gap made any precipitating use energetically wrong. Rather than fix the gap (a core tendency-interface change), this PR deletes the scheme:

- `ext/BreezeCloudMicrophysicsExt/zero_moment_microphysics.jl` and `test/cloud_microphysics_0M.jl` removed
- now-unused `Parameters0M` / `Microphysics0M.remove_precipitation` imports pruned from the extension
- stale zero-moment mentions removed from `microphysics_interface.jl` docstrings and the RICO example

`examples/boussinesq_bomex.jl` is untouched — it uses CloudMicrophysics' 0M functions directly through the legacy Boussinesq path, independent of the removed scheme.

## Validation

- **Unit parity 21/21** — the delegated path matches the pre-refactor inline formulas to `1e-9` (T, qᵛ⁺, qᶜ, θᶠ); latent warming retained after rain-out; subsaturated → no-op.
- **Integration 1031/1031** — drives the real `microphysics_model_update!` on an 8³ compressible model: condensation, rain-out, density-consistent saturation evaluated at the cell's own ρ (the #765 property), and the condensed-water budget.
- **ExplicitImports / Aqua / Core.Box 36/36.**
- **Post-removal check** — `BreezeCloudMicrophysicsExt` precompiles and loads with CloudMicrophysics; `cloud_microphysics_1M` suite passes 81/81.
- **End-to-end DCMIP-TC min-MSP** (0.5° ≈ 950.5 hPa, 0.25° ≈ 912.5 hPa) re-run on the cluster is **pending** — expected unchanged: parity is to `1e-9`, and the only behavioral delta is the SA secant converging to `tolerance = 1e-3` K (LSC default) instead of the old fixed-12-iteration machine precision, which is physically negligible for min MSP.


ewquon:
## End-to-end DCMIP-2016 TC validation (gpu-prod, H100)

Validated the refactored `DCMIP2016LargeScaleCondensation` end-to-end against the pre-refactor baselines, in an **isolated environment** (this branch + Oceananigans 0.109.2; the shared cluster env was left untouched). Both runs completed cleanly (10 days, no NaN).

| Resolution | Refactored (this PR) | Baseline (old LSC) | Δ |
|---|---|---|---|
| 0.5° | **951.4 hPa** | 950.5 hPa | +0.9 |
| 0.25° | **917.2 hPa** | 912.5 / 914.4 hPa | +3–5 |

- **0.5°** matches the baseline tightly (Δ0.9 hPa, well within the run-to-run variability — the min MSP oscillates ~955–961 hPa over the final day).
- **0.25°** is reproduced as a deep, coherent TC (max winds ~80 m/s, single wind max, same structure), ~3–5 hPa weaker than the pre-refactor baseline.

**On the 0.25° gap:** it is *not* attributable to the LSC refactor, which is bit-parity to `1e-9` against the pre-refactor formulas in the unit tests (T, qᵛ⁺, qᶜ, θᶠ). The 0.25° baseline (`run_1442/1443`) predates #767, so the new run additionally carries #767's density-state thermodynamics fix **and** Oceananigans 0.108 → 0.109.2 — real changes to the compressible core that shift an intense, chaotic high-res storm by a few hPa (the old LSC itself shows ~2 hPa run-to-run spread). The LSC default solver tolerance (1e-3 K) is hydrostatically negligible and cannot account for it.

**Conclusion:** the refactor preserves the physics — unit parity to 1e-9, full CI green, and both resolutions reproduce a deep coherent TC matching the published Reed–Jablonowski results.

ewquon:
## Follow-up: isolating the 0.25° difference (it's chaos, not the refactor)

To confirm the +3–5 hPa at 0.25° is unrelated to the LSC change, I ran the **old** bespoke-secant kernel and the **new** delegated kernel on the **identical #767 core** (same Oceananigans 0.109.2, same density-state thermodynamics) — an abbreviated 5-day 0.25° run each.

Matched-time min sea-level pressure (hPa):

| time | old kernel | new kernel | Δ |
|---|---|---|---|
| 12 h | 1005.1 | 1005.1 | **0.0** (bit-identical) |
| 1.0 d | 1002.4 | 1002.4 | **0.0** (bit-identical) |
| 2.0 d | 937.3 | 936.7 | 0.6 |
| 3.5 d | 919.7 | 919.0 | 0.7 |
| 4.0 d | 922.1 | 921.6 | 0.5 |
| 4.5 d | 931.8 | 931.5 | 0.3 |
| 4.7 d | 924.8 | 926.0 | 1.2 |

The two kernels are **bit-identical at the start** and **track within ~1 hPa** at calm sample points. The only large instantaneous gaps (≈5–7 hPa at 3.0 d / 4.3 d) occur at moments of rapid MSP fluctuation during eyewall deepening and **re-converge** within fractions of a day — the textbook signature of chaotic divergence of two near-identical trajectories, not a systematic offset.

The [0,5d] **instantaneous min** differs more (911.3 vs 918.5, Δ7.2) because the deepest transient is the most chaos-sensitive metric. That ~7 hPa spread between two *numerically-equivalent* runs is larger than the +3–5 hPa seen against the pre-#767 baseline — so the 0.25° difference is comfortably within the storm's chaotic envelope.

**Conclusion:** consistent with the 1e-9 unit parity, the refactor is numerically neutral; the 0.25° min-MSP scatter is the chaotic sensitivity of an intense, rapidly-deepening TC, not a behavioral change from this PR.

kaiyuan-cheng:
It looks that fixing the bug in `ZeroMomentCloudMicrophysics` will allow it to achieve the same result as `DCMIP2016LargeScaleCondensation`, am I correct?

glwagner:
Can we come up with a better name? I don't think we need to attached DCMIP2016, do we? Also, what does "large scale" refer to? We can use this scheme in a direct numerical simulation with a resolution of mm. Perhaps something more explicit would be useful?

ewquon:
Well first decision as noted in #772: do we build instantaneous condensation (as done here) into `ZeroMomentCloudMicrophysics`? If so, then this gets thrown out.

If we keep this, I'm not tied to the "DCMIP2016" prefix, which was intended to follow the existing "DCMIP2016KesslerMicrophysics", or the "LargeScaleCondensation" name. I agree that in general "large" as a qualifier for scale is vague. We could go with something like `PrecipitatingSaturationAdjustment` or `InstantaneousPrecipitation`

glwagner:
I like `InstantaneousPrecipitation`! Fair point about Kessler, but I feel that scheme is truly specific to DCMIP whereas this one is fairly generic.

kaiyuan-cheng:
> It looks that fixing the bug in `ZeroMomentCloudMicrophysics` will allow it to achieve the same result as `DCMIP2016LargeScaleCondensation`, am I correct?

It appears any fix to `ZeroMomentCloudMicrophysics` would be intrusive (updating θˡⁱ in-place inside the compressible dynamics) because θˡⁱ cannot distinguish the removal of rain from that of water vapor. And I don't think it is worth the effort as it is not really being used. I recommend implementing this `InstantaneousPrecipitation` and dropping `ZeroMomentCloudMicrophysics`

kaiyuan-cheng (approved):

kaiyuan-cheng:
close #772

PR #773, opened 2026-06-09
Cap compatibility with Reactant to v0.2.262

giordano:
We're seeing XLA compile time skyrocketing since Reactant [v0.2.263](https://github.com/EnzymeAD/Reactant.jl/releases/tag/v0.2.263), let's cap compat to avoid pulling newer versions until that's resolved (I believe @Pangoraw may be looking into it at some point)

PR #775, opened 2026-06-09
Revert "Updates for Oceananigans v0.110"

giordano:
Reverts NumericalEarth/Breeze.jl#770.  I want to see if this is what caused the slowdown, rather than Reactant v0.2.263 (#773).  CC @dkytezab.


giordano:
This also hangs.  I'm so confused 😂

giordano:
Looking at [last CI runs on main](https://github.com/NumericalEarth/Breeze.jl/actions/workflows/CI.yml?query=branch%3Amain), Reactant tests got slower already before #770, so this is definitely not the culprit.

PR #776, opened 2026-06-10
Terrain-following: sample topography at ξnode/ηnode for lat-lon support

ewquon:
## Summary

`set_topography!` sampled the user terrain function at `xnode(i, grid, Center())` / `ynode(j, grid, Center())`, which are `RectilinearGrid`-oriented and don't generalize: on a `LatitudeLongitudeGrid` those accessors don't hand the topography function the horizontal `(λ, φ)` it expects.

This switches to **`ξnode`/`ηnode`** — the grid-agnostic horizontal node accessors:

- `(x, y)` on a `RectilinearGrid`
- `(λ, φ)` on a `LatitudeLongitudeGrid` → terrain-following now works on lat-lon LAMs
- `nothing` for the degenerate meridional node on a 2D x–z `(Periodic, Flat, Bounded)` grid — exactly matching prior behavior

### Why not `node(...)[1:2]`?

The obvious grid-agnostic substitute is wrong on Flat grids: `node` *drops* Flat directions from its returned tuple, so on a 2D x–z grid it yields `(x, z)` and leaks the **vertical** coordinate into the meridional slot. `ξnode`/`ηnode` are the primitives `node` is built from but never drop Flat directions. A regression test pins this down.

## Tests

- **Horizontal-coordinate handling** across Flat / 3D / lat-lon grids. The Flat-grid case uses a probe `h(x, y) = y === nothing ? 1.0 : 2.0` asserting the second argument is `nothing` (not `z`) — this assertion fails against the `node[1:2]` approach.
- **`LatitudeLongitudeGrid` terrain-physics testset** mirroring the existing rectilinear one: `follow_terrain!` + `CompressibleDynamics(...; terrain_metrics)` + one `time_step!` → finite `w` and contravariant vertical velocity. Cheap insurance that the metric terms behave under spherical geometry.

Full `test/terrain_following.jl` passes 664/664; `quality_assurance` (ExplicitImports/Aqua) passes 36/36.

### DIFF
--- a/src/TerrainFollowingDiscretization/follow_terrain.jl
+++ b/src/TerrainFollowingDiscretization/follow_terrain.jl
@@ -70,7 +70,7 @@ function follow_terrain!(grid, topography, ::BasicTerrainFollowing, pressure_gra
     h_field = CenterField(grid, indices=(:, :, 1))
 
     # Set topography values and fill halos
-    set_topography!(h_field, grid, topography)
+    set_topography!(h_field, topography)
     fill_halo_regions!(h_field)
 
     # Compute sigma and eta on the grid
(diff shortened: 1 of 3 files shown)


src/TerrainFollowingDiscretization/follow_terrain.jl
```diff
@@ -98,7 +98,12 @@ end
 # is intentionally more general.
 function set_topography!(h_field, grid, topography::Function)
```
glwagner: what is wrong with using `set!(h_field, topography)` here? Is there a difference between `h_field.grid` and the argument `grid`?

ewquon: This is more robust, per the comment below on Flat dims. Good call on the redundant `grid` arg

ewquon: Ah, I missed the `::Function` qualifier. I've added a generic version, which is actually what would be called when using real terrain data:

`regrid_topography(grid; ETOPO2022()) → field → follow_terrain!(grid, field) → set_topography! → set!`

glwagner:
This was superceded by #712 (the work here ended up being incorporated into #712, which was easier than merging this PR separately). Thank you @ewquon and @kaiyuan-cheng for your work on this!

PR #780, opened 2026-06-11
Unify iterative solvers; move the temperature solver from dynamics to the formulation

glwagner:
Follow-up to #767, addressing the two design concerns raised in its discussion: the θˡⁱ→T inversion's solver controls lived on the *dynamics* rather than the formulation, and they were raw `tolerance`/`maxiter` scalars (with a sign-of-tolerance hack selecting the Reactant-safe unrolled loop) rather than a proper abstraction.

## `Breeze.Solvers`

A new module defining a vocabulary of isbits solver types that iterative algorithms dispatch on:

* `NewtonSolver(reltol, abstol, maxiter)` — Newton iteration, stops when `|Δx| ≤ max(abstol, reltol|x|)`
* `SecantSolver(reltol, abstol, maxiter)` — secant iteration, stops when `|r| ≤ max(abstol, reltol|scale|)`
* `FixedIterations(n)` — exactly `n` steps, **no convergence test at all**: the fixed trip count unrolls to straight-line code, the form Reactant/Enzyme need (the tolerance-based `while` loop traces to an XLA `while` op with a pathological adjoint — #767)
* `nothing` — no iteration; the algorithm returns its closed-form initial guess

plus shared `newton_solve`/`secant_solve` drivers so the loop logic (including the degenerate-slope guards) exists once. Solvers take the place of every ad hoc `tolerance`/`maxiter` pair in the codebase:

| site | before | after |
|---|---|---|
| compressible θˡⁱ→T inversion | `temperature_tolerance`/`temperature_maxiter` on `CompressibleDynamics`, sign hack for the unrolled form | `temperature_solver` on the formulation (see below) |
| `SaturationAdjustment` | `tolerance`, `maxiter::FT` (default `Inf`!) | `solver = SecantSolver(abstol=1e-3, maxiter=100)` |
| `dewpoint_temperature` / `DewpointTemperature` | `tolerance`/`maxiter` kwargs | `solver = SecantSolver(reltol=1e-4, maxiter=10)` |
| legacy `MoistAirBuoyancies` secant | hardcoded `δ=1e-3`, **no iteration cap** | shared driver, `SecantSolver(abstol=1e-3, maxiter=100)` |

Scalar (kernel-level) functions take the solver positionally; only host-level constructors use keyword arguments.

## Solver moves to the formulation

The need for an iterative inversion is dictated by the intersection of the formulation and the dynamics (anelastic + θˡⁱ and static energy are closed-form; only compressible + θˡⁱ is implicit). The solver therefore lives on `LiquidIcePotentialTemperatureFormulation`:

```julia
formulation = LiquidIcePotentialTemperatureFormulation(temperature_solver = FixedIterations(2))
model = AtmosphereModel(grid; formulation, dynamics = CompressibleDynamics(...))
```

The default is a `DefaultTemperatureSolver()` sentinel resolved at materialization through a new interface hook `default_temperature_solver(dynamics)`: `nothing` generically (anelastic — no solver, closed form), `NewtonSolver(reltol=1e-8, abstol=0, maxiter=8)` for `CompressibleDynamics` (#767's defaults preserved). Tolerances are converted to `eltype(grid)` at materialization so `Float32` kernels don't promote. `LiquidIceDensityState` carries the solver as a type parameter, so `FixedIterations` compiles a branch-free kernel — not a branch that happens to be false.

## Breaking changes

Pre-1.0 clean breaks, each with an informative `ArgumentError` pointing at the replacement:

* `CompressibleDynamics(temperature_tolerance=..., temperature_maxiter=...)` → `LiquidIcePotentialTemperatureFormulation(temperature_solver=...)`
* `SaturationAdjustment(tolerance=..., maxiter=...)` → `SaturationAdjustment(solver=...)`
* `dewpoint_temperature(...; tolerance, maxiter)` → positional `solver`; `DewpointTemperature(model; solver=...)`

Behavior is otherwise preserved: defaults reproduce #767's converged Newton inversion and the existing saturation-adjustment tolerances (the secant default `maxiter` changes from `Inf` to 100, and the legacy `MoistAirBuoyancies` loop's stopping rule changes from step-size to residual at the same 1e-3 K scale — both inconsequential for converging iterations, and the latter can no longer hang).

## Tests

* New `test/solvers.jl`: driver unit tests (tolerance/fixed/`nothing` forms, degenerate-slope guards, `materialize_solver` float conversion).
* `test/compressible_saturation_adjustment.jl`: the knob-honoring tests become solver-type tests (`nothing` reproduces the non-iterated inversion; `FixedIterations(8)` matches converged Newton; an abstol-only `NewtonSolver` converges).
* `examples/acoustic_wave.jl` uses `FixedIterations(2)` via the formulation for the differentiable Reactant pass.

Locally green: `solvers`, `saturation_adjustment`, `compressible_saturation_adjustment`, `moist_air_buoyancy`, `diagnostics`, `dynamics`, `atmosphere_model_construction`, `quality_assurance`, `doctests` (1,249 assertions).

cc @kaiyuan-cheng @ewquon


ewquon:
Do we have justification for `reltol` vs `abstol` and the different `maxiter` settings for the different call sites?

ewquon (approved):

glwagner:
> Do we have justification for `reltol` vs `abstol` and the different `maxiter` settings for the different call sites?

This is a good question. I will look into this.

Also @ewquon @kaiyuan-cheng there is a lesson in the need for this PR -- we should be cautious about adding iterative solvers to the algorithm in general. When we do add them, we should take care to implement the solvers in a way that allows a fixed approximation to be implemented, using dispatch. This is needed for friendliness with Reactant/autodiff.

As an example, this PR replaces float type parameters such as `temperature_tolerance` with a more general `solver` type. Using a more general abstraction allows us to write `solver = nothing` (which is also the correct choice for completely dry simulations), thereby avoiding the solve completely. On the other hand, while one could set `temperature_tolerance=0`, this numeric-based approach (vs a type/dispatched-based approach) is not inferrable by the compiler. In other words it is almost always preferable to use type-based design rather than numeric-based design.

examples/acoustic_wave.jl
```diff
@@ -51,13 +51,14 @@ grid = RectilinearGrid(size = (Nx, Nz), x = (-Lx/2, Lx/2), z = (0, Lz),
...
-# `while` inversion compiles to an XLA `while` op that does not differentiate cheaply, while a high
+# `FixedIterations(2)` selects the fixed-trip (unrolled) inversion with a tiny trip count: the
+# differentiable pass at the end of this example takes a reverse-mode gradient through the
+# compressible time step with Reactant/Enzyme, and the default tolerance-based `NewtonSolver`
+# (a `while` loop) compiles to an XLA `while` op that does not differentiate cheaply, while a high
 # iteration count needlessly inflates the traced graph (NumericalEarth/Breeze.jl#767). The forward
-# and adjoint models use identical dynamics.
-model = AtmosphereModel(grid; dynamics = CompressibleDynamics(ExplicitTimeStepping(); temperature_tolerance = 0, temperature_maxiter = 2))
+# and adjoint models use identical dynamics and formulation.
+formulation = LiquidIcePotentialTemperatureFormulation(temperature_solver = FixedIterations(2))
```
glwagner, line 60: i think for this dry case we can use `temperature_solver=nothing`. But maybe its good to stress test the fixed iterations anyways.

glwagner:
Good catch — there *was* a coherent rule behind these choices, but it wasn't written down or applied consistently. I've now documented it and made the call sites consistent.

**The convention (now in the `Solvers` module docstring):**

Tolerance *type* follows the residual's natural scale:

| | Residual | Why | Setting |
|---|---|---|---|
| **Absolute** | Temperature (K) — `θˡⁱ→T` inversion, saturation adjustment, Boussinesq adjustment | Bounded away from zero with a fixed precision target | `abstol = 1e-4` K |
| **Relative** | Saturation vapor pressure (Pa) — dewpoint | Spans ~2 decades over the atmospheric range; an absolute Pa tolerance is meaningless across it | `reltol = 1e-4` vs `pᵛ` |

`1e-4` K is far below any physically or numerically meaningful temperature increment, and both Newton (quadratic) and secant (superlinear) reach it in a handful of iterations.

**`maxiter` follows convergence order + role:**
- Newton inversion (quadratic): `8`
- Secant temperature solves (superlinear): `20` — was `100`, which was over-provisioned
- Dewpoint (diagnostic, not in the hot loop): `10`

**`FixedIterations` (hydrostatic pressure):** the two solves used `7` and `5` for the *identical* monotone residual — unified to `5`, since Newton reaches machine precision well within that from the continuous-Π guess. There's now a test in `test/solvers.jl` confirming `FixedIterations(5)` matches a 30-iteration reference to machine precision across the atmospheric range, plus a test pinning the default tolerances.

PR #785, opened 2026-06-12
Resolve possible issues detected by JET

giordano:
This resolves
```julia-repl
julia> using Breeze, JET, LinearAlgebra, Oceananigans, MPI, StaticArraysCore, JLD2

julia> JET.test_package(Breeze; ignored_modules=(Base, LinearAlgebra, Oceananigans, MPI, StaticArraysCore, JLD2))
JET-test failed at /home/mose/.julia/packages/JET/Nf2tu/src/JETBase.jl:1400
  Expression: (JET.report_package)(Breeze; toplevel_logger = nothing, ignored_modules = (Base, LinearAlgebra, Oceananigans, MPI, StaticArraysCore, JLD2))
  ═════ 2 possible errors found ═════
  ┌ set_to_mean!(ref::ReferenceState, model::Any; rescale_densities::Any) @ Breeze.AtmosphereModels /home/mose/.julia/dev/Breeze/src/AtmosphereModels/set_to_mean.jl:96
  │ local variable `ρᵣ_old` may be undefined: ρᵣ_old::Any
  └────────────────────
  ┌ acoustic_rk3_substep_loop!(model::Any, substepper::Any, Δt::Any, β_stage::Any, Uᴸ::Any) @ Breeze.CompressibleEquations /home/mose/.julia/dev/Breeze/src/CompressibleEquations/acoustic_substepping.jl:1640
  │ no matching method found `compute_velocities!(::HydrostaticFreeSurfaceModel)`, `compute_velocities!(::NonhydrostaticModel)` (2/3 union split): (Breeze.CompressibleEquations.AtmosphereModels).compute_velocities!::typeof(Breeze.AtmosphereModels.compute_velocities!)(model::Any)
  └────────────────────
  
ERROR: There was an error during testing
```
These are minor issues, it's mainly about the ability of the compiler to understand the code more easily.

Together with #784, JET would be clean after these PRs.

### DIFF
--- a/src/AtmosphereModels/set_to_mean.jl
+++ b/src/AtmosphereModels/set_to_mean.jl
@@ -76,6 +76,11 @@ fields are left as-is and only diagnostics are recomputed.
 function set_to_mean!(ref::ReferenceState, model; rescale_densities=false)
     constants = model.thermodynamic_constants
 
+    if rescale_densities
+        ρᵣ_old = similar(dynamics_density(model.dynamics))
+        parent(ρᵣ_old) .= parent(dynamics_density(model.dynamics))
+    end
+
     # Update reference temperature and moisture from horizontal means
     mean!(ref.temperature, model.temperature)
     fill_halo_regions!(ref.temperature)
@@ -88,8 +93,6 @@ function set_to_mean!(ref::ReferenceState, model; rescale_densities=false)
     compute_hydrostatic_reference!(ref, constants)
 
     if rescale_densities
-        ρᵣ_old = similar(dynamics_density(model.dynamics))
-        parent(ρᵣ_old) .= parent(dynamics_density(model.dynamics))
         rescale_density_weighted_fields!(model, ρᵣ_old)
     end
 
--- a/src/AtmosphereModels/set_to_mean.jl
+++ b/src/AtmosphereModels/set_to_mean.jl
@@ -76,11 +76,6 @@ fields are left as-is and only diagnostics are recomputed.
 function set_to_mean!(ref::ReferenceState, model; rescale_densities=false)
     constants = model.thermodynamic_constants
 
-    if rescale_densities
-        ρᵣ_old = similar(dynamics_density(model.dynamics))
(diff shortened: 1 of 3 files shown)


src/AtmosphereModels/set_to_mean.jl
```diff
@@ -93,6 +88,8 @@ function set_to_mean!(ref::ReferenceState, model; rescale_densities=false)
     compute_hydrostatic_reference!(ref, constants)
 
     if rescale_densities
+        ρᵣ_old = similar(dynamics_density(model.dynamics))
+        parent(ρᵣ_old) .= parent(dynamics_density(model.dynamics))
         rescale_density_weighted_fields!(model, ρᵣ_old)
     end
```
giordano, line 97: This one I'm slightly unsure about: I don't understand why the `if rescale_densities` blocks were split to start with.  The first one is reading from `model`, which as far as I can tell isn't modified before the second block (I only see functions modifying `ref` in between).  Tests pass locally for me with this change, but I don't know if I'm missing something.

giordano: [Code coverage data](https://app.codecov.io/gh/NumericalEarth/Breeze.jl/pull/785) shows that the `rescale_densities = true` case isn't tested at all, so tests passing doesn't say much.

giordano: Uhm, in principle `ref` may be an inner field of `model`, so that modifying in-place the former also modifies the latter?  That'd complicate things.  I'm looking at https://github.com/NumericalEarth/Breeze.jl/blob/c0fd82eddb86b7d6f53e12a16801c2d57fd9042b/test/reference_states.jl#L459-L479 where `model` contains `reference_state`, and then we call `set_to_mean!(reference_state, model)`.

giordano: Ok, I reverted this change for the time being, the other one should be uncontroversial.

navidcy:
Sorry I missed this!

PR #787, opened 2026-06-15
Refresh model state after the InstantaneousPrecipitation operator-split update

kaiyuan-cheng:
## Summary

`InstantaneousPrecipitation` (IP) microphysics did not refresh the model state after its operator-split update (`microphysics_model_update!`). This PR fixes that.

## Root cause

The time-steppers run the operator-split microphysics update (`microphysics_model_update!`) as the **last** substantive step of each RK step, *after* `update_state!`:

```julia
update_state!(model, callbacks; compute_tendencies = true)   # ssp_runge_kutta_3.jl:241
microphysics_model_update!(model.microphysics, model)         # :245
```

For IP, `microphysics_model_update!` launches a `:xyz` kernel that mutates the prognostics `ρqᵛ` and `ρθˡⁱ` **in the interior only**, then returned without refreshing the model state. As a result, after every IP step:

- halo regions of `ρqᵛ`/`ρθˡⁱ` were stale,
- diagnostics (`temperature`, `qᵛ`, …) were inconsistent with the mutated prognostics,
- tendencies were stale, so the next step's first RK substep advanced the microphysics-updated prognostics using tendencies computed from the **pre-microphysics** state.

## Fix

Call `update_state!(model)` at the end of `microphysics_model_update!(::IP, model)` (and import it from `Oceananigans.TimeSteppers`), so the operator-split update finishes by refilling halos and recomputing diagnostics/tendencies.

This is safe because `maybe_adjust_thermodynamic_state(::IP)` is a no-op (so the trailing `update_state!` does not re-condense / double-count latent heat) and `update_microphysical_fields!(::IP)` only writes `qᵛ`, so the kernel-set `precipitation_rate` diagnostic is preserved.

## Testing

`test/instantaneous_precipitation.jl` passes locally (CPU): parity 21/21, integration 1034/1034.


glwagner (approved):
why doesn't InstantaneousPrecipitation get called every stage? would seem appropriate (not a slow process)

glwagner:
Idle question unrelated to this PR, isn't it possible to convert q^v to q^{cl}, and then lose q^{cl} at some rate via tendency without actually directly modeling rain + sedimentation? This seems like it could also be called "instantaneous precipitation". And it would be more natural to implement (does not require the "model update" hack).

kaiyuan-cheng:
> why doesn't InstantaneousPrecipitation get called every stage? would seem appropriate (not a slow process)

"Instantaneous" may be a bit misleading. Whether or not it is a fast process is a open question. (Though I think every dt is fast enough for this kind of parameterization.) One difficulty calling it every RK stage is that I have trouble passing fractional dt into schemes that couple dynamics through `microphysics_model_update!`. `model.clock.last_Δt` gives a full dt. As a result, calling `microphysics_model_update!` at every RK stage ends up tripling the microphysics tendencies, which is what #762 was trying to fix.

ewquon (approved):
Looks good to me. I'm re-running my TC case to verify that we don't get a dramatic change in solution, I'll let you know if anything is amiss.

kaiyuan-cheng:
> Idle question unrelated to this PR, isn't it possible to convert q^v to q^{cl}, and then lose q^{cl} at some rate via tendency without actually directly modeling rain + sedimentation? This seems like it could also be called "instantaneous precipitation". And it would be more natural to implement (does not require the "model update" hack).

I tried to fix `ZeroMomentCloudMicrophysics` that uses a mechanism similar to what you described. Yes, it is fixable but it would have to touch thermodynamics inside the compressible dynamics as θˡⁱ cannot distinguish the removal of rain from that of water vapor, which I don't think it worth the effort. At the end, I decided to go with `InstantaneousPrecipitation` and clean up `ZeroMomentCloudMicrophysics`

glwagner:
A problem we had previously was that `update_state!` couldn't accept the `dt`. However now that `microphysics_model_update!` is separate, we can add more arguments like `stage=true` or whatever `dt` is needed.

Just a thought, but out of scope I think here.

PR #791, opened 2026-06-18
BreezeRRTMGPExt: support compressible dynamics on terrain grids in radiation

kaiyuan-cheng:
## Summary

Radiation now reads the model's *thermodynamic* state through the dynamics interface —
`dynamics_pressure(model.dynamics)` and `total_density(model.dynamics)` — instead of reaching into
`model.dynamics.reference_state`. This makes RRTMGP work with `CompressibleDynamics`, including on
terrain-following grids and with `reference_state = nothing`, and makes compressible radiation see
the actual pressure and density instead of the base state.

## Motivation

All three RRTMGP update paths read `model.dynamics.reference_state.pressure` / `.density` directly
as the layer pressure and density. That is correct for anelastic dynamics, where the hydrostatic
reference *is* the thermodynamic pressure. For compressible dynamics it is wrong twice over:

- **It is the wrong field.** A reference state (flat 1-D or 3-D terrain) is a pressure-gradient-force
  device for base-state subtraction. The compressible core already carries the true diagnosed
  equation-of-state pressure `dynamics.pressure` and total air density `dynamics.total_density`.
- **It may not exist.** `reference_state === nothing` when the user disables it, so radiation hit a
  `nothing.pressure` access.

## Changes

`ext/BreezeRRTMGPExt`:

- Gas state (clear-sky and all-sky), `rrtmgp_shared_utilities.jl`: layer and level pressures from
  `dynamics_pressure`.
- Cloud state, `all_sky_radiative_transfer_model.jl`: cloud water paths weight total air mass, so
  from `total_density`. Kernel field argument renamed `ρᵣ` → `total_density`, matching
  `_compute_temperature_and_pressure!`; the scalar stays `ρ` (avoids `ρᶜ`, which reads as "cloud
  density" given the `ᶜ` species superscript in `notation.md`).
- Gray, `gray_radiative_transfer_model.jl`: `p_lay` / `p_lev` from `dynamics_pressure`. Also corrects
  a stale docstring, which claimed the surface level pressure came from the reference state's
  `surface_pressure` — the kernel has always interpolated it from the layer field.
- Explicit imports for the two accessors.

`src/`:

- Export `total_density` from `AtmosphereModels` and re-export from `Breeze`. It was already the
  internal accessor used by scalar advection, the thermodynamics, the EOS, and buoyancy; the
  radiation extension needs it as public API.
- Extend the `dynamics_pressure` docstring with the contract: this is the pressure every physics
  parameterization reads, radiation included, and a (flat or terrain) reference state must never
  substitute for the thermodynamic state. `total_density` is the density counterpart.

`test/unit_tests.jl`: `total_density(dynamics)` assertions alongside the existing `dynamics_pressure`
ones — `=== reference_state.density` for anelastic, `=== dynamics.total_density` for compressible.

## Behavior

| Dynamics | Before | After |
| --- | --- | --- |
| Anelastic | `reference_state.pressure` / `.density` | **Unchanged** — the accessors resolve to exactly those fields |
| Compressible, flat | Hydrostatic reference base state | Diagnosed `dynamics.pressure` and `dynamics.total_density` |
| Compressible, terrain-following or `reference_state = nothing` | Base state, or an error when absent | Diagnosed total state |

RRTMGP assumes level pressures are positive and monotonically decreasing with height. That holds
exactly for the anelastic hydrostatic reference, and in practice for the compressible diagnosed
pressure, whose hydrostatic part dominates dynamic and acoustic perturbations by orders of magnitude.
This is recorded as a comment at the gas-state read.

## Testing

- `test/unit_tests.jl` extended as above; the CPU CI jobs (Linux / macOS / Windows, Julia min and
  1.12.6) are green, as are docs and the whitespace check.
- The underlying fix was exercised by a compressible-dynamics run on a terrain-following grid with
  RRTMGP radiation on a single H100 during development.
- No automated RRTMGP regression test is added (that path needs artifacts and a GPU). Happy to add a
  small terrain-compressible radiation smoke test if reviewers want one.
- The remaining GPU CI failure is in `test/reactant/lat_lon_compilation.jl` — Reactant tracing a
  `LatitudeLongitudeGrid` constructor (`non-boolean (TracedRNumber{Bool}) used in boolean context`)
  — unrelated to this diff.


ext/BreezeRRTMGPExt/all_sky_radiative_transfer_model.jl
```diff
@@ -310,7 +310,7 @@ function update_rrtmgp_cloud_state!(cloud_state, model, liquid_effective_radius,
     grid = model.grid
     arch = architecture(grid)
 
-    ρᵣ = model.dynamics.reference_state.density
+    ρᵣ = radiation_reference_density(model.dynamics)
```
glwagner: why do we need multiple definitions of the reference density?

kaiyuan-cheng: This is because the direct `reference_state.density` accesses break on terrain-following grids where `reference_state === nothing` and the reference lives in `terrain_reference_density`. We could unify the interface but it would be a breaking change.

glwagner:
Merged `main` into this branch (no conflicts; picks up 0.7.3 and the 8 commits it was behind, including the FlatauPolynomial addition and the Reactant unpin). NumericalEarth.jl's docs environment now pins Breeze to this branch — the land-coupled MC3E downscaling example runs RRTMGP with CompressibleDynamics on a terrain-following grid, which needs exactly this fix.

glwagner:
Pushed a one-line fix: the prognostic-density fallback in `radiation_reference_density` read `dynamics.density`, which no dynamics type carries — `CompressibleDynamics` holds `total_density` (the pressure twin correctly uses `dynamics.pressure`). Hit in NumericalEarth's land-coupled MC3E example, where the nested child builds `CompressibleDynamics` without a reference profile spec on a terrain grid, so both `reference_state` and `terrain_reference_density` are `nothing` and radiation takes exactly this fallback.

glwagner:
Refactored the new accessors per review discussion: `radiation_reference_pressure/density` were radiation-private cascades, but the concept — best available static reference, else terrain reference, else prognostic state — belongs to the dynamics interface. They are now `reference_pressure`/`reference_density` beside `mean_pressure`, with per-dynamics methods, and the RRTMGP paths call those. (Neither existing getter fit: `mean_pressure(::CompressibleDynamics)` is the prognostic pressure and pinned by `pressure_anomaly = 0`, and `dynamics_density` returns the *dry* prognostic density — the wrong layer mass for radiation.)

glwagner:
## Correctness review: reference vs. actual state into RRTMGP

Reviewing the `reference_pressure`/`reference_density` choice, I surveyed what three compressible reference models feed their RRTMG(P) — ClimaAtmos (shares RRTMGP.jl), and WRF-ARW + MPAS-A (the terrain-following/split-state compressible cores closest to Breeze). They're unanimous, and Breeze is the outlier:

| Model | Layer pressure | Air **mass** (gas absorbers / cloud water paths) | Temperature | Static reference? |
|---|---|---|---|---|
| ClimaAtmos | actual diagnosed `p = ρRₘT` | actual `ρ` | actual `T` | No |
| MPAS-A | hydrostatic, integrated from **actual moist ρ** | actual moist `ρ` (hydrostatic `Δp`) | actual `T` | No |
| WRF-ARW | hydrostatic `p_hyd`, from **actual total mass** `MUT` | actual mass (hydrostatic `Δp/g`) | actual `T` | No |
| **Breeze (this PR)** | **reference** pressure | **reference** density | actual `T` | **Yes** |

Evidence:
- **ClimaAtmos** — `layer_pressure .= ᶜp` with `ᶜp = TD.air_pressure(…, ᶜT, Y.c.ρ, …)` (`radiation/update_inputs.jl:52`, `precomputed_quantities.jl:668`); cloud water path `u.c.ρ * qₗ * Δz` (`update_inputs.jl:287`). No reference-state field anywhere in the radiation path.
- **WRF** — radiation is called with `P=grid%p_hyd, P8W=grid%p_hyd_w` (`dyn_em/module_first_rk_step_part1.F:289`), the hydrostatic pressure integrated from total column mass `MUT=mub+mu'` (`module_big_step_utilities_em.F:4957,4968`); absorber amounts `coldry` and `gliqwp/gicewp` from `pdel/g` (`module_ra_rrtmg_lw.F:11404, 12489-12490`). The full pressure `p_phy=p+pb` is built but **not** passed to radiation; base state `pb` only reconstructs it for the Exner/temperature.
- **MPAS** — feeds `pres_hyd_p` reconstructed by integrating the **actual moist density** `g·ρ·dz`, top-anchored to the actual model-top pressure (`mpas_atmphys_interface.F:525-544`), not its base state; cloud paths `q·Δp_hyd/g`.

### The point

Every reference model takes the radiative **air mass** from the *actual* state. Breeze takes temperature and moisture from the live state but pressure and density from the fixed reference — `pᶜ = pᵣ` and the dry-air column mass `Δp = pᶠₖ − pᶠₖ₊₁` from the reference pressure gradient (`rrtmgp_shared_utilities.jl:24,92,100`), and cloud water path `ρᵣ · qˡ · Δz` (`all_sky_radiative_transfer_model.jl:316,351`). So under a warm updraft, a cold pool, or the ordinary diurnal density swing of the PBL, the absorber amounts and liquid/ice water paths are computed against a column mass that never changes, even though the temperature does.

### Why the reference was a reasonable reach — and what WRF/MPAS do instead

The concerns this PR is protecting against are real: a compressible core's raw prognostic pressure carries acoustic/nonhydrostatic fluctuations and can be non-monotonic in z, breaking RRTMGP's `Δp/g` math (the source of the negative-source/DomainError clamps already in the kernel). The reference is hydrostatic and monotonic by construction — but it fixes those by discarding the actual mass, which is too blunt.

WRF and MPAS solve the *same* monotonicity problem without a reference: **hydrostatically re-integrate `p_lev` from the actual moist density** (`g·ρ·dz`, top-anchored). That removes only the nonhydrostatic/acoustic component, leaving a clean monotonic pressure that still reflects the true, time-varying mass. So the physically-correct target for Breeze:

- `p_lev` ← hydrostatic integral of the **actual** moist density (not `reference_pressure`)
- cloud water path ← **actual** `ρ` × condensate fraction × `Δz` (not `reference_density`)
- keep actual `T` and actual moisture (already done)

This keeps every guarantee the reference approach bought (monotonic, hydrostatic, DomainError-safe) while matching all three references on the mass.

### Naming

This also argues the accessor shouldn't be `reference_*` at all — radiation doesn't want a *reference*, it wants a *hydrostatically-consistent actual* pressure/density. Something like `hydrostatic_pressure(dynamics)` / `hydrostatic_density(dynamics)` returning the actual-mass reconstruction, with the current `reference_state.pressure/density` path kept as the anelastic case (where the reference *is* the actual mean, so it's already correct there).

Not a blocker for merging the terrain-following capability — the reference path is stable and gives sensible heating (verified in a coupled MC3E run) — but it's a correctness improvement worth landing before results are used quantitatively. Happy to draft the `hydrostatic_*` reconstruction if we agree on the direction.

glwagner:
## Correctness review, expanded: six reference models, and a correction

Following up with two more compressible cores — **ICON** (the closest structural twin to Breeze: nonhydrostatic compressible *with* an explicit reference atmosphere `theta_ref`/`exner_ref`/`rho_ref`) and **FV3/GFS** (mass-coordinate nonhydrostatic, operational). The survey is now six models and it is unanimous:

| Model | Core | Layer pressure | Air **mass** (absorbers / water paths) | Temp | Static reference in radiation? |
|---|---|---|---|---|---|
| ClimaAtmos | compressible | actual `p = ρRₘT` | actual `ρ` | actual | No |
| WRF-ARW | compressible, terrain-following | hydrostatic `p_hyd` from **actual** total mass `MUT` | actual (`Δp/g`) | actual | No |
| MPAS-A | nonhyd. compressible, split base/pert | hydrostatic, integrated from **actual** moist `ρ` | actual `ρ` | actual | No (base state exists, unused) |
| **ICON** | **nonhyd. compressible + reference atmosphere** | **actual** diagnosed `pres` (hydrostatic integ. of prognostic Exner) | actual `ρ`/`dp` | actual | **No — reference confined to dynamics** |
| FV3/GFS | nonhyd., mass coordinate | hydrostatic `prsi = ptop + Σδp` (mass) | actual (`δp/g`) | actual | No |
| **Breeze (this PR)** | nonhyd. compressible + reference atmosphere | **reference** pressure | **reference** density | actual | **Yes** |

**ICON is the decisive precedent.** It has Breeze's exact architecture — a compressible core with a reference atmosphere used for the pressure-gradient/buoyancy — and a grep of every reference-field name (`theta_ref`, `exner_ref`, `rho_ref*`, `t0`, …) across *all* its radiation sources (`mo_nwp_ecrad_interface.f90`, `mo_nwp_rrtm_interface.f90`, `mo_nwp_ecrad_utilities.f90`, the RTE-RRTMGP dir) returns **zero hits**. Radiation is handed the actual state, re-diagnosed each radiation step from the prognostic `exner`/`rho`/`theta_v` (`mo_nwp_ecrad_interface.f90:385`, `mo_nh_diagnose_pres_temp.f90:286-290`). The reference atmosphere lives only in the dynamical core.

**On monotonicity — a correction to my earlier comment.** I speculated the reference was chosen to dodge non-monotonic/noisy compressible pressure. Two things: (a) that rationale isn't documented in Breeze — radiation used `reference_state.pressure` *before* this PR; the actual origin is the **anelastic** heritage of the ext, where reference density genuinely *is* the dynamics' density (`dynamics_density(::AnelasticDynamics) = reference_state.density`), and the pattern was carried onto the compressible core where it becomes an approximation. And (b) the sound-wave concern is real, but every reference model that faces it solves it on the **actual** state, never with a static reference:
- **ICON** diagnoses pressure *"by vertical integration of the hydrostatic equation because the physics schemes actually need the air mass represented by a given model layer"* and integrates surface pressure from the third-lowest level *"to avoid contamination by sound-wave activity in the presence of strong latent heating"* (`mo_nh_diagnose_pres_temp.f90:273-276, 262-263`) — hydrostatic integration of the actual Exner.
- **FV3** clamps the layer-mean pressure to stay monotonic within its hydrostatic interface pressures (`atmosphere.F90:2301-2306`, "Ensure subgrid MONOTONICITY of Pressure").
- **WRF/MPAS** hydrostatically integrate the actual density.

So the monotonicity argument, to whatever extent it applies to Breeze's split-explicit acoustics, points *toward* hydrostatic reconstruction of the actual state, not toward a frozen reference.

**Recommendation stands, now six-for-six:** feed radiation the actual air mass. For a terrain-following compressible core the ICON/WRF/MPAS pattern is the natural fit — a `hydrostatic_pressure`/`hydrostatic_density` reconstruction from the actual moist density (one O(Nz) column integral per radiation call, negligible next to the RRTMGP solve, and no more than the reference path already pays once at construction) — keeping actual `T` and actual moisture, which Breeze already does. Anelastic keeps `reference_state.*`, correctly. Happy to draft it.

glwagner:
## Correction to my two comments above

I overstated this and want to fix the record. I claimed Breeze feeds radiation a *static reference* mass, making it "the outlier" vs. the six surveyed models. That's wrong for the case this PR enables. Tracing `reference_density(::CompressibleDynamics)` (compressible_dynamics.jl:314-319):

```julia
reference_state === nothing || return reference_state.density
terrain_reference_density === nothing ? dynamics.total_density : terrain_reference_density
```

it is "configured reference **if one exists, else the actual diagnosed state**." A nested LAM built without `reference_potential_temperature` (the default — breeze_nested_atmosphere.jl:221) has `ref_spec === nothing`, so both `reference_state` and `terrain_reference_density` are `nothing` (compressible_dynamics.jl:210-211, 237-238), and the accessor returns **`dynamics.total_density`** — the actual per-step total density; pressure likewise returns `dynamics.pressure`, the actual diagnosed pressure. (The `dynamics.density`→`total_density` fix earlier in this PR was in exactly that fallback branch, and it fired in a real terrain-compressible run — proof the terrain case reaches it.)

So on a terrain-following compressible LAM, Breeze **already** feeds radiation the actual air mass, consistent with ClimaAtmos/WRF/MPAS/ICON/FV3. The static reference is used only for (a) anelastic — where `reference_state.density` *is* the dynamics density, correct — and (b) compressible runs that explicitly configure a `reference_potential_temperature` (idealized LES), where reference ≈ actual mean so the approximation is benign, and which isn't the large-anomaly regime anyway.

Net: the survey's "use actual mass" conclusion is already met by the fallback here; my "correctness gap / outlier" framing doesn't apply to this PR. The two genuine residual points are smaller: (1) the **naming** — `reference_density` returning the *actual* density in the common path is confusing, and a name like `radiation_density`/`layer_density` (or documenting the fallback prominently) would help; (2) optionally, whether a compressible run that *does* configure a reference should also use actual mass for radiation rather than the static reference. Apologies for the noise on the earlier comments.

glwagner:
Pushed the fix from the discussion: radiation now reads the **total** thermodynamic state, and the `terrain_reference_*` fallthrough is gone.

```julia
reference_pressure(::CompressibleDynamics) = dynamics.pressure       # actual diagnosed total
reference_density(::CompressibleDynamics)  = dynamics.total_density   # moist total
reference_pressure(::AnelasticDynamics)    = reference_state.pressure # anelastic's only total
reference_density(::AnelasticDynamics)     = reference_state.density
```

Rationale: RRTMGP needs the total pressure and total moist density. A compressible atmosphere carries both directly (`dynamics.pressure`, `dynamics.total_density`) regardless of terrain or whether a reference profile is configured, so radiation should always use them. `terrain_reference_density`/`_pressure` are a *pressure-gradient* device for the well-balanced terrain dynamics — not the air mass — and were only reaching radiation as an accidental piggyback: configuring a reference to get well-balanced dynamics would silently switch radiation from the actual mass to a static reference. That branch is removed. Anelastic keeps `reference_state.*`, which is correct there because the anelastic model has no thermodynamic pressure other than its reference.

Verified the accessors are radiation-only (the momentum closure binds `dynamics_density`, not these), so this doesn't touch the dynamics.

Follow-up (separate issue incoming): the `reference_pressure`/`reference_density` names are now misleading — they return the *actual* total, not a reference — and the flat-vs-terrain reference split itself is redundant. That cleanup is orthogonal to this correctness fix.

ext/BreezeRRTMGPExt/all_sky_radiative_transfer_model.jl
glwagner: I opened an issue about fixing the multiple reference states: #862

glwagner:
Pushed a merge of current `main` into this branch (no textual conflicts beyond the
`dynamics_pressure` docstring/export union with the `thermodynamic_pressure` collapse; #890's
upwelling-shortwave change composes with the state routing here).

Empirical data point from a production case (NumericalEarth's December-2025 atmospheric-river
hindcast — ERA5-nested compressible on a terrain-following grid, 540×315×50, all-sky RRTMGP,
hourly solves for 72 simulated hours): the case runs **stably on both main and this branch**.
On main the defect manifests as a quiet bias — the cloud state is built on
`dynamics.reference_state.density` (a static column) instead of the evolving total density, so
water paths and heating rates are scaled by the reference-vs-actual mismatch, largest over steep
terrain — rather than as a crash. That makes this fix a correctness issue for the heating rates
more than a stability one, at least in that configuration.

glwagner (approved):

PR #794, opened 2026-06-23
Damp the θ-flux divergence (fix split-explicit cold-start blow-up)

### LINKED ISSUE #793: Split-explicit cold start blows up: horizontal divergence damping injects a spurious ∝α/Δτ force on unbalanced states
## Summary

The split-explicit (acoustic-substepping) `CompressibleDynamics` integrator blows up on an **unbalanced cold start** — interior horizontal winds grow far beyond the flow — while the **explicit** integrator stays bounded with the *same* initial conditions, grid, and Δt. The cause is the Klemp-2018 horizontal divergence damping (`ThermalDivergenceDamping`, default `coefficient = 0.1`).

## Reproducer (self-contained, CPU, ~1 min)

An isothermal hydrostatic atmosphere at rest plus a horizontal warm patch, on a coarse large-Δx grid (Δx ≈ 9 km, like a limited-area model):

```julia
# Self-contained reproducer: split-explicit cold-start blow-up from the
# horizontal divergence damping (ThermalDivergenceDamping).
#
# An isothermal hydrostatic atmosphere at REST with a horizontal warm patch
# (out of horizontal balance). On a coarse, large-Δx grid (Δx ≈ 9 km, like a
# limited-area model) the split-explicit integrator's divergence damping injects
# a spurious horizontal force that blows up max|u|, while the explicit
# integrator (no divergence damping) stays bounded — same IC, same Δt-physics.
#
# Run with a Breeze checkout active:  julia --project=<env> divdamp_repro.jl
using Breeze
using Breeze: dry_air_gas_constant
using Breeze.AtmosphereModels: thermodynamic_density
using Oceananigans
using Oceananigans.TimeSteppers: update_state!, time_step!
using Printf

const T₀ = 250.0; const G = 9.80665; const CPD = 1005.0
θ_isothermal(z) = T₀ * exp(G * z / (CPD * T₀))

const Nx = 16; const Ny = 16; const Nz = 40
const Lx = 150e3; const Ly = 150e3; const Lz = 20e3   # Δx = Δy ≈ 9.4 km

function build_model(scheme)
    grid = RectilinearGrid(CPU(); size = (Nx, Ny, Nz), halo = (5, 5, 5),
                           x = (0, Lx), y = (0, Ly), z = (0, Lz),
                           topology = (Periodic, Periodic, Bounded))
    constants = ThermodynamicConstants()
    td = scheme === :explicit ? ExplicitTimeStepping() : SplitExplicitTimeDiscretization()
    dyn = CompressibleDynamics(td; reference_potential_temperature = θ_isothermal,
                               surface_pressure = 1e5, standard_pressure = 1e5)
    return scheme === :explicit ?
        AtmosphereModel(grid; dynamics = dyn, thermodynamic_constants = constants) :
        AtmosphereModel(grid; dynamics = dyn, thermodynamic_constants = constants,
                        timestepper = :AcousticRungeKutta3)
end

# Rest state = discrete hydrostatic reference, plus a horizontal warm patch in ρθ.
function set_initial_condition!(model; Δθ = 5.0)
    ref = model.dynamics.reference_state
    Rᵈ  = dry_air_gas_constant(model.thermodynamic_constants)
    parent(model.dynamics.density) .= parent(ref.density)
    ρθ = thermodynamic_density(model.formulation)
    parent(ρθ) .= parent(ref.pressure) ./ (Rᵈ .* parent(ref.exner_function))

    ρ  = interior(model.dynamics.density)
    rθ = interior(ρθ)
    xc = Lx/2; yc = Ly/2; Lpatch = 30e3
    for i in 1:Nx, j in 1:Ny
        x = (i - 0.5) * Lx / Nx
        y = (j - 0.5) * Ly / Ny
        patch = Δθ * exp(-((x - xc)^2 + (y - yc)^2) / Lpatch^2)
        for k in 1:Nz
            rθ[i, j, k] += ρ[i, j, k] * patch
        end
    end
    fill!(parent(model.velocities.u), 0)
    fill!(parent(model.velocities.v), 0)
    fill!(parent(model.velocities.w), 0)
    update_state!(model)
end

function run_case(scheme; Δt, stop_time)
    model = build_model(scheme)
    set_initial_condition!(model)
    mu() = maximum(abs, interior(model.velocities.u))
    Nsteps = round(Int, stop_time / Δt)
    for n in 1:Nsteps
        time_step!(model, Δt)
        isfinite(mu()) || (return Inf)
    end
    return mu()
end

stop_time = 30.0
u_split = run_case(:split;    Δt = 1.0, stop_time = stop_time)
u_expl  = run_case(:explicit; Δt = 0.1, stop_time = stop_time)
@printf("\nt=%.0fs:  SPLIT max|u| = %.4g    EXPLICIT max|u| = %.4g\n", stop_time, u_split, u_expl)
println(u_split > 5 * max(u_expl, 1) ? "REPRODUCED: split blows up vs explicit" : "split ≈ explicit (bounded)")
```

Output on current `main`:

```
t=30s:  SPLIT max|u| = 12.5    EXPLICIT max|u| = 1.5
REPRODUCED: split blows up vs explicit
```

Split grows ~8× past explicit and keeps growing; explicit stays bounded.

## Root cause

`_thermal_divergence_damping!` damps the (ρθ)′ tendency `(ρθ)′ − (ρθ)′ˢ⁻` as a proxy for `−Δτ·θᴸ·∇·m`, with diffusivity `γ = α·Δx²/Δτ`. The `Δτ` cancels **only** when that change is divergence-dominated. On an unbalanced state the (ρθ)′ change is O(1) — driven by the hydrostatic / pressure-gradient adjustment, not by divergence — so the proxy leaves a spurious horizontal force `∝ α/Δτ`. The explicit path has no divergence damping, hence stays bounded.

Bisection (on the reproducer above and a 3-D ERA5-driven slab) pins it down: the blow-up is split-specific, flow-independent (grows from rest), terrain-independent, off-centering- and substep-count-independent, scales with `1/Δτ` (worse for smaller substeps), and **`coefficient = 0` makes split ≡ explicit bit-for-bit**.

## Fix

Damp the **true** horizontal mass divergence `δ = ∂x(ρu)′ + ∂y(ρv)′` (terrain-aware `div_xyᶜᶜᶜ`) directly, as a Δτ-independent Laplacian diffusion `Δ(ρu)′ = α·Δx²·∂x δ` — equivalent to the current scheme for a balanced acoustic wave, robust for the transient. With the fix the reproducer gives `SPLIT 1.45 ≈ EXPLICIT 1.5`.

Fix on branch `eq/divergence-damping-fix` (PR to follow).

ewquon:
Fixes #793.

The Klemp–Skamarock–Ha (2018) horizontal divergence damping (`ThermalDivergenceDamping`) used the `(ρθ)′` tendency `(ρθ)′ − (ρθ)′ˢ⁻` as the divergence proxy, with diffusivity `γ = α·Δx²/Δτ`. On an unbalanced cold start that tendency is O(1) — driven by the hydrostatic/pressure-gradient adjustment, not by divergence — so the `Δτ` stops cancelling and the damper leaves a spurious `∝ α/Δτ` horizontal force that blows up the split-explicit integrator (interior winds far past the flow). Explicit stepping has no divergence damping and stays clean.

**Fix:** compute the acoustic (Θ-equation) divergence directly — `δ = ∇·(θᴸ 𝐦′) = ∂x(θᴸ(ρu)′) + ∂y(θᴸ(ρv)′)`, per KSH18 eq. (36) — and damp it as a Δτ-**independent** Laplacian diffusion `Δ(ρu)′ = α·Δx²·∂x δ / θᴸ`. This is equivalent to the previous scheme for a balanced acoustic wave but dissipative and `1/Δτ`-free for the transient. Two kernels (`_compute_horizontal_divergence!` → halo fill → `_thermal_divergence_damping!`) avoid a read/write race; the damping scratch field is renamed `previous_density_potential_temperature_perturbation` → `horizontal_divergence`. The diffusion-number stability bound (`α ≤ 0.25`) is unchanged.

**Verification:**
- Reproducer from #793 (isothermal rest + horizontal warm patch, Δx ≈ 9 km): split `max|u|` 12.5 → 1.45 ≈ explicit 1.5.
- 3-D ERA5-driven slab (real terrain; from rest and with the jet): split now tracks explicit (`max|u|` 37 vs the previous 97).
- New tests (`test/acoustic_substepping.jl`): the damper reduces `‖∇·(θᴸ𝐦)′‖` on flat **and** terrain-following grids, and the diffusivity is `α·Δx²` (no `1/Δτ`). 7/7 pass.

A separate, smaller top-lid `w`-mode (≈10 vs explicit ≈4) remains — a distinct follow-up, not addressed here.

### DIFF
--- a/src/CompressibleEquations/acoustic_substepping.jl
+++ b/src/CompressibleEquations/acoustic_substepping.jl
@@ -147,8 +147,8 @@ Per-column scratch (column kernel only):
 
 - `density_predictor`, `density_potential_temperature_predictor`: explicit
   predictors built before the implicit vertical solve.
-- `previous_density_potential_temperature_perturbation`: ``η`` from the
-  previous substep, used by Klemp 2018 damping.
+- `horizontal_divergence`: scratch for the horizontal mass divergence
+  ``∂x(ρu)′ + ∂y(ρv)′``, recomputed each substep for the Klemp 2018 damping.
 
 Time-averaged velocity for non-acoustic scalar transport (WRF/MPAS-style
 dynamics-transport split):
@@ -204,7 +204,7 @@ struct AcousticSubstepper{N, FT, D, AD, US, CF, MP, TAV, GT, TS}
 
     density_predictor :: CF
     density_potential_temperature_predictor :: CF
-    previous_density_potential_temperature_perturbation :: CF
+    horizontal_divergence :: CF
 
     # Time-averaged velocities for non-acoustic scalar advection (WRF/MPAS
     # dynamics-transport split).
@@ -250,7 +250,7 @@ Adapt.adapt_structure(to, a::AcousticSubstepper) =
                         w = adapt(to, a.momentum_perturbation.w)),
                        adapt(to, a.density_predictor),
                        adapt(to, a.density_potential_temperature_predictor),
-                       adapt(to, a.previous_density_potential_temperature_perturbation),
+                       adapt(to, a.horizontal_divergence),
                        (u = adapt(to, a.time_averaged_velocities.u),
                         v = adapt(to, a.time_averaged_velocities.v),
                         w = adapt(to, a.time_averaged_velocities.w)),
@@ -314,7 +314,7 @@ function AcousticSubstepper(grid, split_explicit::SplitExplicitTimeDiscretizatio
 
     density_predictor                                = CenterField(grid)
     density_potential_temperature_predictor          = CenterField(grid)
-    previous_density_potential_temperature_perturbation = CenterField(grid)
+    horizontal_divergence = CenterField(grid)
 
     # Substep-averaged velocities for scalar transport.
     time_averaged_velocities = (u = XFaceField(grid),
@@ -348,7 +348,7 @@ function AcousticSubstepper(grid, split_explicit::SplitExplicitTimeDiscretizatio
                               momentum_perturbation,
                               density_predictor,
                               density_potential_temperature_predictor,
-                              previous_density_potential_temperature_perturbation,
+                              horizontal_divergence,
                               time_averaged_velocities,
                               slow_vertical_momentum_tendency,
                               vertical_solver)
@@ -873,8 +873,8 @@ function initialize_stage_perturbations!(substepper, model, Uᴸ_outer)
     grid = model.grid
     arch = architecture(grid)
 
-    # Auxiliary workspaces: zero.
-    fill!(parent(substepper.previous_density_potential_temperature_perturbation), 0)
+    # Auxiliary workspaces: zero. (`horizontal_divergence` is recomputed each
+    # substep before the damping, so it needs no stage-init zero.)
     fill!(parent(substepper.density_predictor), 0)
     fill!(parent(substepper.density_potential_temperature_predictor), 0)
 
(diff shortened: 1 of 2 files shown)


OsKnoth:
I could not follow your arguments, note that the difference is in the small time step and could not be O(1).

glwagner:
> I could not follow your arguments, note that the difference is in the small time step and could not be O(1).

Right, also interestingly an important difference between this implementation and `ThermalDivergenceDamping` is that this implementation uses only the horizontal components of divergence. In other words, replacing the "thermal approximation" with explicit divergence would require the 3D divergence. This can be seen in eq 6 of Klemp et al 2018 (note that this divergence is unfortunately 2D in x-z, but one can generalize to 3D by adding V_y):

image

glwagner:
Another interesting thing is that the Klemp et al 2018 is not a pure divergence damping. See their text around equation 35 (I can post here too). Basically it doesn't damp divergence to zero but rather to the potential temperature tendency. Also, the damping is effectively semi-implicit because the treatment of vertical velocity. As a result `ThermalDivergenceDamping` is numerically identical to divergence that appears in the temperature equation. In contrast with a "direct" method, there is a mixture of time discretizations.

PR #798, opened 2026-06-23
[CI] Run tests with `--quickfail`

giordano:
This may be slightly annoying when there are errors in multiple test files, but the benefit is getting quick feedback about test failures instead of having to wait possibly for a long time for all other jobs to finish, which should hopefully speed up the development cycle.  The only good thing is that we have several independent CI builds, if there are failures in multiple test files there's a good chance we see at least some of them across the different CI builds, with the caveat we have to look to all the logs to find them all.


Pangoraw (approved):

giordano:
Oh, I see Reactant v0.2.267 is among us now.....

PR #802, opened 2026-06-24
Set `SROA_ATTRIBUTOR` to true to resolve floating point exception

### LINKED ISSUE #801: NaNs & Floating point exception following Reactant@0.2.267
> ```
> │ [2760] signal 8 (1): Floating point exception
> │ in expression starting at /__w/Breeze.jl/Breeze.jl/test/reactant/advection_diffusion_ad.jl:116
> └ Allocations: 437508571 (Pool: 437485389; Big: 23182); GC: 219
> ```
> 
> Fun 😅

Ok, Reactant v0.2.267 has this issue with GPUCompiler v1.22.1-3 (and with v1.22.4 it just hangs).  CC @Pangoraw

_Originally posted by @giordano in https://github.com/NumericalEarth/Breeze.jl/issues/800#issuecomment-4788716786_

giordano:
Fix #801.


.github/workflows/CI.yml
```diff
@@ -174,6 +174,10 @@ jobs:
               # We use earlyoom only on the Linux machines
               earlyoom -m 3 -s 100 -r 300 --prefer 'julia' &
           fi
+
+          export TMPDIR="${{ github.workspace }}/tmp"
+          mkdir -p "${TMPDIR}"
+
```
giordano, line 180: Memo to self: take this commit out to its own PR.

dkytezab:
I suspect that with that loss the gradient simply folds away any computation involving the model

test/reactant/advection_diffusion_ad.jl
```diff
@@ -10,6 +10,10 @@ using GPUArraysCore: @allowscalar
 using Enzyme
 using Test
 
+if @show(get(ENV, "GITHUB_ACTIONS", "false") == "true")
+    Reactant.MLIR.IR.DUMP_MLIR_ALWAYS[] = true
```
Pangoraw, line 14: We can also try this since it's one of the change from 0.2.267 that changes the IR.
```suggestion
    Reactant.MLIR.IR.DUMP_MLIR_ALWAYS[] = true
    Reactant.Compiler.SROA_ATTRIBUTOR[] = true
```

giordano: That did the the trick, somehow

.github/workflows/CI.yml
giordano: This still doesn't work for some reason (no files are found in this directory), but let's postpone investigation to later, this was broken before this anyway.

PR #805, opened 2026-06-25
Add benchmarks for WENO scalar, model tendency

dkytezab:
We will use this in large part for Reactant-side kernel perf optimization, cc @Pangoraw


dkytezab:
Looking at https://github.com/NumericalEarth/Breeze.jl/actions/runs/28196903527/job/83526006409?pr=805#step:8:778, do we need a different optimization flag here? Maybe `:before_raise`?

dkytezab:
```
Benchmark                                                       Float    Backend           Grid    Time/Step    Steps/s        Points/s
----------------------------------------------------------------------------------------------------------------------------------
ScalarTendency_256x256x128_F32_WENO3_PPB_vanilla              Float32    vanilla    256×256×128     0.7252 ms    1378.85        1.16e+10
ScalarTendency_256x256x128_F32_WENO3_PPB_reactant             Float32   reactant    256×256×128     0.6996 ms    1429.39        1.20e+10
ScalarTendency_256x256x128_F32_WENO5_PPB_vanilla              Float32    vanilla    256×256×128     1.1618 ms     860.70        7.22e+09
ScalarTendency_256x256x128_F32_WENO5_PPB_reactant             Float32   reactant    256×256×128     0.9582 ms    1043.60        8.75e+09
ScalarTendency_256x256x128_F32_WENO7_PPB_vanilla              Float32    vanilla    256×256×128     1.7675 ms     565.78        4.75e+09
ScalarTendency_256x256x128_F32_WENO7_PPB_reactant             Float32   reactant    256×256×128     1.8012 ms     555.18        4.66e+09
ScalarTendency_256x256x128_F32_WENO9_PPB_vanilla              Float32    vanilla    256×256×128     2.6104 ms     383.09        3.21e+09
ScalarTendency_256x256x128_F32_WENO9_PPB_reactant             Float32   reactant    256×256×128    51.5393 ms      19.40        1.63e+08
```
Pretty precipitous performance falloff for WENO9 with reactant. Seems like stencil-size-scaling is worse with reactant than with vanilla.

glwagner:
also interesting to see parity for lower order !

glwagner:
is this cpu or gpu? I think we should test both

dkytezab:
this is on gpu, we can look into setting up a cpu job for this benchmark. in which case do we want to run all reactant benchmarks on cpu? or at that point then all benchmarks generally?

giordano:
Should we do a lot more steps, so that the runtime is at least a few seconds?   These are way too fast.

benchmarking/src/tendency_profiling.jl
```diff
@@ -0,0 +1,127 @@
...
+        gpu_memory_used,
+        metadata,
+    )
+
+    if verbose
+        @info "  Results:"
+        @info "    Time per call: $(@sprintf("%.6f", time_per_call_seconds)) s"
+        @info "    Grid points/s: $(@sprintf("%.2e", grid_points_per_second))"
+        @info "    Compile time: $(@sprintf("%.3f", compile_time_seconds)) s"
+    end
```
giordano: Can we report also the total time, like in https://github.com/NumericalEarth/Breeze.jl/blob/d79c4ce40b558388d08fce1b040f5426b626bb99/benchmarking/src/utils.jl#L169?  Also, is there a way to reuse the code and reduce duplication and slight differences?  For example here there's "time per call", while there we have "time per step": https://github.com/NumericalEarth/Breeze.jl/blob/d79c4ce40b558388d08fce1b040f5426b626bb99/benchmarking/src/utils.jl#L170

glwagner:
Also wonder if a momentum tendency would be more challenging. Are the overall benchmarks so close for vanilla and Reactant?

giordano:
> Are the overall benchmarks so close for vanilla and Reactant?

They are in the same ballpark, but vanilla is a bit faster at the moment: https://numericalearth.github.io/BreezeBenchmarks/#benchmark-cbl--dynamics--compressible-explicit--microphysics--1m-mixednonequilibrium--float32-

dkytezab:
We can also get BFloat16s in this bad boy...

dkytezab:
```
Benchmark                                                       Float    Backend           Grid    Time/Step    Steps/s        Points/s
----------------------------------------------------------------------------------------------------------------------------------
ScalarTendency_256x256x128_F32_WENO9_PPB_vanilla              Float32    vanilla    256×256×128     2.6562 ms     376.48        3.16e+09
ScalarTendency_256x256x128_F32_WENO9_PPB_reactant             Float32   reactant    256×256×128    51.5078 ms      19.41        1.63e+08
ScalarTendency_256x256x128_BF16_WENO9_PPB_vanilla            BFloat16    vanilla    256×256×128     3.7656 ms     265.56        2.23e+09
ScalarTendency_256x256x128_BF16_WENO9_PPB_reactant           BFloat16   reactant    256×256×128     2.8241 ms     354.10        2.97e+09
```
BF16 Reactant significantly outstrips FP32 Reactant. Much more fusion based off traces (5 kernels post xla for the former, ~60+ for the latter)

dkytezab:
@giordano how would you feel about splitting up the benchmarking page into two separate windows, specifically into (a) vanilla only benchmarks and (b) benchmarks that involve reactant at all. If the number of benchmarks grows this would help keep the amount of visual info manageable

giordano:
Interesting.  I think it's worth exploring this layout, but there's also a value in a direct comparison between the two backends.  @glwagner opinions?

benchmark_model_tendency.jl
Pangoraw, line 1: is this file intended in the PR?

glwagner:
@dkytezab Reactant should see a big speed up for BF16 because it can actually use the compute right? (whereas for vanilla, only the data is stored in BF16 but compute is F32).

cc @wsmoses who may be interested. We are seeing parity performance for vanilla and reactant which is quite amazing news. I think this implies that performance differences in the full app have to do with too much kernel fusion? Tentative idea?

glwagner:
> this is on gpu, we can look into setting up a cpu job for this benchmark. in which case do we want to run all reactant benchmarks on cpu? or at that point then all benchmarks generally?

If the cost is ok, I don't see a downside of running all the benchmarks on CPU.

glwagner:
> Interesting. I think it's worth exploring this layout, but there's also a value in a direct comparison between the two backends. @glwagner opinions?

I think for the second Reactant window, you can also include vanilla data for reference. I think at this stage in terms of comparisons we want this information

* vanilla page: how performance is affected by grid size, microphysics, advection scheme, float type
* reactant page: reactant vs vanilla comparisons, across similar axes as vanilla page but out of necessity, showing fewer datapoints

dkytezab:
> cc @wsmoses who may be interested. We are seeing parity performance for vanilla and reactant which is quite amazing news. I think this implies that performance differences in the full app have to do with too much kernel fusion? Tentative idea?

we should still figure out why Reactant perf scales worse with WENO order than vanilla. I'm curious why the jump from WENO7 to 9 is so high?

> I think for the second Reactant window, you can also include vanilla data for reference. I think at this stage in terms of comparisons we want this information
> 
> vanilla page: how performance is affected by grid size, microphysics, advection scheme, float type
reactant page: reactant vs vanilla comparisons, across similar axes as vanilla page but out of necessity, showing fewer datapoints

I back this. I think we should try to come up with a way of quantifying Reactant full model benchmarks against single kernel benchmarks, probs by number of kernel—then we can see "gestalt performance dropoff"

benchmark_model_tendency.jl
dkytezab: certainly not!

glwagner:
> > cc @wsmoses who may be interested. We are seeing parity performance for vanilla and reactant which is quite amazing news. I think this implies that performance differences in the full app have to do with too much kernel fusion? Tentative idea?
> 
> we should still figure out why Reactant perf scales worse with WENO order than vanilla. I'm curious why the jump from WENO7 to 9 is so high?

Could it be explained by the fact that whatever fancy optimizations we get are somehow controlled by a heuristic, so with wide enough stencil it stops working? I guess this isn't much of an explanation, but I'd wonder whether any optimization have magic numbers in them somewhere.

dkytezab:
Definitely plausible. @giordano is there a way to preview the benchmarks?

wsmoses:
it would also be interesting to benchmark reactant raise=true and reactant raise=false

giordano:
Yes, it's a hack (that will have to be reverted before merge), but it's doable:

* https://github.com/NumericalEarth/Breeze.jl/blob/67db12e6683350241cf786fbfd536903ec8e89d0/.github/workflows/Benchmarks.yml#L282-L283 swap the comments in these two lines
* https://github.com/NumericalEarth/Breeze.jl/blob/67db12e6683350241cf786fbfd536903ec8e89d0/.github/workflows/Benchmarks.yml#L290-L291 you know the drill now: swap the comments
* create the subdir in https://github.com/NumericalEarth/BreezeBenchmarks/tree/gh-pages and copy the index.html file in there

dkytezab:
I agree that we should compare with and without raising, but our matrix of toggles is getting very large (raise, unraised, vanilla) x (weno3, 5, 7, 9) x (FP32, BFP16). May need some better UI where the user can toggle certain configs

dkytezab:
Got it, could you give me edit access to https://github.com/NumericalEarth/BreezeBenchmarks/tree/gh-pages?

dkytezab:
https://numericalearth.github.io/BreezeBenchmarks/previews/PR805/#benchmark-scalartendency
i'm pleased with this.

Pangoraw:
I am surprised that BF16 raise=false and vanilla run at all since we emit codegen for float32 and "cast" the kernel float type in the raise pipeline. Results looks nice otherwise!

giordano:
Why is `raise=false` faster than `raise=true`?

dkytezab:
My understanding is that no raising shields us from bad XLA-side optimizations because we primarily emit stablehlo custom calls

giordano (approved):

PR #806, opened 2026-06-25
Convert CompressibleEquations to dry-density (ρᵈ) evolution

glwagner:
## Summary

Converts the compressible core to prognose **dry-air density ρᵈ** instead of total density, following the MPAS-A dry-coupled formulation.

- `dynamics_density` now returns **ρᵈ** — the *coupling density*: it weights the flux variables (momentum `ρu = ρᵈu`, thermodynamic `ρθ = ρᵈθ`), divides velocity/θ, and is advanced by continuity.
- The **total** air density `ρ = ρᵈ + Σρˣ` is a diagnosed field (`total_air_density`, stored on the dynamics, computed once per `update_state!`) used wherever total mass enters the physics: moisture mass-fraction recovery, scalar/water advection carriers, the EOS, buoyancy, and saturation.
- **Moisture stays in mass fractions throughout — no mixing ratios.** Anelastic is unaffected (`total_air_density === dynamics_density`), which is why the anelastic LES suites pass unchanged.

## Why

With *total* density prognostic, mass consistency requires a hydrometeor sedimentation flux in the `ρ`, `ρqᵗ`, and energy equations computed with an *identical* discrete operator — the WENO-inconsistency that stalled #458/#614. With **ρᵈ** prognostic, dry air has no source and no fall flux, so sedimentation lives only in the condensate equations and total continuity holds by summation (it is never discretized).

## Key changes

- Rename the prognostic density field/key to `ρᵈ`/`:ρᵈ`; **keep generic flux keys** (`ρu`, `ρθ`, `ρqˣ`) — they denote coupling-weighted quantities.
- Add the diagnosed `total_density` field + `total_air_density` accessor + compute kernel.
- Two-density EOS: `θ = ρθ/ρᵈ` (coupling), inversion + `p = ρRᵐT` on total ρ.
- Buoyancy on total ρ (explicit + acoustic-substepper slow tendency; the substep loop uses the dry perturbation `ρ′ = ρᵈ′` with the moisture loading entering once via the stage-entry total).
- Route moisture recovery, water/tracer advection carriers, both formulation tendencies, microphysics state/fractions, and the substepper γᵐRᵐ through total ρ; velocity / continuity / θ- and energy-coupling stay on ρᵈ.
- `set!`: `ρ=` means **total** density (backs out `ρᵈ = ρ·qᵈ` so moist columns initialize in hydrostatic balance); `ρᵈ=` sets dry density directly.
- Fix the potential-temperature and number-concentration diagnostics to use total ρ.

## Validation (CPU)

- **Dry: bit-for-bit** — `acoustic_substepping`, `dynamics`.
- **Moist correct** — `compressible_saturation_adjustment`, `instantaneous_precipitation`, `cloud_microphysics_1M`, `cloud_microphysics_2M`.
- `diagnostics`, `number_concentration`, `quality_assurance` (ExplicitImports + Aqua).

## Follow-ups (not in this PR)

- **`terrain_following` full suite**: its conversion fixes (own EOS launch + two buoyancy paths + a positional-type-param dispatch shift from the new struct field) are in and the path is smoke-confirmed (nonzero `w̃`, finite `T`, `total ≡ dry` for explicit + split-explicit), but the full 152-assertion suite gets earlyoom-killed on the shared head node. Needs confirmation on CI / a larger runner.
- **Energy-flux bottom BC** (`_energy_flux_density`) uses `fields.ρᵈ`; for *moist* compressible it should use total ρ (needs `total_density` exposed in `model_fields`). Edge case (moist + energy-flux BC), not prognostic.


src/PotentialTemperatureFormulations/potential_temperature_tendency.jl
```diff
@@ -234,7 +234,7 @@ end
 
     @inbounds begin
         pᵣ = dynamics_pressure(dynamics)[i, j, k]
-        ρᵣ = dynamics_density(dynamics)[i, j, k]
+        ρᵣ = total_air_density(dynamics)[i, j, k]
```
glwagner: Not sure why "total density" is denoted rho_r here.

src/StaticEnergyFormulations/static_energy_tendency.jl
```diff
@@ -196,7 +196,7 @@ end
 
     @inbounds begin
         pᵣ = dynamics_pressure(dynamics)[i, j, k]
-        ρᵣ = dynamics_density(dynamics)[i, j, k]
+        ρᵣ = total_air_density(dynamics)[i, j, k]
```
glwagner: ditto as above

src/CompressibleEquations/compressible_time_stepping.jl
```diff
@@ -38,6 +38,73 @@ No-op for `CompressibleDynamics` - pressure is computed from the equation of sta
...
+    @inbounds ρ_total[i, j, k] =
+        AtmosphereModels.total_air_density(i, j, k, dry_density, microphysics,
+                                           moisture_density, microphysical_fields)
+end
+
+#####
+##### Initial-condition reconciliation: total density `:ρ` → dry density ρᵈ
+#####
+
+# `set!(model, ρ=ρ_total, …)` puts ρ_total into the dry-density field as a placeholder, so the
```
glwagner: For clarity lets use `rho=total_density` rather than `rho_total` (which in mathematical notation conflictrs with plain "rho", which also stands for total density)

src/CompressibleEquations/compressible_time_stepping.jl
```diff
@@ -38,6 +38,73 @@ No-op for `CompressibleDynamics` - pressure is computed from the equation of sta
...
+# recovery / EOS / buoyancy consume it. It is invariant under saturation adjustment (which
+# conserves total water and dry mass), so a single evaluation per update is correct.
+function AtmosphereModels.compute_total_density!(model::CompressibleModel)
+    grid = model.grid
+    arch = grid.architecture
+    ρ_total = model.dynamics.total_density
+    launch!(arch, grid, :xyz, _compute_total_density!,
+            ρ_total, grid, model.dynamics.dry_density, model.microphysics,
+            model.moisture_density, model.microphysical_fields)
+    fill_halo_regions!(ρ_total)
```
glwagner:
```suggestion
    ρ = model.dynamics.total_density
    launch!(arch, grid, :xyz, _compute_total_density!,
            ρ, grid, model.dynamics.dry_density, model.microphysics,
            model.moisture_density, model.microphysical_fields)
    fill_halo_regions!(ρ)
```

PR #609, opened 2026-04-03
Reland "Test raising one moment microphysics"

giordano:
Reverts #608 and relands #595.  This will need some work to fix the raising failure caused by #587 + #595, but opening the PR not to forget about it

PR #810, opened 2026-06-26
Reduce reproducer for stackoverflow in julia v1.12

giordano:
Taking extra commits out of #609 not to mess up that PR


test/reactant/1m_microphysics_compilation.jl
```diff
@@ -80,27 +77,16 @@ end
...
-            ad_grad = @allowscalar Array(interior(dθ))
-
-            @test loss_val > 0
-            @test isfinite(loss_val)
-            @test maximum(abs, ad_grad) > 0
-            @test !any(isnan, ad_grad)
+            # compiled_grad = Reactant.@compile raise=true raise_first=true sync=true grad_loss(
+            #     model, dmodel, θ_init, dθ_init, Δt, Ns)
+            # dθ, loss_val = compiled_grad(model, dmodel, θ_init, dθ_init, Δt, Ns)
+            Reactant.@compile raise=true raise_first=true sync=true loss(model, θ_init, Δt, Ns)
```
giordano: The stackoverflow is in the call place, not during compilation, so this has to be called. Also, if we have to trust the bot I queried (see the other PR), wwe may need to return also θ_init to trigger the issue

dkytezab: Seems like we need grad as well

giordano:
Interesting, https://github.com/NumericalEarth/Breeze.jl/blob/7f336472fa8951a57229263c418c1a1150d3a7a4/test/reactant/1m_microphysics_compilation.jl still reproduces the issue, without `dθ_init` being returned in the gradient function

giordano:
I'm running out of ideas here.  Amazon Linux seems to set hard limit of stack size to 10240 KiB (unlike every other distro, which leaves that unlimited), but [even after setting the stack size via Docker options](https://github.com/NumericalEarth/Breeze.jl/compare/4badf3420aee41f3d5eaed48e7d30cdff571532a...5a2812d52591db354f8ff1fbefb224bee9f26341) this is still reproducible.

giordano:
Issue identified: it was the task's stack size that was too small.  This consistently happens only on CI, I can't reproduce on other machines, but we're probably close to the default size, and somehow getting over the edge only here on CI.

In any case, a workaround is to manually call [`Task(..., reserved_stack)`](https://docs.julialang.org/en/v1/base/parallel/#Tasks) to set the stack size to a desired value (16 MiB seems to be sufficient in this case)

PR #811, opened 2026-06-29
Add `set!(model; balancer = AdiabaticBalancer(...))` adiabatic initialization

glwagner:
## Summary

Collapses the FV3 `na_init` dynamical-initialization **boilerplate** — the old hand-rolled recipe of building a *separate* twin model, running `balance_adiabatically!`, then grafting fields back (`interior(model.ρu) .= interior(twin.ρu)`, …) — into a single keyword on `set!`:

```julia
set!(model; ρ, u, v, qᵗ, θˡⁱ, balancer = true)                       # auto Δt
set!(model; ρ, u, v, qᵗ, θˡⁱ, balancer = AdiabaticBalancer(Δt = 0.1))  # explicit
```

The balance runs on a twin that **shares all field memory** with the production model (momentum, velocities, densities, ρθ, moisture, temperature, pressure solver, dynamics fields) — identical references, not copies — and steps it **in place**. Because the twin's prognostic arrays *are* the production model's arrays, the balanced state is already in the model when the excursion finishes: **no second field set, and no graft step at all.**

## How the memory-sharing twin works

Only the pieces that must differ for a reversible adiabatic excursion are rebuilt:

- **Time discretization** → swapped via the new `with_time_discretization` (a fresh immutable wrapper around the *same* `CompressibleDynamics` fields). The sponge is always stripped (`without_sponge`) since it is irreversible.
- **Timestepper** → built via `TimeStepper(default_timestepper(twin_dynamics), …)`, its `Gⁿ`/`U⁰` tendency storage **aliasing** the production stepper's same-named arrays (production prognostics are a superset of the twin's; the moisture key is re-mapped from the microphysics name, e.g. `:ρqᵉ`, to the moistureless `:ρqᵛ`). Adds a `U⁰` keyword to both `SSPRungeKutta3` and `AcousticRungeKutta3` to enable this.
- **microphysics → nothing, closure → nothing (incl. the irreversible implicit vertical-diffusion solver), forcing → zeroed, radiation → nothing**, fresh `Clock` (so the balance's clock reset doesn't touch the production clock).

The only residual allocation is one-time NamedTuple/struct construction scratch; the heavy field arrays and both tendency sets are aliased (asserted by `===` tests).

## API

```julia
AdiabaticBalancer(; Δt = nothing, cycles = 1, weight = 2, with_moisture = true,
                 time_stepping = ExplicitTimeStepping())
```

- **`time_stepping`** — the scheme for the balance excursion (sponge always stripped):
  - `ExplicitTimeStepping()` (default): replace with fully-explicit stepping. Memory-minimal (no acoustic substepper; only aliased `Gⁿ`/`U⁰`), cleanly reversible, `Δt` bounded by the vertical acoustic CFL.
  - `nothing`: reuse the model's **native** scheme (e.g. split-explicit) with the sponge stripped — production-consistent numerics and a larger permissible outer `Δt`, at the cost of rebuilding the acoustic substepper's scratch. (Note: the off-centered acoustic solve is dissipative, so this path is only *approximately* reversible — that dissipation acts as a DFI filter.)
  - any other time-discretization object: swapped in as-is.
- `Δt = nothing` auto-derives the vertical acoustic-CFL step (`0.85 · Δz_min / c`).
- `with_moisture` controls the moisture density `ρqᵉ`, which the twin shares and `balance_adiabatically!` nudges in place:
  - `true` (default): `ρqᵉ` relaxes with the other prognostics (more physically consistent).
  - `false`: `ρqᵉ` is snapshotted before the balance and restored after, so it is preserved bit-for-bit and only `(ρ, ρu, ρv, ρw, ρθ)` change — matching the behavior of the old hand-rolled DFI, which grafted back only those five fields.

Supported for `CompressibleDynamics`; other dynamics throw a clear error pointing at `balance_adiabatically!`.

## Tests

`test/adiabatic_balance_set.jl` (auto-discovered) covers: field + tendency aliasing (`===`), correct stripping (microphysics/closure/implicit-solver/radiation; `ExplicitTimeStepping`; separate clock), no-field-set allocation, native time-stepping (`time_stepping = nothing`: split-explicit reused with sponge removed, tendency storage still aliased), the moisture-key remap (`SaturationAdjustment` `:ρqᵉ` → twin `:ρqᵛ`), `resolve_balance_Δt`, the discrete rest-state fixed point (production clock untouched), `with_moisture` both ways, and the non-compressible error.

Validated locally and in CI: tests pass (incl. both GPU runners), existing `balance_adiabatically.jl` / `set_atmosphere_model.jl` still pass, docs build, ExplicitImports (all checks), and Aqua all green.

## Follow-up (not in this PR)

The NumericalEarth `breeze_downscaling_era5` example's hand-rolled DFI block can collapse to `balance = true` once a Breeze release ships.

### DIFF
--- a/test/adiabatic_balance_set.jl
+++ b/test/adiabatic_balance_set.jl
@@ -1,0 +1,149 @@
+#####
+##### Tests for `set!(model; balance = AdiabaticBalance(...))` — in-place adiabatic (FV3 na_init)
+##### initialization on a memory-sharing, stripped twin (`adiabatic_balance_twin`).
+#####
+##### - Memory sharing: the twin aliases every heavy field + the stepper's Gⁿ/U⁰ tendency storage,
+#####   and is correctly stripped (ExplicitTimeStepping, no microphysics/radiation, own clock).
+##### - Moisture-key remap: a SaturationAdjustment production model (:ρqᵉ) maps to the twin's :ρqᵛ.
+##### - resolve_balance_Δt: auto acoustic-CFL step, explicit override round-trips.
+##### - Fixed point: a discrete-balanced rest state is preserved; the production clock is untouched.
+##### - with_moisture: false preserves ρqᵉ bit-for-bit; true lets it relax.
+##### - Edge: non-compressible dynamics throws a clear error.
+#####
+
+using Breeze
+using Breeze: dynamics_density, adiabatic_balance_twin, resolve_balance_Δt, AdiabaticBalance
+using Oceananigans
+using Oceananigans.Grids: minimum_zspacing
+using Oceananigans.TimeSteppers: update_state!
+using Test
+
+const T₀_BS  = 250.0
+const G_BS   = 9.80665
+const CPD_BS = 1005.0
+θ_iso_bs(z) = T₀_BS * exp(G_BS * z / (CPD_BS * T₀_BS))
+
+ρθ_of(model) = Breeze.AtmosphereModels.thermodynamic_density(model.formulation)
+
+# Production-style compressible model: split-explicit WITH an upper sponge and (default) divergence
+# damping — exactly the ingredients the twin must strip for a reversible adiabatic excursion.
+function _build_production(arch; microphysics = nothing, Nx = 8, Ny = 8, Nz = 16, Lz = 10e3, Lh = 100e3)
+    grid = RectilinearGrid(arch; size = (Nx, Ny, Nz), halo = (5, 5, 5),
+                           x = (0, Lh), y = (0, Lh), z = (0, Lz),
+                           topology = (Periodic, Periodic, Bounded))
+    constants = ThermodynamicConstants(eltype(grid))
+    td  = SplitExplicitTimeDiscretization(sponge = UpperSponge(damping_rate = 1/5, depth = 2000.0))
+    dyn = CompressibleDynamics(td; reference_potential_temperature = θ_iso_bs,
+                               surface_pressure = 1e5, standard_pressure = 1e5)
+    return AtmosphereModel(grid; dynamics = dyn, thermodynamic_constants = constants,
+                           microphysics, coriolis = nothing, timestepper = :AcousticRungeKutta3)
+end
+
+function _set_discrete_rest!(model)
+    ref = model.dynamics.reference_state
+    Rᵈ  = Breeze.dry_air_gas_constant(model.thermodynamic_constants)
+    parent(model.dynamics.dry_density) .= parent(ref.density)
+    parent(ρθ_of(model)) .= parent(ref.pressure) ./ (Rᵈ .* parent(ref.exner_function))
+    fill!(parent(model.velocities.u), 0)
+    fill!(parent(model.velocities.v), 0)
+    fill!(parent(model.velocities.w), 0)
+    update_state!(model)
+    return nothing
+end
+
+@testset "set!(; balance) adiabatic initialization" begin
+
+    @testset "twin shares memory and is correctly stripped" begin
+        model = _build_production(default_arch)
+        twin  = adiabatic_balance_twin(model)
+
+        # Every heavy field is the SAME object, not a copy.
+        @test twin.momentum.ρu      === model.momentum.ρu
+        @test twin.momentum.ρv      === model.momentum.ρv
+        @test twin.momentum.ρw      === model.momentum.ρw
+        @test twin.velocities       === model.velocities
+        @test twin.moisture_density === model.moisture_density
+        @test twin.temperature      === model.temperature
+        @test twin.pressure_solver  === model.pressure_solver
+        @test dynamics_density(twin.dynamics) === dynamics_density(model.dynamics)
+        @test ρθ_of(twin)           === ρθ_of(model)
+        # Stepper tendency storage is aliased, not reallocated.
+        @test twin.timestepper.Gⁿ.ρu === model.timestepper.Gⁿ.ρu
+        @test twin.timestepper.U⁰.ρθ === model.timestepper.U⁰.ρθ
+
+        # Stripped for a reversible adiabatic excursion.
+        @test twin.microphysics === nothing
+        @test twin.radiation    === nothing
+        @test twin.dynamics.time_discretization isa ExplicitTimeStepping
+        @test model.dynamics.time_discretization isa SplitExplicitTimeDiscretization
+        @test twin.clock !== model.clock
+
+        # No field SETS allocated: the two aliased tendency sets (Gⁿ+U⁰ ≈ 12 fields) dwarf the
+        # one-time NamedTuple/struct scratch the builder does allocate.
+        one_field_bytes = sizeof(parent(model.momentum.ρu))
+        adiabatic_balance_twin(model)  # compile
+        @test (@allocated adiabatic_balance_twin(model)) < 12 * one_field_bytes
+    end
+
+    @testset "moisture-key remap (SaturationAdjustment :ρqᵉ → twin :ρqᵛ)" begin
+        model = _build_production(default_arch; microphysics = SaturationAdjustment())
+        @test Breeze.AtmosphereModels.moisture_prognostic_name(model.microphysics) == :ρqᵉ
+        twin = adiabatic_balance_twin(model)
+        @test twin.moisture_density === model.moisture_density
+        @test twin.timestepper.Gⁿ.ρqᵛ === model.timestepper.Gⁿ.ρqᵉ
+        @test :ρqᵛ ∈ keys(Oceananigans.prognostic_fields(twin))
+    end
+
+    @testset "resolve_balance_Δt" begin
+        model = _build_production(default_arch)
+        _set_discrete_rest!(model)
+        @test resolve_balance_Δt(AdiabaticBalance(Δt = 0.123), model) == 0.123  # explicit round-trips
+        Δt_auto = resolve_balance_Δt(AdiabaticBalance(), model)               # auto from acoustic CFL
+        @test 0 < Δt_auto < minimum_zspacing(model.grid) / 250                # ≈ Δz/c, c ≳ 300 m/s
+    end
+
+    @testset "fixed point: discrete rest state preserved; production clock untouched" begin
+        model = _build_production(default_arch)
+        _set_discrete_rest!(model)
+        ρ  = dynamics_density(model.dynamics)
+        ρθ = ρθ_of(model)
+        ρ₀  = Array(interior(ρ))
+        ρθ₀ = Array(interior(ρθ))
+
+        set!(model; balance = AdiabaticBalance(Δt = 0.1, cycles = 1))
+
+        @test maximum(abs, Array(interior(ρ))  .- ρ₀)  <= 1e-8 * maximum(abs, ρ₀)
+        @test maximum(abs, Array(interior(ρθ)) .- ρθ₀) <= 1e-8 * maximum(abs, ρθ₀)
+        @test maximum(abs, Array(interior(model.momentum.ρw))) <= 1e-7
+        @test model.clock.iteration == 0
+        @test model.clock.time == 0
+    end
+
+    @testset "with_moisture: false preserves ρqᵉ exactly, true lets it relax" begin
+        # Seed a horizontally-uniform vapor perturbation so the nudge would move ρqᵉ.
+        model = _build_production(default_arch; microphysics = SaturationAdjustment())
+        _set_discrete_rest!(model)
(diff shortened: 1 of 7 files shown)


test/adiabatic_balance_set.jl
```diff
@@ -0,0 +1,173 @@
...
+        @test twin.radiation    === nothing
+        @test twin.dynamics.time_discretization isa ExplicitTimeStepping
+        @test model.dynamics.time_discretization isa SplitExplicitTimeDiscretization
+        @test twin.clock !== model.clock
+
+        # No field SETS allocated: the two aliased tendency sets (Gⁿ+U⁰ ≈ 12 fields) dwarf the
+        # one-time NamedTuple/struct scratch the builder does allocate.
+        one_field_bytes = sizeof(parent(model.momentum.ρu))
+        adiabatic_balance_twin(model)  # compile
+        @test (@allocated adiabatic_balance_twin(model)) < 12 * one_field_bytes
```
ewquon: **Correction (ran it locally).** My earlier `2×` suggestion was wrong — I claimed the builder's scratch was "≪ 1 field," but measuring it shows `adiabatic_balance_twin` allocates **270,032 B = 4.01 × `one_field_bytes`** (NamedTuple/forcing/stepper construction). So `< 2 × one_field_bytes` *fails*.

The right tightened bound is `8×`: it clears the ~4-field scratch floor, and still trips if a single dropped `Gⁿ`/`U⁰` keyword reallocates one tendency set (~6 fields → ~10 total). The original `12×` only catches a *both-keyword* break (~16 fields); `8×` additionally catches a single-keyword drop. Verified: full file 49/49 pass with this bound.

```suggestion
        # The twin must ALIAS both tendency sets (Gⁿ, U⁰ — 6 fields each), not reallocate them.
        # The builder's one-time NamedTuple/forcing/stepper scratch measures ~4 fields; bound at 8
        # clears that but still trips if a single dropped Gⁿ/U⁰ keyword reallocates one set (~6 more).
        one_field_bytes = sizeof(parent(model.momentum.ρu))
        adiabatic_balance_twin(model)  # compile
        @test (@allocated adiabatic_balance_twin(model)) < 8 * one_field_bytes
```

test/adiabatic_balance_set.jl
```diff
@@ -0,0 +1,173 @@
...
+        set!(model; qᵗ = (x, y, z) -> 1e-3 * exp(-z / 2e3))
+        set!(model; w = (x, y, z) -> 0.01 * sin(2π * z / 2e3))
+
+        ρqᵉ₀ = Array(interior(model.moisture_density))
+        set!(model; balance = AdiabaticBalance(Δt = 0.1, with_moisture = false))
+        @test Array(interior(model.moisture_density)) == ρqᵉ₀   # bit-identical
+
+        set!(model; balance = AdiabaticBalance(Δt = 0.1, with_moisture = true))
+        @test all(isfinite, Array(interior(model.moisture_density)))
+    end
```
ewquon, line 161: **Revised (and verified locally).** My first version was numerically unstable — the analysis state (θ = 300 K against a ~250 K reference base state) was far too out of balance and the excursion went to `NaN` under *every* scheme, including the explicit default. Rebuilt around a near-balanced IC (discrete rest state + a small smooth thermal bump) and the **explicit default** scheme on both paths.

Result with this version: the in-place twin reproduces the hand-stripped graft **bit-for-bit** (`max_reldiff = 0.0`), `ρw` spins up to ≈2e-8 (non-degenerate), and `ρqᵉ` is preserved. Full file: **49/49 pass**.

This also folds in the non-degeneracy assertion and uses the explicit default (the common path) rather than `time_stepping = nothing`.

```suggestion
    end

    @testset "set!(; balance, with_moisture=false) reproduces a hand-stripped explicit graft" begin
        Δt, cycles = 0.5, 2

        # Near-balanced moist analysis shared by both paths: discrete rest state + a small smooth
        # thermal bump (mild hydrostatic imbalance → ρw spins up but stays finite) + subsaturated qᵗ.
        function analysis!(m)
            _set_discrete_rest!(m)
            set!(m; qᵗ = (x, y, z) -> 4e-3 * exp(-z / 3e3))
            δ = CenterField(m.grid)
            set!(δ, (x, y, z) -> 5.0 * exp(-((z - 5e3) / 1.5e3)^2))
            parent(ρθ_of(m)) .+= parent(δ)
            update_state!(m)
            return m
        end

        # Path A — in-place twin via set!(; balance) on a moist model. The DEFAULT ExplicitTimeStepping
        # twin balances only (ρ, ρu, ρv, ρw, ρθ); with_moisture = false preserves ρqᵉ.
        model = _build_production(default_arch; microphysics = SaturationAdjustment())
        analysis!(model)
        ρqᵉ_analysis = Array(interior(model.moisture_density))
        set!(model; balance = AdiabaticBalance(Δt = Δt, cycles = cycles, with_moisture = false))

        # Path B — the previous path: an independent, hand-stripped moistureless model built directly
        # on ExplicitTimeStepping (matching the default twin), balanced via balance_adiabatically!.
        grid = model.grid
        dyn  = CompressibleDynamics(ExplicitTimeStepping(); reference_potential_temperature = θ_iso_bs,
                                    surface_pressure = 1e5, standard_pressure = 1e5)
        ref  = AtmosphereModel(grid; dynamics = dyn,
                               thermodynamic_constants = ThermodynamicConstants(eltype(grid)),
                               microphysics = nothing, coriolis = nothing)
        analysis!(ref)
        balance_adiabatically!(ref; Δt = Δt, cycles = cycles)

        # The in-place twin reproduces the graft bit-for-bit (identical explicit scheme + identical IC).
        for (a, b) in zip((dynamics_density(model.dynamics), model.momentum.ρu, model.momentum.ρv,
                           model.momentum.ρw, ρθ_of(model)),
                          (dynamics_density(ref.dynamics), ref.momentum.ρu, ref.momentum.ρv,
                           ref.momentum.ρw, ρθ_of(ref)))
            A, B = Array(interior(a)), Array(interior(b))
            @test maximum(abs, A .- B) <= 1e-8 * max(maximum(abs, B), eps())
        end

        # Balance is non-degenerate: the imbalance develops nonzero ρw (else the match above is vacuous).
        @test maximum(abs, Array(interior(model.momentum.ρw))) > 0
        @test maximum(abs, Array(interior(ref.momentum.ρw)))   > 0

        # ρqᵉ is preserved bit-for-bit — the graft never touches moisture.
        @test Array(interior(model.moisture_density)) == ρqᵉ_analysis
    end
```

test/adiabatic_balance_set.jl
```diff
@@ -0,0 +1,173 @@
...
+        ρ  = dynamics_density(model.dynamics)
+        ρθ = ρθ_of(model)
+        ρ₀  = Array(interior(ρ))
+        ρθ₀ = Array(interior(ρθ))
+
+        set!(model; balance = AdiabaticBalance(Δt = 0.1, cycles = 1))
+
+        @test maximum(abs, Array(interior(ρ))  .- ρ₀)  <= 1e-8 * maximum(abs, ρ₀)
+        @test maximum(abs, Array(interior(ρθ)) .- ρθ₀) <= 1e-8 * maximum(abs, ρθ₀)
+        @test maximum(abs, Array(interior(model.momentum.ρw))) <= 1e-7
```
ewquon: This `ρw ≤ 1e-7` check — like every fixed-point testset in this file — only confirms that a *rest* state is preserved. Rest is a trivial fixed point that any consistent scheme satisfies, **including a silent no-op** (`Δt` collapsing to 0, `cycles = 0`, or production `ρw` never actually shared/stepped). Nothing else here asserts the balance is *non-degenerate*: that an out-of-balance analysis actually develops nonzero `ρw`.

I folded the guard into the revised equivalence test (its IC is a discrete rest state + a small thermal bump, so the excursion develops `ρw`). Verified locally: `ρw` reaches ≈2e-8 (> 0), and `> 0` also catches a `NaN` blow-up:

```julia
# Balance is non-degenerate: the imbalance develops nonzero ρw (else the match above is vacuous).
@test maximum(abs, Array(interior(model.momentum.ρw))) > 0
@test maximum(abs, Array(interior(ref.momentum.ρw)))   > 0
```

These are the only assertions in the file that catch a balance which silently does nothing.

test/adiabatic_balance_set.jl
```diff
@@ -0,0 +1,173 @@
...
+
+    @testset "fixed point: discrete rest state preserved; production clock untouched" begin
+        model = _build_production(default_arch)
+        _set_discrete_rest!(model)
+        ρ  = dynamics_density(model.dynamics)
+        ρθ = ρθ_of(model)
+        ρ₀  = Array(interior(ρ))
+        ρθ₀ = Array(interior(ρθ))
+
+        set!(model; balance = AdiabaticBalance(Δt = 0.1, cycles = 1))
```
ewquon: The `balance = true` success path (`balance_initial_state!(model, ::Bool)` → `AdiabaticBalance()`) is never exercised on a compressible model — only the anelastic *error* test passes `true`. Flipping this rest-state test covers both the `Bool` dispatch and the auto-`Δt` path (`resolve_balance_Δt(::Nothing, …)`) in one go. Rest is a fixed point for any `Δt`, so the tolerances below are unaffected; explicit-`Δt` coverage remains in the native / `with_moisture` / equivalence tests. Verified locally: this testset passes with the flip, full file 49/49.

```suggestion
        set!(model; balance = true)
```

ewquon (approved):
Looks good to me. I suggested some additional tests that came to mind — they passed when I ran them locally.

test/adiabatic_balance_set.jl
glwagner: would increasing the grid size also work?

kaiyuan-cheng:
The mechanics look good to me. One nitpick is that it does not seem to work with anelastic dynamics. `balance_adiabatically!` is designed to be agnostic to the solver.

test/adiabatic_balance_set.jl
glwagner: Done — flipped to `set!(model; balancer = true)` (the kwarg was renamed `balance` → `balancer` in the meantime), so the compressible fixed-point test now exercises the `Bool` dispatch and the auto-`Δt` path. Thanks for catching that the success path was uncovered on a compressible model. Explicit-`Δt` coverage remains in the native / `with_moisture` tests, and there's now also an `AnelasticDynamics` balance test that uses the positional `balance_adiabatically!(model, balancer)` interface.

test/adiabatic_balance_set.jl
glwagner: Yes — increasing the grid size works and is the cleaner fix. The builder's scratch is grid-INDEPENDENT (fixed ~270 KB: NamedTuple/forcing/stepper construction holding references, no grid-scaling arrays), so its ratio to one field falls as the grid grows — measured: **4.0× at 8×8×16, 1.19× at 16×16×32, 0.33× at 32×32×48**. So I bumped this testset to 32×32×48 and assert `@allocated < one_field_bytes` — a clean "less than one field" bound with ~3× headroom that still trips if either aliased tendency set (Gⁿ or U⁰, 6 fields) is reallocated. No tuned multiplier needed.

test/adiabatic_balance_set.jl
glwagner: Added — adapted to the renamed `AdiabaticBalancer` / `balancer` kwarg and the field-reduction + `isapprox` test style. The in-place twin reproduces the hand-stripped explicit graft bit-for-bit, ρqᵉ is preserved, and the non-degeneracy guard is folded in. Thanks — much stronger than the rest-state fixed points.

test/adiabatic_balance_set.jl
glwagner: Agreed — rest is a trivial fixed point any no-op satisfies. Folded the `ρw > 0` non-degeneracy guard into the new graft-equivalence test (IC = discrete rest + small thermal bump, so the excursion develops ρw ≈ 2e-8). That assertion now rejects a silent no-op (Δt→0 / cycles=0 / ρw never shared) and a NaN blow-up.

glwagner:
@kaiyuan-cheng good catch — fixed. `AdiabaticBalancer` now supports `AnelasticDynamics` as well as `CompressibleDynamics`: twin construction dispatches on the dynamics, and the anelastic path reuses the projection-based scheme as-is (no time-discretization or sponge to strip), still sharing field memory and aliasing the SSP stepper's Gⁿ/U⁰. New test balances an anelastic model in place via `balance_adiabatically!(model, balancer)`. The low-level `balance_adiabatically!` stays solver-agnostic as before.

kaiyuan-cheng (approved):

PR #812, opened 2026-06-30
Initialize models through `set!`: `compute_reference_state` + `HydrostaticallyBalancedDensity`

glwagner:
## Summary

Builds out `set!` as a model-*initialization* interface (in the spirit of `NonhydrostaticModel`'s `enforce_mass_conservation`): after setting the basic state, `set!` can apply *global conditions* derived from it. This PR adds two, both for compressible dynamics, plus the machinery they share.

```julia
set!(model; θˡⁱ, qᵗ, u, v,
     ρ = HydrostaticallyBalancedDensity(),   # density (and pressure) in discrete hydrostatic balance
     compute_reference_state = true,          # perturbation-form base state = the analysis mean
     balancer = true)                         # spin ρw into nonhydrostatic balance (#811)
```

### 1. `set!(model; compute_reference_state = true)`
Recompute the dynamics' hydrostatic reference state from the horizontal means of the just-set state, *after* the IC exists — so a limited-area run can derive its perturbation-form base state `p_ref(z)` from the analysis instead of precomputing the mean profile by hand at `CompressibleDynamics` construction.

- **`set_to_mean!(ref::ExnerReferenceState, model)`** extends the existing anelastic-`ReferenceState` recompute to the split-explicit `ExnerReferenceState`: it re-runs the constructor's exact 1-D Exner column integration with the horizontal-mean `θˡⁱ`/`qᵛ`, reusing the existing reference fields (preserving their bottom `ValueBoundaryCondition`s).
- **`reset_reference_state!`** dispatches over both `ReferenceState` and `ExnerReferenceState`; a no-op for dynamics without a reference state.
- Runs before the mass-conservation correction so the pressure projection uses the new reference.

### 2. `set!(model; ρ = HydrostaticallyBalancedDensity(; surface_pressure))`
Set the density — and seed the pressure — in **discrete moist hydrostatic balance** with the just-set `θˡⁱ`/`qᵛ`, so a cold start carries no spurious vertical pressure-gradient force (addresses @kaiyuan-cheng's review point that initialization is incomplete without `pressure_balanced_density`-style balancing).

- It's a **deferred density** (depends on the thermodynamic state): `set!` writes a unit placeholder in phase 1, sets `θ`/momentum, then — after `compute_reference_state`, before mass conservation — runs `set_hydrostatically_balanced_density!`.
- That reuses the reference constructor's per-column Exner integration (`_compute_exner_reference_3d!` — a Newton solve of the discrete balance `(pᵏ − pᵏ⁻¹)/Δz + g(ρᵏ + ρᵏ⁻¹)/2 = 0`) on the model's *actual* `θˡⁱ`/`qᵛ` and the surface pressure, seeds the pressure, then scales the prognostic dry density and `rescale_density_weighted_fields!`s the density-weighted prognostics so the total density matches the balanced column while `θ`, `qˣ`, and the velocities are preserved.

## Validation

`test/reference_states.jl`:
- `set_to_mean!` recompute (and the `set!(; compute_reference_state)` public path, and its no-op without a reference state) reproduce a fresh `ExnerReferenceState` at the new θ — machine precision.
- `HydrostaticallyBalancedDensity` at uniform θ reproduces a fresh `ExnerReferenceState` to machine precision, and the balanced state survives the default mass-conservation correction.

Full `reference_states.jl` plus the `set_atmosphere_model` suite pass; ExplicitImports and Aqua are clean. (Also dropped a dead `rescale_densities` branch from the Exner `set_to_mean!` — it was a rescale-by-1 no-op for compressible dynamics, meaningful only for the anelastic `ReferenceState`.)

## Why it matters

Together these let a nested/limited-area compressible run initialize in one `set!` call, and they're the gate for the NumericalEarth ERA5 downscaling example to drop its hand-rolled hydrostatic-pressure + `reference_θ(z)` construction.


kaiyuan-cheng:
If `set!` is designed to automate model initialization, then there is still one missing element for compressible dynamics. We need to do what `pressure_balanced_density` does to ensure the initial condition is in moist hydrostatic balance.

glwagner:
> If `set!` is designed to automate model initialization

Yes I think we are exploring the concept that `set!` should be interpreted as "set the model state with provided data, plus additional global conditions". For example `Oceananigans.NonhydrostaticModel` includes `enforce_mass_conservation=true/false` which will project out a divergent component. (One could also imagine adding a feature to compute geostrophic balance.. but we don't have that yet).

For `AtmosphereModel` my feeling is that we are building out many more features, similar to enforcing a divergence free condition.

One very very very tiny nitpick is that I like to emphasize that `set!` is really about _mutating_ a model state -- it doesn't only have to be applied "initially" (although this is usually how it is used). The only purpose of emphasizing that is to say that there are _other_ initialization steps that must occur for a simulation to run, which are not performed in `set!`. Therefore one is not required to call `set!` in order to run a simulation. There is a function `initialize!(model)` for that.

> , then there is still one missing element for compressible dynamics. We need to do what `pressure_balanced_density` does to ensure the initial condition is in moist hydrostatic balance.

If this computes density, then perhaps it should be a way to set the density, eg

```julia
set!(model, ρ=HydrostaticBalance(), ...)
```

kaiyuan-cheng (approved):

kaiyuan-cheng:
Changes I made:

- Anelastic reference reset (61202a8) — reset_reference_state! now rescales density-weighted prognostics for ReferenceState, so specific quantities don't drift.

- 3D Exner reference (ee82a96) — set_to_mean!(::ExnerReferenceState) broadcasts the single-column reference across all (i,j).

- Terrain reference (61be74c) — added reset_reference_state!(::TerrainCompressibleModel) and HorizontalMeanProfile to rebuild the 3D terrain-following reference from horizontal-mean θˡⁱ/qᵛ.
- Adiabatic balancing (995e74e) — extracted enforce_mass_conservation!; now also reprojects after balance_adiabatically!.
- Balanced-density pressure (94a09cb) — set_hydrostatically_balanced_density! no longer seeds model pressure (scratch field only; recomputed by update_state!).
- Balanced-density moisture (e91b768) — split rescaling into dry-weighted (momentum, ρθ) vs total-weighted (moisture, microphysics, tracers), so specific humidity matches the set qᵗ.

glwagner:
lets go!

PR #813, opened 2026-06-30
Put examples back in the docs

glwagner:


giordano:
Can we revert #795 and then reland it ensuring documentation actually builds?

glwagner:
superceded by #816

PR #814, opened 2026-06-30
Update baroclinic_wave.jl

glwagner:


giordano:
Shouldn't this be done on top of #813?

giordano:
Uh, no, this specific example is already run.

glwagner:
indeed 😄

glwagner:
the result is: the spurious mode is also present with `UpwindBiased()` as well as `WENO()`.

glwagner:
image

PR #822, opened 2026-07-01
Set Reactant upper bound of 0.2.269

dkytezab:
0.2.270 is breaking some tests, I believe something related to adapting kernel clock args, see https://github.com/NumericalEarth/Breeze.jl/actions/runs/28544512103/job/84626340062#step:14:525


xkykai (approved):

giordano:
Is that https://github.com/EnzymeAD/Reactant.jl/pull/3016? CC @Pangoraw

Pangoraw:
> Is that https://github.com/EnzymeAD/Reactant.jl/pull/3016? CC @Pangoraw

Yeah it seems the check is a bit too eager, i will investigate more. I hope that combined with Oceananigans 0.110.6 (https://github.com/CliMA/Oceananigans.jl/commit/1fa302b3c996ec08db7e8fbc4d41c5e4e91e6a53) it may go back to passing.

giordano:
Ok, this is probably fine as a temporary solution, but it'd be good to fix the integration as soon as possible.

dkytezab:
That perf regression is a bit mysterious — why would bumping Reactant affect a vanilla run?

giordano:
I don't think this is real, the problem is probably that the overall benchmark is still quite fast, normally takes less than half second. A perhaps GC-related jitter is larger than that.

giordano:
Closing as no longer necessary: https://github.com/NumericalEarth/Breeze.jl/pull/823#issuecomment-4860606996. But if there's something to make the changes in Reactant more robust in general, that'd still be worth pursuing upstream.

PR #823, opened 2026-07-01
Try latest Reactant with latest Oceananigans

dkytezab:
Updated Oceananigans and Reactant package versions.


giordano:
So we're good with [Oceananigans@0.110.6](https://github.com/CliMA/Oceananigans.jl/releases/tag/v0.110.6) and we don't need #822?

dkytezab:
Looks like it

PR #829, opened 2026-07-03
[CI] Temporarily sidestep PkgServer

giordano:
There's currently an outage in PkgServer, which backtracked to using an older version of the registry.  By setting   `JULIA_PKG_SERVER=''` we avoid using the PkgServer entirely (which however has other downsides, so shouldn't be a permanent solution).


giordano:
PkgServer has been just fixed, so this is now unnecessary.  Will monitor the situation to see if there are issues again.

PR #830, opened 2026-07-04
Use stable version of Metal

giordano:


giordano:
Right, we need to wait for https://github.com/JuliaGPU/AMDGPU.jl/pull/950 to make into a stable release.

navidcy (approved):

giordano:
Now we need to wait also for https://github.com/JuliaGPU/CUDA.jl/pull/3183 to make it into a release.

PR #831, opened 2026-07-05
Enforce wall impenetrability per acoustic substep (closed-wall mass conservation)

macmccallum:
## Summary

The split-explicit acoustic substepper does not conserve total mass on a **closed (`Bounded`) lateral boundary** — it slowly creates/destroys mass. This PR adds per-substep impenetrability enforcement on the wall-normal momentum perturbations, the **closed-wall companion of #747** (which added the same per-substep boundary enforcement for *open* boundaries).

## Root cause

The slow wall-normal momentum tendency `Gⁿρv` is computed on the **south/west** wall face but not the north/east one: the tendency kernel's `:xyz` launch covers index 1 of a `Face` dimension (south wall, j=1) but never reaches index N+1. On an impermeable wall it should be zero on both. At the south wall it equals the Coriolis force `ρ·f·u`, which the substep loop integrates as `Δτ·Gⁿρv` every substep; the predictor's horizontal divergence then pumps a spurious mass flux through the wall face, growing total mass **∝ Gⁿρv(wall)·(βΔt)²** per RK3 stage (hence the observed Δt²-per-step, south/bottom-heavy drift).

Stage-end recovery + the prognostic no-penetration BC re-zero the *prognostic* wall `ρv`, so the defect is invisible in the recovered state — a closed-wall check on the final field passes even though total mass drifted. The `ExplicitTimeStepping` path conserves because it zeroes the wall momentum via the BC *before* the next divergence is evaluated.

On the DCMIP baroclinic wave (1° lat-lon, ±75° walls) this produced +8.4×10⁻⁴ total-mass growth over 4 days (≈ +4–6 hPa surface-pressure-equivalent by day 30).

## Fix

`enforce_wall_impenetrability!` zeroes the wall-normal momentum perturbations (`ρu′` on Bounded-x faces, `ρv′` on Bounded-y faces) on any side without an active open boundary, **twice per substep** — before the predictor's divergence and after the divergence damping — matching the WRF/MPAS per-substep boundary-enforcement convention. ~55-line change to `src/CompressibleEquations/acoustic_substepping.jl`.

## Regression test (T7)

Added to `test/substepper_rest_state.jl`. A rest state (u=0) cannot expose the bug (the wall-face tendency is `ρ·f·u`), so T7 uses `FPlane` Coriolis + a uniform wall-parallel flow on a `(Periodic, Bounded, Bounded)` grid and asserts total dry mass is conserved to 1e-12 over 30 steps.

| core | relative mass drift | vs 1e-12 bound |
|------|---------------------|----------------|
| without fix (= current `main`) | **3.4×10⁻⁴** | ❌ FAIL |
| with fix (this PR) | **1.5×10⁻¹⁶** | ✅ PASS |

Full `substepper_rest_state` suite (T1–T7) passes in-harness (12/12).

## Notes

- Full diagnosis (magnitude/scaling, Δt² dependence, south-wall localization, stage instrumentation, root cause) added as `validation/baroclinic_wave_mass_conservation_findings.md`, referenced from the fix's code comment.
- Companion to #747.


validation/baroclinic_wave_mass_conservation_findings.md
giordano, line 1: Extra file?

src/CompressibleEquations/acoustic_substepping.jl
```diff
@@ -1323,6 +1323,51 @@ function apply_open_boundary_relaxation!(substepper, model, grid, arch)
...
+    j, k = @index(Global, NTuple)
+    @inbounds ρu′[iᶠ, j, k] = 0
+end
+
+@kernel function _zero_y_wall_face!(ρv′, jᶠ)
+    i, k = @index(Global, NTuple)
+    @inbounds ρv′[i, jᶠ, k] = 0
+end
+
+function enforce_wall_impenetrability!(substepper, model, grid, arch)
```
glwagner, line 1347: isn't there another function for open boundaries? would it be simpler to just have one function (also combined kernels) that enforce Bounded bcs, whether they are open or impenetrable?

validation/baroclinic_wave_mass_conservation_findings.md
macmccallum: yes thanks

src/CompressibleEquations/acoustic_substepping.jl
```diff
@@ -1323,6 +1323,51 @@ function apply_open_boundary_relaxation!(substepper, model, grid, arch)
...
+    bcs_u = model.momentum.ρu.boundary_conditions
+    bcs_v = model.momentum.ρv.boundary_conditions
+    if TX === Bounded
+        is_active_open_bc(bcs_u.west) || launch!(arch, grid, :yz, _zero_x_wall_face!, ρu′, 1)
+        is_active_open_bc(bcs_u.east) || launch!(arch, grid, :yz, _zero_x_wall_face!, ρu′, Nx + 1)
+    end
+    if TY === Bounded
+        is_active_open_bc(bcs_v.south) || launch!(arch, grid, :xz, _zero_y_wall_face!, ρv′, 1)
+        is_active_open_bc(bcs_v.north) || launch!(arch, grid, :xz, _zero_y_wall_face!, ρv′, Ny + 1)
+    end
```
ewquon, line 1361: I think this will have unexpected results if the user attempts to configure a momentum BC with a non-`NormalFlow` BC (ill-posed) or with an outflow `PerturbationAdvection` scheme — `is_active_open_bc` returns false in both cases.

test/substepper_rest_state.jl
```diff
@@ -331,4 +335,108 @@ end
...
+    @test all(ρu_arr[Nx + 1, :, :] .== 0)
+    # Interior x-faces (2:Nx) must still be 1.
+    @test all(ρu_arr[2:Nx, :, :] .== 1)
+
+    # y-wall faces: ρv′[:, 1, :] and ρv′[:, Ny+1, :] must be zero.
+    ρv_arr = Array(interior(ρv′, :, :, :))
+    @test all(ρv_arr[:, 1, :] .== 0)
+    @test all(ρv_arr[:, Ny + 1, :] .== 0)
+    # Interior y-faces (2:Ny) must still be 1.
+    @test all(ρv_arr[:, 2:Ny, :] .== 1)
```
giordano: This isn't terribly important here in the tests, but in general
```suggestion
    @test all(==(0), ρu_arr[1, :, :])
    @test all(==(0), ρu_arr[Nx + 1, :, :])
    # Interior x-faces (2:Nx) must still be 1.
    @test all(==(1), ρu_arr[2:Nx, :, :])

    # y-wall faces: ρv′[:, 1, :] and ρv′[:, Ny+1, :] must be zero.
    ρv_arr = Array(interior(ρv′, :, :, :))
    @test all(==(0), ρv_arr[:, 1, :])
    @test all(==(0), ρv_arr[:, Ny + 1, :])
    # Interior y-faces (2:Ny) must still be 1.
    @test all(==(1), ρv_arr[:, 2:Ny, :])
```
is a bit faster because

1. you don't need to allocate the `A .== c` arrays
1. `all(predicate, array)` shortcircuits at the first element `x` of `array` for which `predicate(x)` is false, while with the present form you have to test all elements (after allocating the full array at the step above).

Using [`@view`](https://docs.julialang.org/en/v1/base/arrays/#Base.@view) would further avoid unnecessary copies when slicing, but I don't want to get too pedantic now 🙂

macmccallum:
@glwagner this is good to merge and close right?

src/CompressibleEquations/acoustic_substepping.jl
glwagner: @macmccallum what did you think about this comment?

PR #832, opened 2026-07-06
AIVA-aware advective timescale: let the time-step wizard float Δt on the horizontal CFL

glwagner:
## Summary

Adaptive implicit vertical advection (#809, #826) lifts the vertical-advective CFL as a *stability* constraint — but the `TimeStepWizard` never learned that, so it still clamped Δt to `|w|/Δz`. In the deep-updraft / fast-hydrometeor regimes AIVA targets, turning AIVA on changed nothing about the Δt the wizard would pick. This PR makes the advective timescale AIVA-aware, so the wizard actually floats Δt on the horizontal CFL.

Measured on an 8×8×32 box with a 12 m/s updraft (Δz = 62.5 m): the wizard at cfl 0.7 picked **3.6 s before, 175 s after — 49× larger Δt**, with AIVA absorbing a vertical CFL of ~33 implicitly.

## The mechanism

`TimeStepWizard` already stores `cell_advection_timescale` as a callable applied to the model (`advective_Δt = wizard.cfl * wizard.cell_advection_timescale(model)`), defaulting to `Advection.cell_advection_timescale`. So this needs no new wizard machinery and no `dims=` kwarg — just a smarter callable in the slot Breeze already overrides.

**Automatic default.** `cell_advection_timescale(model::AtmosphereModel)` now drops the vertical advective term exactly when every vertically-advected prognostic uses AIVA. This mirrors Oceananigans #5642, which made `cell_diffusion_timescale` return `Inf` for vertically-implicit diffusion on the same reasoning: an unconditionally-stable vertical treatment imposes no explicit time-step restriction. It's provably safe here — AIVA's explicit vertical velocity is `wᵉ = w·min(1, cfl/α)`, so the explicit vertical CFL is `min(α, cfl) ≤ cfl` *independent of Δt*; floating Δt on the horizontal CFL can never push the vertical explicit residual past AIVA's own bound.

The predicate is **per-field**, not per-direction: the prognostics share the advecting `w`, so a single explicit scalar (e.g. moisture left on an explicit scheme) re-imposes the vertical CFL and the term stays. That's physically honest — an explicit `ρqᵛ` tendency really is CFL-unstable at large Δt — and it means the automatic benefit kicks in only when AIVA covers all vertical advection (momentum + every scalar).

**Explicit override.** `CellAdvectionTimescale(formulation)` is a callable you drop into the wizard's `cell_advection_timescale` keyword or into `AdvectiveCFL`:

```julia
conjure_time_step_wizard!(sim; cfl=0.7,
    cell_advection_timescale = CellAdvectionTimescale(HorizontalFormulation()))   # force horizontal
AdvectiveCFL(Δt, CellAdvectionTimescale(ThreeDimensionalFormulation()))            # monitor true 3D CFL
```

`ThreeDimensionalFormulation()` / `HorizontalFormulation()` are reused from Oceananigans (the same singletons `min_Δxyz` already dispatches on — the name comes from diffusivity land but reads as "which spatial directions"). The three-dimensional method reproduces Oceananigans' plain timescale exactly. Both are re-exported from Breeze.

Two things fall out for free:
- **Terrain.** The vertical term uses `advecting_vertical_velocity(dynamics, velocities)` (#826), so on TFVD grids the timescale uses the contravariant `w̃` — fixing a latent inaccuracy where the timescale used Cartesian `w` regardless of coordinate.
- **Implementation reuse.** The horizontal-only timescale is Oceananigans' own kernel called with a `ZeroField` in the vertical slot — same reduction, same Flat/topology handling, `|w|/Δz` identically zero. No re-derived spacing operators.

## Trade-off (documented)

Dropping the vertical term lets Δt float to the horizontal CFL, and AIVA's implicit part is only first-order upwind — so at high vertical CFL you trade the stability limit for vertical smearing, governed by the AIVA target `cfl`. That's the intended trade; the escape hatch (`ThreeDimensionalFormulation()` in a monitoring `AdvectiveCFL`) lets you watch how far above 1 the vertical CFL is running.

## Upstream opportunity

The generic "AIVA vertical advection imposes no explicit advective-CFL restriction" belongs in Oceananigans alongside #5642, and would fix the wizard for `NonhydrostaticModel` momentum-AIVA too. Breeze keeps the thin `AtmosphereModel` method for the per-field predicate and the contravariant-`w̃` substitution.

## Tests

New testset in `test/implicit_vertical_advection.jl`: horizontal/3D methods match analytic and the 3D method reproduces Oceananigans' timescale; the automatic default drops the vertical term for an all-AIVA model and keeps it when one scalar is explicit; and the wizard's own hook (`wizard.cell_advection_timescale(model)`) rides the horizontal CFL, leaving a vertical CFL > 1 it no longer clamps.


glwagner:
**Ultimate goal: upstream this.** The AIVA-aware advective timescale is the direct analogue of [#5642](https://github.com/CliMA/Oceananigans.jl/pull/5642), which taught `cell_diffusion_timescale` that vertically-implicit diffusion imposes no explicit time-step restriction (`→ Inf`). The same logic — "adaptive implicit vertical advection contributes no explicit advective-CFL restriction" — belongs in Oceananigans' `cell_advection_timescale`, where it would also fix the `TimeStepWizard` for `NonhydrostaticModel` momentum-AIVA, not just Breeze.

Plan: once this lands and settles, open an Oceananigans PR moving the generic direction-aware / scheme-aware timescale logic upstream (alongside #5642 and the z-Face AIVA work in [#5759](https://github.com/CliMA/Oceananigans.jl/pull/5759)). Breeze would then keep only the thin `AtmosphereModel` method for the pieces that are genuinely Breeze-specific: the per-field predicate over `model.advection` and the contravariant-`w̃` substitution on terrain-following grids.

kaiyuan-cheng (approved):
looks good!

PR #833, opened 2026-07-06
Fix specific-keyed Relaxation forcings under Oceananigans ≥ 0.110.7: materialize SpecificForcing's inner forcing against the specific field

ewquon:
## Summary

Oceananigans v0.110.7 ([CliMA/Oceananigans.jl#5620](https://github.com/CliMA/Oceananigans.jl/pull/5620)) reworked `Relaxation` so that `materialize_forcing(::Relaxation, field, field_name, …)` **captures `field` as the relaxed variable** instead of binding `field_name` (which previously routed through `ContinuousForcing(…, field_dependencies=field_name)`).

`SpecificForcing` materializes its inner forcing with the **ρ-weighted prognostic** and only the **specific name**. That pairing was correct under the old name-based binding, but under the new field-capture binding — and Breeze main's compat is already `Oceananigans = "0.110.8"` — every specific-keyed `Relaxation` silently relaxes the prognostic instead of the specific variable:

- `forcing = (; u = Relaxation(rate, mask, target))` computes `rate·mask·(u_target − ρu)` — an O((ρ−1)·u) ≈ 15 % velocity-scale spurious force;
- a θ fringe computes `(θ_target − ρθ) ≈ −45 K` at standard surface density — a large fictitious cooling wherever the mask is active, on **both** the explicit and split-explicit paths.

In a compressible open-boundary vortex validation case with a 5-cell Davies fringe on u/v/θ, this drives `max|w|` from 5×10⁻⁴-class to >150 m/s within 30 s of model time; the failure bisects exactly to the Oceananigans 0.110.6 → 0.110.7 boundary and disappears with the fringe disabled (trajectories then bit-identical across versions).

## Fix

Materialize the inner forcing against `context.specific_fields[specific_name]` — the same pattern `SubsidenceForcing` already uses (`src/Forcings/subsidence_forcing.jl:127`) — falling back to the prognostic for names with no specific diagnostic (user tracers). The specific fields (`u`, `v`, `w`, `θ`/`e`, moisture) are refreshed in place by `update_state!`, so the captured reference stays live; they share the prognostic's staggered location, so location-only consumers are unaffected. On Oceananigans ≤ 0.110.6 (name-based binding) the change is a verified no-op.

## Verification

- **New regression test** (`test/specific_forcing.jl`): a specific-keyed `Relaxation` must produce `Gρϕ = ρ·rate·(target − ϕ)` on both the Center (θ) and Face (u) paths. On Oceananigans 0.110.8 it **fails before** this change (θ residual 0.77 = rate·(ρ−1)·θ-scale; u residual 5.4×10⁻³) and **passes after**.
- Full `test/specific_forcing.jl` and `test/geostrophic_subsidence_forcings.jl` pass with the change (CPU, Float64).
- Integration: the vortex case above, run on Oceananigans 0.110.8 with this fix, recovers the clean 0.110.6 trajectory (`max|w|` 5.2497 vs 5.2498, matching to 0.002 %; unfixed: 161.7).
- No-op check: the same case on Oceananigans 0.110.6 with this fix reproduces the unfixed 0.110.6 result exactly.

### DIFF
--- a/src/Forcings/specific_forcing.jl
+++ b/src/Forcings/specific_forcing.jl
@@ -86,14 +86,7 @@ function AtmosphereModels.materialize_atmosphere_model_forcing(forcing::Specific
     # name to look up the field they advect or apply at.
     specific_name = startswith(string(name), "ρ") ?
                     Symbol(string(name)[nextind(string(name), 1):end]) : name
-    # Materialize the inner forcing against the *specific* field, not the ρ-weighted
-    # prognostic: forcing types that capture the forced field at materialization
-    # (Oceananigans ≥ v0.110.7 `Relaxation`) must relax the specific variable the
-    # forcing is keyed on, or a fringe relaxation of u computes rate·(u_target − ρu).
-    # The specific diagnostic fields (u, v, w, θ/e, moisture) are refreshed in place
-    # by `update_state!`, so the captured reference stays live; they share the
-    # prognostic's location, so location-only consumers are unaffected. Prognostics
-    # without a specific diagnostic (user tracers) fall back to the prognostic field.
+    # Materialize the inner forcing against the *specific* field, not the ρ-weighted prognostic.
     specific_field = get(context.specific_fields, specific_name, field)
     inner = materialize_atmosphere_model_forcing(forcing.forcing, specific_field, specific_name,
                                                  model_field_names, context)
--- a/src/Forcings/specific_forcing.jl
+++ b/src/Forcings/specific_forcing.jl
@@ -86,7 +86,16 @@ function AtmosphereModels.materialize_atmosphere_model_forcing(forcing::Specific
     # name to look up the field they advect or apply at.
     specific_name = startswith(string(name), "ρ") ?
                     Symbol(string(name)[nextind(string(name), 1):end]) : name
-    inner = materialize_atmosphere_model_forcing(forcing.forcing, field, specific_name,
+    # Materialize the inner forcing against the *specific* field, not the ρ-weighted
+    # prognostic: forcing types that capture the forced field at materialization
+    # (Oceananigans ≥ v0.110.7 `Relaxation`) must relax the specific variable the
+    # forcing is keyed on, or a fringe relaxation of u computes rate·(u_target − ρu).
+    # The specific diagnostic fields (u, v, w, θ/e, moisture) are refreshed in place
+    # by `update_state!`, so the captured reference stays live; they share the
+    # prognostic's location, so location-only consumers are unaffected. Prognostics
+    # without a specific diagnostic (user tracers) fall back to the prognostic field.
+    specific_field = get(context.specific_fields, specific_name, field)
+    inner = materialize_atmosphere_model_forcing(forcing.forcing, specific_field, specific_name,
                                                  model_field_names, context)
     ρ = context.density
     target_location = instantiated_location(field)
--- a/test/specific_forcing.jl
+++ b/test/specific_forcing.jl
@@ -193,7 +193,7 @@ end
     ρᵣ = model.dynamics.reference_state.density
     θ = model.formulation.potential_temperature
     expected = ρᵣ * rate * (θ_target - θ)
-    @test maximum(abs(Gρθ - expected)) ≈ 0 atol = 100 * eps(FT) * maximum(abs(expected))
+    @test Gρθ ≈ expected rtol=100 * eps(FT)
 
     # Face path: u relaxation. ρᵣ varies only in z, so its x-Face interpolation equals
     # the Center value and Gρu = ρᵣ · rate · (u_target − u).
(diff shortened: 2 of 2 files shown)


src/Forcings/specific_forcing.jl
```diff
@@ -86,7 +86,16 @@ function AtmosphereModels.materialize_atmosphere_model_forcing(forcing::Specific
...
                     Symbol(string(name)[nextind(string(name), 1):end]) : name
-    inner = materialize_atmosphere_model_forcing(forcing.forcing, field, specific_name,
+    # Materialize the inner forcing against the *specific* field, not the ρ-weighted
+    # prognostic: forcing types that capture the forced field at materialization
+    # (Oceananigans ≥ v0.110.7 `Relaxation`) must relax the specific variable the
+    # forcing is keyed on, or a fringe relaxation of u computes rate·(u_target − ρu).
+    # The specific diagnostic fields (u, v, w, θ/e, moisture) are refreshed in place
+    # by `update_state!`, so the captured reference stays live; they share the
+    # prognostic's location, so location-only consumers are unaffected. Prognostics
+    # without a specific diagnostic (user tracers) fall back to the prognostic field.
```
glwagner: can we reduce the length of this comment?

glwagner: also I think comments should not reference versions except for extraordinary circumstances

test/specific_forcing.jl
```diff
@@ -169,6 +169,45 @@ end
...
+    model = AtmosphereModel(grid; forcing=(; θ=Relaxation(; rate, target=θ_target)))
+    θ₀ = model.dynamics.reference_state.potential_temperature
+    set!(model; θ=θ₀)
+    update_state!(model)
+
+    Gρθ = interior(model.timestepper.Gⁿ.ρθ) |> Array
+    ρᵣ = interior(model.dynamics.reference_state.density) |> Array
+    θ = interior(model.formulation.potential_temperature) |> Array
+    expected = ρᵣ .* rate .* (θ_target .- θ)
+    @test maximum(abs.(Gρθ .- expected)) < 100 * eps(FT) * maximum(abs.(expected))
```
glwagner: I think this test could use abstract ops and avoid conversion to `Array`, possibly. I think `maximum(abs(field))` will work

test/specific_forcing.jl
```diff
@@ -169,6 +169,45 @@ end
...
+    # Face path: u relaxation. ρᵣ varies only in z, so its x-Face interpolation equals
+    # the Center value and Gρu = ρᵣ · rate · (u_target − u).
+    u_target = FT(10)
+    u₀ = FT(2)
+    model = AtmosphereModel(grid; forcing=(; u=Relaxation(; rate, target=u_target)))
+    set!(model; θ=θ₀, u=u₀)
+    update_state!(model)
+
+    Gρu = interior(model.timestepper.Gⁿ.ρu) |> Array
+    expected_u = ρᵣ .* rate .* (u_target - u₀)
```
glwagner: same comment here re: abstract ops as above

glwagner (approved):

test/specific_forcing.jl
giordano: Why not using `isapprox`?  Same below.

src/Forcings/specific_forcing.jl
ewquon:
```suggestion
    # Materialize the inner forcing against the *specific* field, not the ρ-weighted prognostic.
```

test/specific_forcing.jl
```diff
@@ -169,6 +169,45 @@ end
...
+    model = AtmosphereModel(grid; forcing=(; θ=Relaxation(; rate, target=θ_target)))
+    θ₀ = model.dynamics.reference_state.potential_temperature
+    set!(model; θ=θ₀)
+    update_state!(model)
+
+    Gρθ = model.timestepper.Gⁿ.ρθ
+    ρᵣ = model.dynamics.reference_state.density
+    θ = model.formulation.potential_temperature
+    expected = ρᵣ * rate * (θ_target - θ)
+    @test maximum(abs(Gρθ - expected)) ≈ 0 atol = 100 * eps(FT) * maximum(abs(expected))
```
giordano: I meant
```suggestion
    @test Gρθ ≈ expected rtol=100 * eps(FT)
```
that's a lot simpler than the current expression

test/specific_forcing.jl
```diff
@@ -169,6 +169,45 @@ end
...
+    # the Center value and Gρu = ρᵣ · rate · (u_target − u).
+    u_target = FT(10)
+    u₀ = FT(2)
+    model = AtmosphereModel(grid; forcing=(; u=Relaxation(; rate, target=u_target)))
+    set!(model; θ=θ₀, u=u₀)
+    update_state!(model)
+
+    Gρu = model.timestepper.Gⁿ.ρu
+    expected_u = ρᵣ * rate * (u_target - u₀)
+    @test maximum(abs(Gρu - expected_u)) ≈ 0 atol = 100 * eps(FT) * maximum(abs(expected_u))
```
giordano:
```suggestion
    @test Gρu ≈ expected_u rtol=100 * eps(FT)
```

test/specific_forcing.jl
glwagner: that should work, I think we also have `isapprox` properly defined for `Field`

giordano: Yes, I tested locally before suggesting (and verified that inflating `expected`'s value a little bit would make the test fail) 🙂

ewquon: Thanks for the careful review as always, @giordano!

PR #834, opened 2026-07-06
Reversible leapfrog + digital-filter adiabatic balance (`na_init`)

glwagner:
## Problem

The adiabatic-balance (`na_init`) initialization runs a symmetric ±Δt excursion to spin up the nonhydrostatic state (`ρw`, pressure balance) from an analysis IC. Run with the explicit RK3 acoustic scheme, that excursion is **not time-reversible** — RK3 has amplitude error on the acoustic eigenvalues — so on an energetic (e.g. ERA5-interpolated) IC it drives density negative and crashes the Exner `θ^γ`.

## Fix

Make the excursion **reversible** and filter it properly:

- **New `Leapfrog` time stepper** — three-level, time-symmetric, no acoustic amplitude error. Forward-Euler bootstrap + `turnaround!` for exact reversal; optional weak Robert–Asselin filter for the 2Δt computational mode.
- **`balance_adiabatically!` as a Lynch–Huang DFI** — `span` forward + `span` backward leapfrog steps from the analysis state, combined by a Lanczos-windowed low-pass time average. The reversible trajectory stays bounded while the filter removes the fast acoustic modes, leaving a balanced state with a consistent `ρw`.
- **Dynamics-dependent twin stepper** — leapfrog for `CompressibleDynamics` (acoustic reversibility is the whole point); the native projection scheme for `AnelasticDynamics`/others (no acoustic modes, and the pressure projection must run every step). Solver-agnostic `hasfield`/`restart_bootstrap!` guards keep the low-level routine generic.
- The compressible twin is always `ExplicitTimeStepping` with `Centered` advection — WENO's upwind dissipation is the other irreversibility source. `time_stepping` is therefore ignored for compressible.
- **`acoustic_cfl_safety` 0.85 → 0.4** — the leapfrog excursion is neutrally stable only for acoustic Courant ≤ 1 and, being reversible, provides no damping; a cell above the limit *grows* its oscillation over the span until a step tips it negative. 0.4 gives ~2.5× headroom, covering terrain-metric-compressed columns whose true spacing `minimum_zspacing` understates.
- The twin shares all field storage; **ρθ is rebuilt over the same data with microphysics-stripped surface-flux BCs** (`without_microphysics_bcs`) so the moistureless twin doesn't evaluate the production microphysics scheme on its stripped field set.

No clipping / positivity floors — the fix is purely numerical stability of the excursion.

## Tests

`test/adiabatic_balance_set.jl` rewritten for the inverted (leapfrog + `Centered` + `ExplicitTimeStepping`) twin — **68/68 passing** on CPU. Covers memory-sharing, the leapfrog/Centered assertions, the `time_stepping`-ignored-for-compressible contract, the equivalence graft rebuilt on leapfrog+Centered, and the anelastic twin retaining its native projection stepper (incl. the mass-projection test).

## Validation

Validated end-to-end on the real ERA5 12 km nested IC (via NumericalEarth): the init blow-up went from ρ ≈ −1.3 (RK3) to a stable ρ > 0 with `acoustic_cfl_safety = 0.4` — `iter=0 max|w|=0.10, max|u|=30.75`, no `DomainError`. (Session-34 will add the detailed run note.)


glwagner:
**Validated on the real ERA5→Breeze nested 12 km hindcast IC** (the energetic case this DFI redesign targets — a NumericalEarth.jl nested `AtmosphereModel` initialized from ERA5 pressure-level analysis over Lamont, OK).

**Pre-fix** (leapfrog + `acoustic_cfl_safety = 0.85`): the balancer crashed *mid-span* with an Exner `DomainError` — a cell's θ=ρθ/ρᵈ tipped slightly negative (~−3e-4) during a leapfrog step. Root cause (confirmed): the reversible leapfrog has zero damping and is only neutrally stable for acoustic Courant ≤ 1; a terrain-metric-compressed steep-terrain column sits above the nominal Courant, so its acoustic oscillation *grows* over the span until a step tips it negative. The IC itself was verified physically clean (θ ∈ [295, 302] K, ρᵈ ∈ [0.097, 1.15], zero sub-50 K cells), so it was purely the excursion, not the IC.

**Post-fix** (`acoustic_cfl_safety = 0.4`, ~2.5× headroom below the leapfrog limit): balances cleanly on the same real IC — iter 0 `max|w| = 0.10`, `max|u| = 30.75`, ρ > 0, **no `DomainError`**. Pure stability, no clipping.

LGTM from the downscaling side. (The residual instability we see afterward is a separate *run-phase* issue in the NumericalEarth nesting layer — not the balancer.)

PR #837, opened 2026-07-08
[CI] Always run docs jobs, and internally filter based on files

giordano:
This works around GitHub's limitation that required jobs are _**always**_
required to run, even when they should be skipped.  With this strategy the jobs
are always started, but we only actually run the tests/build the docs if needed,
based on the modified files.

### DIFF
--- a/.github/workflows/Documentation.yml
+++ b/.github/workflows/Documentation.yml
@@ -38,7 +38,6 @@ jobs:
         with:
           filters: |
             files:
-              - ".github/workflows/Documentation.yml"
               - "docs/**"
               - "examples/**"
               - "src/**"
--- a/.github/workflows/Documentation.yml
+++ b/.github/workflows/Documentation.yml
@@ -38,6 +38,7 @@ jobs:
         with:
           filters: |
             files:
+              - ".github/workflows/Documentation.yml"
               - "docs/**"
               - "examples/**"
               - "src/**"
--- a/.github/workflows/Documentation.yml
+++ b/.github/workflows/Documentation.yml
@@ -5,15 +5,7 @@ on:
     branches:
       - main
     tags: "*"
-    paths: &paths
-      - ".github/workflows/Documentation.yml"
-      - "docs/**"
-      - "examples/**"
-      - "src/**"
-      - "Project.toml"
-      - "CHANGELOG.md"
   pull_request:
-    paths: *paths
 
 concurrency:
   # Skip intermediate builds: always.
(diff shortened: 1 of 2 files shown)


giordano:
Sigh, I was too optimistic.  Wouldn't it be great if GitHub just fixed that on their end???!?

giordano:
Ok, there's a small progress in my understanding of the situation:

* skipped jobs are actually fine, they don't block merge, the problem it's the pending ones
* with this PR, "Documentation / build-docs" is skipped (it's not pending anymore), so that's sufficient not to block the PR, per the point above
* the other required job is "Julia min - ubuntu-latest - default - check-bounds yes - pull_request", but that's generated by a matrix, and the matrix generating this job is never run, so that the skipped job is called "CI / Julia ${{ matrix.version }} - ${{ matrix.os }} - ${{ matrix.arch }} - check-bounds ${{ matrix.check-bounds }} - ${{ github.event_name }} (pull_request)", not "Julia min - ubuntu-latest - default - check-bounds yes - pull_request", therefore the required job remains pending

I'm not seeing immediate solutions to this, without further massive complications.  We could at least keep the change for the documentation, that'd at least unblock PRs which don't trigger docs building, like #836.

giordano:
[GitHub Actions docs](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/collaborating-on-repositories-with-code-quality-features/troubleshooting-required-status-checks#handling-skipped-but-required-checks) mention this as a limitation of required jobs, without providing a solution.  An [older version of the docs](https://web.archive.org/web/20230321195500/https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/defining-the-mergeability-of-pull-requests/troubleshooting-required-status-checks) (found in [this discussion](https://github.com/orgs/community/discussions/49124#discussioncomment-5209069)) suggested creating a dummy duplicate workflow, with the desired required name, triggered on the complementary file paths, but that's honestly madness (it's clearly a waste, and error-prone).

The only realistic option I was thinking above was to add yet another job after `test`, requiring `test` but run `if: always()`, which would become the new required job, and would return the exit status of the desired job, if not skipped (if skipped, it'd return 0, that's ok), but this is frankly a bit of a mess (although it _should_ work in theory).

.github/workflows/Documentation.yml
```diff
@@ -30,7 +22,31 @@ env:
...
+      - uses: dorny/paths-filter@7b450fff21473bca461d4b92ce414b9d0420d706 # v4.0.2
+        id: filter
+        with:
+          filters: |
+            files:
+              - ".github/workflows/Documentation.yml"
+              - "docs/**"
+              - "examples/**"
+              - "src/**"
+              - "Project.toml"
```
giordano, line 45: Here we could do something fancy like automatically deciding whether to launch the full docs build based on the touched examples (but can always be overridden by the magic label).  I'll postpone this for another time, but it'd be fun to do

PR #838, opened 2026-07-08
[test] Split Reactant WENO compilation tests into multiple files

giordano:
I noticed that #836 added 15-20 minutes to the test run time just because the Reactant WENO compilation test now takes longer.  Let's see if splitting it into multiple files helps anything.


giordano:
This does save 5-10 minutes (it depends a lot on when the tests are scheduled, re-runs with warm cache should use the historical timings to schedule earlier longest-running tests)

dkytezab (approved):

Issue #825, opened 2026-07-02
Split-explicit open-boundary momentum imbalance: PR #747's per-substep density relaxation kicks the boundary-adjacent momentum (acoustic PGF)


ewquon:
## Summary

`SplitExplicitTimeDiscretization` (`AcousticRungeKutta3`) develops a boundary-adjacent **momentum**
imbalance at open lateral boundaries. PR #747 (`open_boundary_relaxation = α`) fixed the
acoustic-reflection **mass drift** — but, in doing so, **traded it for this momentum kick**: relaxing
the boundary density every substep raises a boundary pressure perturbation whose gradient accelerates
the adjacent momentum (overshoot ∝ α). PR #747's "Known limitation / follow-up" flagged the momentum
imbalance and hypothesized a frozen-tendency (`Gⁿρu`) cause; that hypothesis is **now falsified by
direct test**, and the mechanism is the acoustic pressure-gradient response to the density relaxation.
The imbalance is **still present on Breeze `main`** (untracked), and in an ERA5-driven limited-area
hindcast it appears as a **spurious vertical velocity concentrated at the lateral boundaries** (masked
in the horizontal winds by Davies relaxation). This issue documents the reproducers, results, and the
revised mechanism, and asks for a fix direction.

## Background — the PR #747 known limitation

PR #747 ("Per-substep open-boundary enforcement", tracked by issue #738) added per-substep relaxation of the
perturbation `ρ′, (ρθ)′` toward the prescribed wall value and eliminated the split-explicit **density
drift** (−27% → ≈−0.01% in vortex-transit tests). Paraphrasing its follow-up note:

> It does **not** address a separate, pre-existing **boundary-momentum** imbalance that is intrinsic to
> the acoustic-substepping open-boundary treatment. […] almost entirely in the **momenta** at the
> boundary-adjacent cell (e.g. `ρv` ≈ 30× its own field scale), with `ρ`/`ρθ` only ~1%. The **explicit**
> timestepper shows **no such imbalance** at the same (vertically stable) Δt […] It is **not removable
> by tuning `α`` […] A momentum analog of the enforcement (relaxing the boundary-*face* `(ρu)′/(ρv)′`
> toward 0) […] is a **no-op** […] the spurious signal is at the first *interior* face.
>
> **Suggested follow-up:** the most likely root cause […] is the **frozen stage-entry slow momentum
[...]

PR #839, opened 2026-07-08
Update open-boundary specified zone by value + tendency each acoustic substep

ewquon:
Finalizes the MPAS/WRF "specified zone" boundary drive introduced in NumericalEarth/Breeze.jl#825, for open<br>lateral boundaries under acoustic substepping. When a momentum `NormalFlowBoundaryCondition`<br>carries the `SubstepBoundaryUpdate` scheme, that side's outermost interior cells become a<br>*specified zone* that is (1) excluded from the coupled acoustic update, (2) closed for<br>`(ρw)′` by a per-substep zero-gradient copy from the nearest interior column, and<br>(3) updated each acoustic substep by its boundary value + time-tendency. This supplies the<br>substepped open boundary with time-accurate data *inside* the substep loop, collapsing the<br>acoustic reflection at open boundaries.

### Interface — fields only (updated per review)

Specified-zone tendencies are supplied through the single `boundary_tendencies(model)`<br>property (`∂ₜρu, ∂ₜρv, ∂ₜρᵈ, ∂ₜρθ, ∂ₜρqᵛ` — bundled from five separate substepper fields<br>per review), filled in place by a driver each outer step and persisting until overwritten.<br>The earlier callable-source slots have been removed and the scheme is now a fieldless<br>marker, `SubstepBoundaryUpdate` (it names the behavior — refresh the boundary each acoustic<br>substep — rather than the internals; the internal helpers speak "specified zone"<br>throughout).

**Migration:** `BoundaryTendencyMarch(ρu_tendency = f, …)` → `SubstepBoundaryUpdate()`<br>followed by `set!(boundary_tendencies(model).ρu, …)`.

### Terrain support

The scheme now runs on `TerrainCompressibleDynamics`. The terrain horizontal<br>pressure-gradient corrections (the slow-tendency slope projection and the<br>linearized-perturbation correction) are the only non-column-local stencils; each is gated<br>with a one-sided, specified-face-dropping interpolation so a re-imposed specified cell's<br>pressure never projects into an interior column through the slope. The slow `ρw̃`<br>projection substitutes the boundary tendency at specified faces. All gating collapses to<br>the original interpolation on flat grids and on terrain without a specified zone<br>(`::Nothing`/default methods — bit-identical, zero added cost). The `AcousticSubstepper`<br>≥3-cell requirement is now enforced per specified direction, so a Flat-y terrain grid whose<br>specified zone is only west/east constructs.

### Verification

* `test/substep_boundary_update.jl`: all green (551/551), incl. new terrain rest-hold,<br>composition, and scalar-leak-discriminator testsets.
* `test/acoustic_substepping.jl` (364/364) and the `terrain_following_*` split files: no<br>new failures vs baseline (only the 4 pre-existing zero-terrain-vs-height tolerance<br>failures, which fail on `main` too). `quality_assurance` (Aqua/ExplicitImports) clean.
* Flat 3-arm gauntlet (gpu-dev, T4, 18 jobs): all finite, no blowup. The scheme arms<br>(mfull/phys) collapse the boundary-reflection max|w| to O(1e-4) at the stationary vortex<br>and O(1e-7) as the vortex transits, vs the schemeless base arm's O(1e-3) — the scheme<br>behaves as designed. The base (schemeless) arm is **bit-identical** to pre-A1: its full<br>trajectory logs (every snapshot, all 6 scenarios) are character-identical between the<br>pre-A1 and current code on gpu-dev, confirming the `::Nothing` specializations compile the<br>specified-zone branches away with zero effect on schemeless runs.
* Terrain gauntlet (gpu-dev, T4, Float64): the terrain specified-zone rest-hold holds at<br>machine level — max|w̃| = 4.58e-12 (with the specified-zone update) vs 4.16e-12 (without),<br>mass drift 1.1e-16 — confirming the specified-zone terrain kernels are correct and stable<br>on GPU.

Coexistence with the terrain-kinematic lateral `ρw` BC (separate PR) is verified by a<br>follow-up smoke once both land.

### Follow-ups (out of scope for this PR)

1. **Unify the open-boundary acoustic relaxation under** `scheme=`**.** The per-substep<br>`ρ′`/`(ρθ)′` relaxation (NumericalEarth/Breeze.jl#738) is currently a *global* default (`open_boundary_relaxation`,<br>default 0.5, applied on every active-open side without the scheme, and not disable-able —<br>`0` is rejected). Making it a per-side scheme (e.g. `SubstepRelaxation(α)`), mutually<br>exclusive with `SubstepBoundaryUpdate`, would allow per-side α and explicit opt-out and<br>replace the `!specified_sides` supersession gate with plain dispatch. (Note it relaxes<br>mass/thermo, not momentum.) A test asserting the scheme skips the relaxation on specified<br>sides would go with it.
2. **Specified-zone-aware slow (Davies) relaxation mask.** Under substepping the<br>specified-zone cells are hard-set, so a slow lateral `u`/`v` nudge in those cells is<br>overwritten by the re-imposition (redundant); under explicit stepping (no specified-zone<br>update) that nudge is the boundary's only interior driver. A specified-zone-aware mask —<br>excluding specified cells when substepping, covering them otherwise — mirrors MPAS's<br>disjoint *specified* vs *relaxation* zones. (Forcing / NumericalEarth-layer change.)
3. **Strip storage for the boundary-tendency fields.** They are currently full 3-D fields<br>read only in the \~1-cell specified zone (≈ 30× per-field over-allocation; the 5 fields are<br>≈ 11% of a 120×120×50 model, \~39 MB). A per-side strip layout would cut that to \~1 MB —<br>worth it at large/memory-bound domains. (Prototype branch: `eq/boundary-tendency-strip-storage`.)

🤖 Generated with [Claude Code](<https://claude.com/claude-code>)


glwagner:
So the idea here is to replace a prescription of just boundary values with the prescription of an entire near-boundary zone? Is that correct? Does this improve stability somehow?

ewquon:
No, it's not replacing the prescription zone, just changing intermediate values during the substepping. Eliminates transients that were previously mitigated by boundary acoustic relaxation + Davies relaxation zone. 

Still in draft because a few more improvements are needed to generalize this.

glwagner:
How do you compute the tendencies?

glwagner:
It seems to have a complexity downside and also introduces more memory movement which may have a significant cost (since substepping accounts for roughly 50% of the cost as-is, and is memory bound). 

> tendency sources are (x, y, z, t) callables

I don't understand this statement... but, we should never use continuous form for built-in features (it is nice to get users going quickly in scripts, but not for source code / production implementations)

ewquon:
> How do you compute the tendencies?

These are finite differences from the boundary data, evaluated at every substep.

> It seems to have a complexity downside and also introduces more memory movement which may have a significant cost (since substepping accounts for roughly 50% of the cost as-is, and is memory bound).

The scheme is opt-in. For my 2.5D toy 128x128 toy problem on a T4, with all 4 boundaries marched, the increase in walltime is ~2% (Claude's estimate).

> we should never use continuous form for built-in features (it is nice to get users going quickly in scripts, but not for source code / production implementations)

The callables are just a convenience layer that is never evaluated inside the substep kernels, which I can drop.

src/CompressibleEquations/acoustic_substepping.jl
```diff
@@ -119,6 +119,16 @@ struct AcousticSubstepper{N, FT, D, AD, US, CF, MP, TAV, GT, TS}
...
+
+    # BoundaryTendencyMarch (#825): per-substep march of the specified-zone
+    # perturbations by their boundary time-tendencies. `Nothing` unless a
+    # momentum BC carries the scheme; filled once per outer step (from the
+    # scheme's sources, or in place by a driver via `boundary_tendency_fields`).
+    boundary_momentum_tendency_u :: BTU
+    boundary_momentum_tendency_v :: BTV
+    boundary_density_tendency :: BTR
+    boundary_density_potential_temperature_tendency :: BTX
+    boundary_moisture_tendency :: BTQ
```
glwagner: I suggest bundling these into a single `boundary_tendencies` property

ewquon: Done in `a451a2f1` — bundled the five separate fields into a single `boundary_tendencies` property (a `NamedTuple` `(ρu, ρv, ρᵈ, ρθ, ρqᵛ)` when marched, `nothing` otherwise), which collapses the five type params to one and simplifies Adapt/constructor with them. I also renamed the public accessor `boundary_tendency_fields` → `boundary_tendencies` to match. `boundary_tendency_march.jl` stays 551/551.

PR #840, opened 2026-07-08
[CI] Run fake test job when non-relevant files are touched

giordano:
Follow up to #837, let's see if this does the trick for the test jobs.


giordano:
skywalker-star-wars

PR #843, opened 2026-07-10
Strip diabatic surface flux BCs from the adiabatic-balance twin

### LINKED ISSUE #842: balance_adiabatically! fails GPU compilation: adiabatic twin keeps a diabatic surface flux BC
## Impact

`balance_adiabatically!(model, AdiabaticBalancer())` throws `InvalidIRError` at GPU
compile time whenever the model has a moisture-coupled surface flux boundary condition and a
saturation-adjustment 1-moment microphysics (`OneMomentCloudMicrophysics` with
`MixedPhaseEquilibrium`). The adiabatic DFI spin-up is unusable on GPU for this common
configuration, so cold-started analyses (ERA5, GFS) cannot be balanced and must run from an
unbalanced state, which rings up and blows up.

## Root cause

`assemble_adiabatic_twin` (`src/AtmosphereModels/adiabatic_balance.jl`) strips `microphysics`,
`closure`, `forcing`, and the sponge from the twin, but reuses the production prognostic fields
(`momentum`, `moisture_density`, `ρθ`) verbatim. In Oceananigans a `Field` carries its own
boundary conditions, so the twin's `ρθ` still holds the production model's bottom
`EnergyFluxBoundaryConditionFunction`. That BC captured its own `microphysics` reference at build
time, independent of `model.microphysics`, so nulling `model.microphysics` does not disarm it.

During the twin's adiabatic excursion, `update_state!` / `time_step!` evaluates that BC:

```
getbc(::BottomEnergyFluxBC)                      src/BoundaryConditions/thermodynamic_variable_bcs.jl:87
  -> 𝒬_to_Jᶿ                                     src/BoundaryConditions/thermodynamic_variable_bcs.jl:78
     -> grid_moisture_fractions(ef.microphysics=MP1M, fields)
```

The MP1M `grid_moisture_fractions` override reads microphysics fields off its `μ` argument:

```julia
# ext/BreezeCloudMicrophysicsExt/one_moment_microphysics.jl:547-549
@inline function AM.grid_moisture_fractions(i, j, k, grid, bμp::MP1M, ρ, qᵉ, μ)
    qᶜˡ = @inbounds μ.qᶜˡ[i, j, k]     # line 549
    qᶜⁱ = @inbounds μ.qᶜⁱ[i, j, k]
    qʳ  = @inbounds μ.ρqʳ[i, j, k] / ρ
    qˢ  = @inbounds μ.ρqˢ[i, j, k] / ρ
    ...
```

But `𝒬_to_Jᶿ` passes the model `fields`, and the stripped twin's fields no longer contain
`qᶜˡ`, `qᶜⁱ`, `ρqʳ`, `ρqˢ` (the twin drops precip prognostics and remaps `ρqᵉ -> ρqᵛ`). The
`μ.qᶜˡ` access cannot resolve statically, so it lowers to a runtime `getproperty` /
`jl_f_getfield`, which is unsupported in GPU kernels.

There are two problems, both fixed by the same change:

1. Compilation: the BC needs microphysics fields the twin deliberately dropped.
2. Physics: an adiabatic `na_init` excursion must have zero surface heat flux, so evaluating an
   energy-flux BC at all is inconsistent with "adiabatic," even if it compiled.

## Error

```
InvalidIRError: compiling ... gpu__compute_z_bcs!(...) resulted in invalid LLVM IR
Reason: unsupported call to an unknown function (call to jl_f_getfield)
Stacktrace:
 [1] getproperty                 ./Base_compiler.jl:54
 [2] grid_moisture_fractions     ext/BreezeCloudMicrophysicsExt/one_moment_microphysics.jl:549
 [3] 𝒬_to_Jᶿ                     src/BoundaryConditions/thermodynamic_variable_bcs.jl:78
 [4] getbc                       src/BoundaryConditions/thermodynamic_variable_bcs.jl:87
 [6] compute_z_bottom_bc!        Oceananigans .../compute_flux_bcs.jl:128
 [9] gpu__compute_z_bcs!
   ...
[36] balance_adiabatically!      src/AtmosphereModels/adiabatic_balance.jl:49
```

## Reproduction

A GPU `AtmosphereModel` on a bounded terrain-following `LatitudeLongitudeGrid` with
`OneMomentCloudMicrophysics(cloud_formation = SaturationAdjustment(MixedPhaseEquilibrium()))` and a
bottom `EnergyFluxBoundaryCondition` on the thermodynamic variable, then
`balance_adiabatically!(model, AdiabaticBalancer())`.

Isolated confirmation via the twin (GPU):

```
child.microphysics  = BulkMicrophysics
child  ρθ bottom BC = EnergyFluxBoundaryConditionFunction

twin  = adiabatic_balance_twin(child, AdiabaticBalancer())
twin.microphysics   = nothing                              # stripped, as intended
twin  ρθ bottom BC  = EnergyFluxBoundaryConditionFunction  # NOT stripped -> the bug
time_step!(twin, Δt) -> InvalidIRError (above)
```

## Suggested fix

In `assemble_adiabatic_twin`, strip diabatic surface flux boundary conditions from the twin's
thermodynamic (and moisture) prognostic fields, the same way microphysics / closure / forcing are
already stripped. Concretely, rebuild the twin's `ρθ` (and moisture) fields to share the production
data but with `NoFlux` top/bottom BCs, so the adiabatic excursion sees no diabatic surface forcing.
This both fixes the GPU compile and makes the excursion correctly adiabatic.

## Environment

- Breeze `main` @ `c5232501` (tree `b6eeef20`)
- Oceananigans `main`
- CUDA on an NVIDIA T4, Julia 1.12.6
- Microphysics and the surface energy-flux BC are supplied by NumericalEarth's nested-atmosphere
  child; the failure is entirely inside Breeze's balancer + energy-flux BC.

ewquon:
Fixes NumericalEarth/Breeze.jl#842.

## Problem

The adiabatic initialization twin built by `assemble_adiabatic_twin` strips microphysics, closure, and forcing but reuses the production thermodynamic field, so its **diabatic surface flux BC survives on the twin**. That is non-adiabatic, and for a moisture-coupled `EnergyFluxBoundaryCondition` with saturation-adjustment 1-moment microphysics it also fails to compile on GPU: the physics-free twin drops the microphysics fields that `grid_moisture_fractions` reads, so the flux BC lowers to a runtime `getfield` (`InvalidIRError: jl_f_getfield`).

## Fix

Rebuild the twin's thermodynamic density sharing the production data (so the balanced state still lands in `model`) but with no-flux vertical BCs:

* `adiabatic_thermodynamic_bcs` (BoundaryConditions): replace sensible-heat / energy / θ surface flux BCs with no-flux, leaving dynamical (open/impenetrable) BCs untouched. Mirrors the existing `energy_to_theta_bcs` transform.
* `with_thermodynamic_density` (formulation interface): swap a thermodynamic field into a formulation without reallocating diagnostics.

## Verification

* New testset in `test/balance_adiabatically.jl`: the production model carries the `EnergyFluxBoundaryConditionFunction`, the twin does not, the twin shares the production data, and `balance_adiabatically!` runs to completion from a rest state. `test/balance_adiabatically.jl` passes 25/25 on CPU.
* Confirmed on GPU (T4) that stripping the twin's diabatic BC clears the `InvalidIRError` and lets `balance_adiabatically!(model, AdiabaticBalancer())` build, run its excursion, and step.


src/AtmosphereModels/adiabatic_balance.jl
```diff
@@ -210,9 +210,20 @@ adiabatic_balance_twin(model::AtmosphereModel, balancer::AdiabaticBalancer = Adi
...
     grid        = model.grid
     arch        = model.architecture
-    formulation = model.formulation
     constants   = model.thermodynamic_constants
 
+    # Strip diabatic surface fluxes from the twin's thermodynamic field. The excursion is adiabatic,
+    # so surface heating must be absent; and a moisture-coupled energy-flux BC additionally fails to
+    # compile on GPU against the physics-free twin, whose field set drops the microphysics fields the
+    # BC reads (issue #842). Rebuild the thermodynamic density sharing the production data (so the
+    # balanced state still lands in `model`) but with no-flux vertical BCs.
```
ewquon:
```suggestion
    # Strip diabatic surface fluxes from the twin's thermodynamic field.
    # The excursion is adiabatic—i.e., pure dynamics—so surface heating must be absent.
```

kaiyuan-cheng (approved):

glwagner:
Should we just strip all bcs? I don't think it's actually possible (not simply anyways) to figure out whether a bc is diabatic or adiabatic, right?

src/BoundaryConditions/thermodynamic_variable_bcs.jl
```diff
@@ -285,15 +292,22 @@ theta_to_energy_bc(bc::EnergyFluxBCType) = BoundaryCondition(Flux(), bc.conditio
...
-function theta_to_energy_bcs(fbcs::FieldBoundaryConditions)
-    return FieldBoundaryConditions(; west     = theta_to_energy_bc(fbcs.west),
-                                     east     = theta_to_energy_bc(fbcs.east),
-                                     south    = theta_to_energy_bc(fbcs.south),
-                                     north    = theta_to_energy_bc(fbcs.north),
-                                     bottom   = theta_to_energy_bc(fbcs.bottom),
-                                     top      = theta_to_energy_bc(fbcs.top),
-                                     immersed = theta_to_energy_bc(fbcs.immersed))
-end
+theta_to_energy_bcs(fbcs::FieldBoundaryConditions) = map_field_boundary_conditions(theta_to_energy_bc, fbcs)
```
glwagner, line 295: Is this needed? I think that the `AtmosphereModel` constructor will correctly handle either boundary condition formulation. We don't need to manually fiddle with the bcs for the twin.

test/balance_adiabatically.jl
```diff
@@ -22,7 +26,8 @@ const CPD_AI = 1005.0
 
 # Init-mode model: no upper sponge (the one irreversible substep ingredient),
 # no microphysics. Otherwise mirrors the production split-explicit config.
-function _build_adiabatic_model(arch; Nx = 8, Ny = 8, Nz = 32, Lz = 10e3, Lh = 100e3)
+function _build_adiabatic_model(arch; Nx = 8, Ny = 8, Nz = 32, Lz = 10e3, Lh = 100e3,
```
glwagner, line 30: spurious `_`

src/AtmosphereModels/adiabatic_balance.jl
```diff
@@ -207,12 +207,29 @@ not touch the production clock.
 adiabatic_balance_twin(model::AtmosphereModel, balancer::AdiabaticBalancer = AdiabaticBalancer()) =
     assemble_adiabatic_twin(model, adiabatic_twin_dynamics(model.dynamics, balancer.time_stepping))
 
+# Rewrap a prognostic scalar `field` with its surface fluxes stripped to no-flux (see
+# [`adiabatic_scalar_bcs`](@ref)), sharing the production `data`, indices, and operand so no memory
+# is reallocated and the balanced state still lands in `model`.
+adiabatic_field(field) =
+    Oceananigans.Field(Oceananigans.instantiated_location(field), field.grid, field.data,
+                       adiabatic_scalar_bcs(field.boundary_conditions),
```
glwagner: should we just strip the bottom boundary conditions? Is that what we need?

src/AtmosphereModels/adiabatic_balance.jl
```diff
@@ -207,12 +207,29 @@ not touch the production clock.
 adiabatic_balance_twin(model::AtmosphereModel, balancer::AdiabaticBalancer = AdiabaticBalancer()) =
     assemble_adiabatic_twin(model, adiabatic_twin_dynamics(model.dynamics, balancer.time_stepping))
 
+# Rewrap a prognostic scalar `field` with its surface fluxes stripped to no-flux (see
+# [`adiabatic_scalar_bcs`](@ref)), sharing the production `data`, indices, and operand so no memory
+# is reallocated and the balanced state still lands in `model`.
+adiabatic_field(field) =
```
glwagner: `twin_field` might be clearer here

kaiyuan-cheng:
> Should we just strip all bcs? I don't think it's actually possible (not simply anyways) to figure out whether a bc is diabatic or adiabatic, right?

I thought that momentum BC should be fine but I agree that we should just strip all bcs.

src/AtmosphereModels/adiabatic_balance.jl
kaiyuan-cheng: In practice, bottom bcs should be enough, right? I don't expect we will apply lateral forcings.

glwagner: I think we can start with that for simplicity and refactor if we need something more general

glwagner (approved):

PR #844, opened 2026-07-10
Refine baroclinic wave example: 64-level grid, WENO5, richer diagnostics

macmccallum:
Refines the DCMIP2016 baroclinic wave example (`examples/baroclinic_wave.jl`) so it resolves the developing Rossby wave more faithfully.

## Changes

- **Grid**: 64 vertical levels on a plain `z = (0, H)` column. Adds an `exponential_scale` helper that bisects the `ExponentialDiscretization` e-folding `scale` to hit an exact 150 m surface cell (kept commented for now), replacing the previous ad-hoc analytic stretched `z_faces`.
- **Dynamics**: removes the horizontal biharmonic hyperviscosity closure and switches advection to `WENO(order=5)`.
- **Output & visualization**: saves at four levels — surface, ~2.9 km, the 10.5 km jet level, and the 29.5 km model top — and renders a two-row, six-panel sphere animation (surface θ/ζ/w plus jet-level ζ and model-top w/ζ). Output renamed to `baroclinic_wave_f32_w5.mp4`.

## Notes

- Scope is limited to the single example file; no source or `Project.toml` changes.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #846, opened 2026-07-15
Moist baroclinic wave on the sphere (DCMIP2016)

macmccallum:
Adds a moist DCMIP2016 baroclinic wave example — the moist companion to `examples/baroclinic_wave.jl` — plus a moist power-method validation script. Kessler warm-rain microphysics on the 1° sphere, `CompressibleDynamics` with acoustic substepping.

## What's here

- **`examples/moist_baroclinic_wave.jl`** — DCMIP2016 §1.1 moist baroclinic wave: balanced jet in virtual temperature + background specific humidity, Kessler microphysics (prognostic `qᵛ`/`qᶜˡ`/`qʳ`), bounds-preserving tracer WENO, `cos⁴φ` biharmonic hyperviscosity.
- **`validation/power_method_moist_baroclinic.jl`** — moist counterpart of the dry power-method study (Park, Skamarock & Klemp 2013 procedure on the moist jet).

## Getting it to a full 30-day run (no changes to the core dynamics)

Four independent obstacles, each diagnosed and fixed:

1. **Supersaturated IC.** An early height-coordinate `qᵛ` profile was ~40% supersaturated; Kessler condensed it instantly → latent-heat shock. The **DCMIP2016 §1.1 pressure-coordinate profile** is correctly subsaturated.
2. **First-step adjustment transient** (full res) → **gentle Δt start** (`Δt₀=1 min`, wizard ramps to 12 min).
3. **Day-8 crash = a baroclinic instability of the computational kind (BICK).** From the 2Δx vorticity spectrum: moist frontogenesis excites a grid-scale (2Δx) mode near the jet, growing ~4× faster than the physical wave and decoupled from it. Cured with a **`cos⁴φ`-scaled biharmonic hyperviscosity** (`ν₄=1e14·cos⁴φ`, momentum only) — the scale-selective filter WRF provides via 6th-order diffusion. (Breeze's core is flux-form like WRF-ARW, so this is the general computational mode, not the vector-invariant Hollingsworth instability.)
4. **Day-14.5 wall = Float32 precision.** In the intense mature phase (~80 m/s), the `θ→T` inversion trips on sharp fronts in F32. **Float64 completes the full 30-day life cycle** (verified: 121 snapshots to day 30, all finite/physical). F32 is enough through the DCMIP2016 day-10–15 evaluation window; F64 (~2× cost) gives the complete run.

## Validation

- **Growth rate robust to moisture:** moist power method converges to **σ ≈ 0.765 day⁻¹** (wavenumber-9), within ~2% of the dry DCMIP2016 rate — Park et al.'s prediction.
- **Full life cycle:** F32 clean through cyclogenesis to ~day 14.5; **F64 through the complete 30 days.**

### DIFF
--- a/examples/moist_baroclinic_wave.jl
+++ b/examples/moist_baroclinic_wave.jl
@@ -315,15 +315,20 @@ set!(model, θ=potential_temperature, u=zonal_velocity, ρ=density, qᵛ=vapor_m
 #
 # Substepping eliminates the acoustic CFL constraint on the outer Δt; the advective
 # CFL remains. A time-step wizard targets advective CFL ≈ 0.7 against the polar
-# `Δx_min ≈ 28.8 km`, capped at Δt = 12 min. This cap is the same aspirational
-# target as the dry case; moist convection may require a smaller step, so reduce
-# `max_Δt` if the run destabilizes.
+# `Δx_min ≈ 28.8 km`, capped at Δt = 12 min (the same target as the dry case).
+#
+# We start from a gentle `Δt = 1 minute` and let the wizard ramp it (`max_change =
+# 1.08` per adjustment) up to the 12 min cap. The moist balanced state launches a
+# brief adjustment transient in the first steps; at full resolution the sharp
+# near-surface layer amplifies it, so taking the first ~hour of simulation at a
+# small step lets it damp out before the step grows. Jumping straight to 12 min on
+# step 1 instead re-triggers the transient into a vertical-velocity shock.
 
-Δt = 12minutes
+Δt = 1minute
 stop_time = 30days
 
 simulation = Simulation(model; Δt, stop_time)
-conjure_time_step_wizard!(simulation; cfl=0.7, max_Δt=12minutes)
+conjure_time_step_wizard!(simulation; cfl=0.7, max_Δt=12minutes, max_change=1.08)
 Oceananigans.Diagnostics.erroring_NaNChecker!(simulation)
 
 # ## Progress callback
--- a/examples/moist_baroclinic_wave.jl
+++ b/examples/moist_baroclinic_wave.jl
@@ -36,14 +36,14 @@
 #     time-step wizard below. A full 30-day sphere run is a workstation-GPU-scale
 #     computation, not a quick doctest.
 #
-# !!! warning "Known limitation (day ~8)"
-#     With the settings below the wave develops realistically — jet, growing
-#     Rossby wave, condensation, and rainfall — through roughly day 8, but the run
-#     currently **crashes during vigorous moist cyclogenesis** around day 8: sharp
-#     F32 frontal features trip the thermodynamic ``θ→T`` inversion. Reaching the
-#     full 30 days needs added frontal dissipation or a more robust inversion, which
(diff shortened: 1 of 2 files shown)


macmccallum:
WIP (Not ready for general review). I believe the current instability may be explained by Arakawa & Moorthi 1987. In which case this is related to the dry  case in #844.

macmccallum:
https://github.com/user-attachments/assets/67582f4a-71aa-4668-b27b-834e0391e0b0

macmccallum:
I am currently hunting down the reason for F32 instability. The Sim above was done with F64.

macmccallum:
## Follow-up: full-30-day run in Float32 (tracked)

**Status (this PR):** F32 runs cleanly through cyclogenesis to **~day 14.5** (past the DCMIP2016 day-10–15 evaluation window); **Float64 completes the full 30-day life cycle**. Documented in the example's warning box.

**Why F32 walls at ~day 14.5:** a *distributed* Float32 precision limit in the intense mature phase (~80 m/s), diagnosed by instrumenting the θ→T inversion inputs at full resolution:
- Advective transport leaves the prognostic condensate `ρqᶜˡ`/`ρqʳ` slightly negative at **F32 roundoff scale** (~1e-4 relative), growing with the field (~1e-11 early → ~1e-6 at day 14.5).
- Amplified by the low upper-atmosphere density into a negative mass *fraction*, this feeds a negative latent-heat term and tips the Exner base `(ρRᵐT/pˢᵗ)^κ` negative → NaN (GPU) / `DomainError` (CPU).
- F64 keeps every roundoff term ~9 orders smaller, below the tripping threshold → day 30.

**Cheap fixes tried, all rejected (don't re-tread):**
- *Diagnostic moisture clip* in `grid_moisture_fractions` → **worse** (crashed day 8.65): zeroing water for the EOS while the density keeps it creates a mass/EOS inconsistency that destabilizes the acoustic substepper.
- *Mixed-precision (F64) θ→T inversion* only → **didn't clear** day 14.5: the limit isn't localized to the inversion (it's where the crash surfaces, not where it originates).
- *Different per-tracer advection* (`UpwindBiased(1)` / `WENO(3)` on moisture while `ρθ` stays `WENO(5)`) → **blocked**: a heterogeneous scalar-advection tuple explodes acoustic-substep kernel-specialization compile time (>25 min vs ~6 min uniform).

**The real lever (follow-up):** a *uniform*, **positivity-preserving / conservative moist transport** (e.g. an FCT-style limiter or a conservative hole-filler applied to `ρqᵛ/ρqᶜˡ/ρqʳ` before the EOS/microphysics read the state) — stopping the negative moisture at the source rather than masking it, and consistently (density + pressure agree). That's a source-level scheme addition, out of scope for this example PR; F64 is the interim path for the complete run.

glwagner:
shouldn't we look for an f32 specific issue in cancellation error? A similar problem for the TC world case eventually lead to a reformulation of the buoyancy computation (which only applies to anelastic dynamics, but anyways...)

One test could be to run with simpler warm-phase saturation adjustment. Does the f32/f64 failure have to do with the microphysics scheme?

PR #851, opened 2026-07-16
[CI] Give 30-minute timeout for instantiating test environment

giordano:
Also, temporarily pin Reactant to previous version because the Windows library can't be loaded: https://github.com/EnzymeAD/ReactantBuilder/pull/198#issuecomment-4990798934


giordano:
I'm hopeful https://github.com/EnzymeAD/ReactantBuilder/pull/200 will resolve the problems with loading the Windows library (after we solve Bazel shenanigans....), I'll unpin Reactant once that's out.

Issue #511, opened 2026-02-22
Possible errors detected by JET


giordano:
With
```julia
using JET, Breeze
using LinearAlgebra, Oceananigans, MPI, StaticArraysCore


ignored_modules = (Base, LinearAlgebra, Oceananigans, MPI, StaticArraysCore)

JET.test_package(Breeze; ignored_modules)
```
I get
```
JET-test failed at /Users/mose/.julia/packages/JET/lX05m/src/JETBase.jl:1375
  Expression: (JET.report_package)(Breeze; toplevel_logger = nothing, ignored_modules = (Base, LinearAlgebra, Oceananigans, MPI, StaticArraysCore))
  ═════ 2 possible errors found ═════
  ┌  @ Breeze.PotentialTemperatureFormulations /Users/mose/.julia/dev/Breeze/src/PotentialTemperatureFormulations/potential_temperature_tendency.jl:91
  │ no matching method found `exner_function(::Breeze.Thermodynamics.StaticEnergyState, ::Any)` (1/2 union split): Π = Breeze.PotentialTemperatureFormulations.exner_function(𝒰, constants::Any)
  └────────────────────
  ┌ time_step!(model::Oceananigans.AbstractModel{<:SSPRungeKutta3}, Δt::Any; callbacks::Any) @ Breeze.TimeSteppers /Users/mose/.julia/dev/Breeze/src/TimeSteppers/ssp_runge_kutta_3.jl:205
[...]

PR #860, opened 2026-07-18
Resolve some issues detected by JETLS

giordano:


src/MoistAirBuoyancies.jl
```diff
@@ -232,10 +230,9 @@ end
     return with_moisture(𝒰₀, q₁)
 end
 
-@inline function saturation_adjustment_residual(T, 𝒰, constants)
+@inline function saturation_adjustment_residual(T, 𝒰::LiquidIcePotentialTemperatureState, constants::ThermodynamicConstants)
     Π = exner_function(𝒰, constants)
```
giordano, line 234: Tightening the signature of this method (and the others related to this) resolves the same issue discussed at https://github.com/NumericalEarth/Breeze.jl/issues/511#issuecomment-3940964127: `exner_function` shouldn't be called with a static energy model, so we inform the whole dispatch mechanism about that

src/AtmosphereModels/cell_advection_timescale.jl
```diff
@@ -65,5 +65,7 @@ function Advection.cell_advection_timescale(model::AtmosphereModel)
 end
 
 # `nothing` schemes advect nothing (no vertical CFL); every other scheme must be AIVA.
-all_vertical_advection_is_implicit(advection::NamedTuple) =
-    all(scheme -> scheme === nothing || needs_implicit_solver(scheme), values(advection))
+# Note: `all` follows the three-valued logic and _may_ return `missing` in some cases.  Let's
+# inform the compiler with the `::Bool` annotation that we know we only deal with booleans.
+all_vertical_advection_is_implicit(advection::NamedTuple)::Bool =
+    all(scheme -> scheme === nothing || needs_implicit_solver(scheme), values(advection))::Bool
```
giordano, line 71: This is similar to https://github.com/CliMA/Oceananigans.jl/pull/5806

src/MoistAirBuoyancies.jl
```diff
@@ -222,7 +220,7 @@ Solution of ``r(T) = 0`` is found via the [secant method](https://en.wikipedia.o
     return secant_solve(residual, solver, T₁, T₂, T₂)
 end
 
-@inline function adjust_state(𝒰₀, T, constants)
+@inline function adjust_state(𝒰₀::LiquidIcePotentialTemperatureState, T, constants::ThermodynamicConstants)
```
glwagner, line 223: I think this should be `Union{LiquidIcePotentialTemperatureState, StaticEnergyState}`. But we may also want to create an abstract type, like `AbstractPerturbationState`, to indicate the state pertubs from a reference state and should have a property `reference_pressure` (or we can create a `reference_pressure(state)` getter)

src/MoistAirBuoyancies.jl
glwagner: I think this is right. But what about `LiquidIceDensityState` ?

src/MoistAirBuoyancies.jl
glwagner: Heh ok claude thinks my name is bad and suggests `AbstractReferencePressureState`. Basically this class of states represents the entire thermodynamic state, but in the presence of a reference.

glwagner: Bah nevermind I am wrong. I didn't realize this was living in `MoistAirBuoyancies.jl`.

glwagner (approved):

src/MoistAirBuoyancies.jl
giordano: https://github.com/NumericalEarth/Breeze.jl/blob/89a1ebf8cf865ca08252d8f99db9be963ecbacfb/src/MoistAirBuoyancies.jl#L178-L181 already constrained `𝒰₀` to `LiquidIcePotentialTemperatureState` only, I didn't change that.  Arguably the compiler should be able to see that `saturation_adjustment_residual` is not going to be called with other states, but having the same consistent signature doesn't hurt if it makes its life easier.

glwagner: yes, and sorry again, i missed that this was constrained to `MoistAirBuoyancies`, for which we only support the liquid ice potential temperature state.

PR #863, opened 2026-07-20
Unify the compressible reference state and set the perturbation-form PGF as default

### LINKED ISSUE #862: Unify the compressible reference state: one grid-polymorphic slot, drop the redundant terrain_reference_*
## Summary

`CompressibleDynamics` represents its hydrostatic reference state **twice** — a 1-D `ExnerReferenceState` for flat grids and a separate pair of loose 3-D fields (`terrain_reference_pressure`/`terrain_reference_density`) for terrain grids — for one concept: the fixed, hydrostatically-balanced reference the pressure-gradient force subtracts. The split forces the redundant terrain fields and overloads `reference_state === nothing` as a "this is a terrain grid" flag. This issue collapses it to one grid-polymorphic reference slot.

## Current state

Two representations of the same thing, chosen by grid:

- **Flat grids** → `reference_state :: ExnerReferenceState` (compressible_dynamics.jl:45); its `pressure`/`density` are 1-D columns, `Field{Nothing,Nothing,Center}` (reference_states.jl:745) — one profile broadcast to every column.
- **Terrain grids** → `terrain_reference_pressure`/`terrain_reference_density` (compressible_dynamics.jl:49-50), loose 3-D `CenterField`s, because a single 1-D column isn't hydrostatically consistent per terrain column.

They are **mutually exclusive** — `materialize_dynamics` forces `reference_state = nothing` whenever terrain metrics are present and builds `terrain_reference_*` only then (compressible_dynamics.jl:210-246) — and `reference_state === nothing` is then relied on as a dispatch flag in ~a dozen kernel methods (`reference_densityᶜᶜᶜ(…, ::Nothing)`, `∂z_reference_pressureᶜᶜᶠ(…, ::Nothing)`, the `terrain_compressible_physics.jl` `::Nothing` set, `seed_pressure!` gated on `reference_state isa ExnerReferenceState`, …) to mean "terrain path, skip the flat base-state machinery." One concept, two field slots, two accessor paths, and an overloaded `nothing`-as-terrain sentinel.

## Solution

End state — `CompressibleDynamics` carries one `reference_state`, nothing else:

```julia
mutable struct CompressibleDynamics{...}
    ...
    reference_state   # ExnerReferenceState (fields 1-D on flat, 3-D on terrain) or Nothing
    terrain_metrics   # already exists — the sole "is this terrain?" signal
    # terrain_reference_pressure / terrain_reference_density  ← DELETED
end
```

**1. Make `ExnerReferenceState` grid-polymorphic.** Its `pressure`/`density`/`exner_function` become `Field{Nothing,Nothing,Center}` on flat grids and `CenterField` on terrain grids, chosen by the constructor from the grid. The per-column 3-D hydrostatic integration already exists in `compute_terrain_reference_state!` — fold it in as the terrain branch of the constructor (the flat branch is today's 1-D column integration). Consumers keep indexing `ref.pressure[i,j,k]` unchanged: a reduced field broadcasts the horizontal indices, so the same kernel line works for both dimensionalities.

**2. Delete `terrain_reference_pressure`/`terrain_reference_density`.** The single `reference_state` now holds the reference at whatever dimensionality the grid needs. `materialize_dynamics` stops building the separate 3-D fields and just calls the (now grid-aware) `ExnerReferenceState` constructor.

**3. Move terrain-vs-flat dispatch off `reference_state` and onto `terrain_metrics`.** This unblocks the unification. Today `reference_state === nothing` means *both* "no reference configured" *and* "terrain grid." Split them: terrain-ness is already carried by `dynamics.terrain_metrics`, so the `::Nothing`/`::ExnerReferenceState` kernel methods dispatch on `terrain_metrics` for the metric/slope terms and read `reference_state` purely for the base-state subtraction. Afterward `reference_state === nothing` means *only* "no reference configured" (→ full-pressure PGF, no base-state removal) on any grid.

**4. Radiation is already decoupled (#791)** — compressible reads `dynamics.pressure`/`total_density`, anelastic reads `reference_state.*`, neither touching `terrain_reference_*`. Nothing to do here.

Optional call inside step 1: store the 3-D Exner field on terrain (memory) vs. recompute `Π = (p/pˢᵗ)^κ` in-kernel. Recommend storing it, matching the flat path; revisit only if memory bites.

## Related

The `reference_pressure`/`reference_density` accessor **names** are now misleading — since #791 they return the actual total, not a reference. Small separate follow-up: radiation can call `mean_pressure` + a `total_density` accessor and delete them.

kaiyuan-cheng:
`CompressibleDynamics` now carries a single hydrostatic reference state and uses the perturbation-form pressure gradient and buoyancy (p′ = p − pᵣ, ρ′ = ρ − ρᵣ) by default. This suppresses the small-difference-of-large-numbers cancellation in the terrain-following PGF that otherwise drives spurious slope-correlated surface winds over steep terrain.

## One reference state, built by default

- `terrain_reference_pressure`/`terrain_reference_density` are gone. `reference_state` is a single grid-polymorphic `ExnerReferenceState`: a 1D column on height-coordinate grids, horizontally-varying 3D fields on terrain-following grids (where one column cannot be in hydrostatic balance for every terrain height). Terrain-vs-flat dispatch keys on `terrain_metrics` alone, so both paths now share `reference_densityᶜᶜᶜ`/`∂z_reference_pressureᶜᶜᶠ` — the terrain-specific buoyancy/`z_pressure_gradient` overrides and the `terrain_reference_*` accessors are removed.
- `reference_state = :auto` (the default) builds a standard-atmosphere (θᵣ = 288 K) hydrostatic reference on any bounded vertical grid, so `model.dynamics.reference_state` is usable immediately after construction and a rest state is discretely balanced to machine precision. Periodic and flat vertical topologies get no automatic reference: gravity makes a nontrivial hydrostatic column nonperiodic, and a flat vertical dimension needs no hydrostatic split.
- `reference_state = nothing` disables the reference entirely — the PGF and buoyancy then difference the full pressure. Mutually exclusive with an explicit reference profile; anything other than `:auto`/`nothing` throws.
- An explicit `reference_potential_temperature` (and optional `reference_vapor_mass_fraction`) replaces the automatic profile. It is authoritative: `set!` never overwrites it implicitly.
- The reference is also built on demand for boundary conditions, which are materialized before the dynamics, so stability-dependent bulk fluxes can use it.

## Recomputation from the initial state is explicit

`set!(model; …, compute_reference_state=true)` replaces the reference with one deduced from the state's height-resolved horizontal mean; the default (`false`) preserves the reference built at construction. Nothing is recomputed implicitly.

`horizontal_mean_profile` interpolates every column onto a common set of physical heights — the columnwise minima of the cell-center heights, i.e. the lowest-terrain column's levels — and averages there over the columns that actually have air at each height. That gives a genuine θ̄(z), instead of a per-model-level mean that blends valley-floor and mountain-top air at the same `k`. Extending raised columns downward instead would bias the mean towards near-surface air by ~(z̄ − z_surface)·dφ/dz at the lowest levels.

## All grid-point work in kernels

- The terrain reference solve is a kernel over columns writing straight into `Field`s — no host arrays, no CPU grid mirror — and shares its interior recurrence with the height-coordinate reference via the extracted `integrate_exner_column!`.
- The horizontal mean is a `minimum!` reduction over a `KernelFunctionOperation` of `znode`, one `:xy` interpolation kernel with a monotone O(Nz) walk, and two `sum!` reductions. The previous height binning and empty-bin filling are gone.
- Reference profiles stay host callables (an arbitrary user `θ(z)`, or the deduced mean profile) but are sampled onto `Field`s at each column's own physical heights via a small `HeightProfile` adapter; the per-column surface anchor is one vectorized pass over a kernel-computed surface-height field, one evaluation per column rather than per cell.
- Every column is now anchored at its terrain surface with the continuous hydrostatic pressure and closed by the discrete-balance Newton solve from there upward. Previously only moist columns were, so dry columns gain a discretely balanced surface-to-first-center half cell.

## Behavior change and tests

Compressible runs that previously differenced the full pressure now use the perturbation form by default. Tests are updated accordingly: operator-identity tests and the flat-vs-terrain identity tests opt out with `reference_state = nothing`, since they assert properties of the full-pressure operator. Added coverage for the `reference_state` switch and its invalid values, for explicit recomputation via `set!`, for the reference being available to boundary conditions, and for the height-resolved mean (a field linear in physical height is recovered exactly over terrain). The moist terrain reference-state tests now run under `default_arch` rather than `CPU()`, since the solve is GPU-capable.

Closes #862.


src/CompressibleEquations/compressible_dynamics.jl
```diff
@@ -48,8 +50,14 @@ struct CompressibleDynamics{TD, D, DT, P, FT, RS, TM, CV, CM, TRP, TRD}
     contravariant_vertical_momentum :: CM      # ρw̃ diagnostic field (or Nothing)
     terrain_reference_pressure :: TRP          # 3D reference pressure for terrain PG (or Nothing)
     terrain_reference_density :: TRD           # 3D reference density for terrain buoyancy (or Nothing)
+    reference_from_state :: Bool               # terrain reference deduced from the initial state (no explicit spec)
```
glwagner: I propose using a design that might be more streamlined:

1. condense the "reference state" to a single property via #862 
2. build a reference state by default: 3D for terrain following, 1D otherwise?
3. allow `reference_state=nothing` when building `CompressibleDynamics`. This means it is not used!
4. use `compute_reference_state=true/false` in `set!` to control "recomputation" of the reference state

kaiyuan-cheng: I've reworked on this PR.:
1. Deleted terrain_reference_pressure/terrain_reference_density; reference_state is now one ExnerReferenceState (1-D column flat, 3-D on terrain). Terrain-vs-flat dispatch keys on terrain_metrics. Fell out of this: the terrain buoyancy/z_pressure_gradient overrides and terrain_reference_* accessors are gone — flat & terrain share reference_densityᶜᶜᶜ/∂z_reference_pressureᶜᶜᶠ.
2. Built by default — 3-D on terrain, 1-D otherwise (I built on flat too; perturbation form is now the default everywhere, discretely balanced so rest state is machine-precision).
3. reference_state = nothing disables it.

glwagner: great!

glwagner: does the title of the PR need to be updated?

kaiyuan-cheng:
close #862

src/AtmosphereModels/dynamics_interface.jl
```diff
@@ -249,6 +249,22 @@ Return the dynamics' reference state (an anelastic `ReferenceState` or a split-e
...
+either absent or supplied up front. `true` for compressible dynamics *that were not given an
+explicit reference profile* on a compatible bounded vertical grid (the default
+`reference_state = :auto`): their valid provisional reference is best replaced by one deduced from
+the initial state's horizontal mean. Automatic reset occurs only when `set!` supplies both density
+and a thermodynamic variable, so partial initialization remains valid. When an explicit reference
+profile was supplied at construction it is authoritative, so this stays `false` and `set!` leaves
+it untouched. An explicit `compute_reference_state` keyword to `set!` overrides this default either
+way.
+"""
+auto_reset_reference_state(dynamics) = false
```
glwagner: do we still need this helper?

src/AtmosphereModels/set_atmosphere_model.jl
```diff
@@ -193,7 +198,7 @@ Variables are set via keyword arguments. Supported variables include:
   `CompressibleDynamics` and `AnelasticDynamics`.
 """
 function Fields.set!(model::AtmosphereModel; time=nothing, enforce_mass_conservation=true,
-                     compute_reference_state=false, balancer=false, kw...)
+                     compute_reference_state=nothing, balancer=false, kw...)
```
glwagner:
```suggestion
                     compute_reference_state=false, balancer=false, kw...)
```

src/AtmosphereModels/set_atmosphere_model.jl
```diff
@@ -335,8 +340,17 @@ function Fields.set!(model::AtmosphereModel; time=nothing, enforce_mass_conserva
...
+    # reset waits until this call supplies both density and a thermodynamic variable. Either one
+    # alone is a partial initialization whose complementary field may still be zero, so integrating
+    # it could corrupt the valid provisional reference. An explicit keyword overrides this safeguard
+    # because it is an intentional request to use the model's current state.
+    density_state_updated = any(name -> name ∈ (:ρ, :ρᵈ), names)
+    thermodynamic_state_updated = any(name -> name ∈ settable_thermodynamic_variables, names)
+    complete_thermodynamic_state = density_state_updated && thermodynamic_state_updated
+    automatic_reset = auto_reset_reference_state(model.dynamics) && complete_thermodynamic_state
+    should_reset = something(compute_reference_state, automatic_reset)
+    if should_reset
```
glwagner:
```suggestion
    thermodynamic_state_updated = any(name -> name ∈ (:ρ, :ρᵈ), names) &&any(name -> name ∈ settable_thermodynamic_variables, names)
    can_compute = something(model.dynamics.reference_state)
    if can_compute && thermodynamic_state_updated && compute_reference_state
```

I think this is a bit clearer and easier to parse

src/CompressibleEquations/terrain_compressible_physics.jl
```diff
@@ -89,26 +102,68 @@ function horizontal_mean_profile(field)
...
-        value_sum = zero(FT)
-
-        for j in 1:Ny, i in 1:Nx
-            height_sum += znode(i, j, k, cpu_grid, c, c, c)
-            value_sum += data[i, j, k]
+    # Locate the lowest-terrain column (minimum bottom-face height): its faces span the full
+    # physical-height range from the bottom and define the bin edges.
+    imin, jmin = 1, 1
+    lowest_surface = FT(Inf)
+    for j in 1:Ny, i in 1:Nx
```
glwagner: Can we put this in a kernel? We shouldn't have `for` loops outside kernels

glwagner: This can possibly be computed as a reduction of some kind

src/CompressibleEquations/terrain_compressible_physics.jl
```diff
@@ -89,26 +102,68 @@ function horizontal_mean_profile(field)
...
 
-        heights[k] = height_sum * normalization
-        values[k] = value_sum * normalization
+    edges = FT[znode(imin, jmin, k, cpu_grid, c, c, f) for k in 1:Nz+1]
+    fallback_heights = FT[znode(imin, jmin, k, cpu_grid, c, c, c) for k in 1:Nz]
+
+    value_sums = zeros(FT, Nz)
+    height_sums = zeros(FT, Nz)
+    counts = zeros(Int, Nz)
+    for k in 1:Nz, j in 1:Ny, i in 1:Nx
```
glwagner: can this be implemented in a kernel?

src/CompressibleEquations/terrain_compressible_physics.jl
```diff
@@ -826,6 +870,7 @@ function compute_terrain_reference_state!(pᵣ, ρᵣ, grid, p₀, ref_spec, pˢ
     FT = eltype(grid)
     p_host = zeros(FT, Nx, Ny, Nz)
     ρ_host = zeros(FT, Nx, Ny, Nz)
+    π_host = zeros(FT, Nx, Ny, Nz)
```
glwagner: Can we compute these on-device?

src/CompressibleEquations/terrain_compressible_physics.jl
```diff
@@ -826,6 +870,7 @@ function compute_terrain_reference_state!(pᵣ, ρᵣ, grid, p₀, ref_spec, pˢ
...
 
     for j in 1:Ny, i in 1:Nx
```
glwagner: this should be implemented in a kernel

src/CompressibleEquations/terrain_compressible_physics.jl
glwagner: shoudl also use `Field(grid)`, not raw arrays

glwagner (approved):

ewquon:
Heads up that the perturbation form this PR makes default is being undercut from the caller side in NumericalEarth: `mean_surface_pressure` passes a domain-mean *terrain-surface* pressure into `surface_pressure`, which Breeze defines at z = 0, so the hydrostatic reduction to each column's terrain height is applied twice.

Measured on a nested child over a Colorado box (mean terrain 1921 m): `pᵣ` at k=1 comes out 16.2 kPa below the actual pressure (20 % of `p`); correcting the anchor brings it to 1.6 kPa (2 %). Details and reproduction in NumericalEarth/NumericalEarth.jl#488.

Nothing to change in Breeze as far as I can tell, the contract is implemented and tested correctly here. Flagging it since it affects how much of this PR's benefit actually lands over elevated terrain.

PR #864, opened 2026-07-22
[CI] Use multiple GC threads on GPU runners

giordano:


giordano:
This seems to halve the time spent in the garbage collector for `reactant/centered_compilation` (the most GC-intensive test job) from 80-90 seconds to 45 seconds, not too bad.

PR #867, opened 2026-07-23
Collapse mean_pressure into dynamics_pressure

kaiyuan-cheng:
## Summary

The `AtmosphereModel` dynamics interface exposed **two** field accessors for the same
pressure field:

- `dynamics_pressure(dynamics)`
- `mean_pressure(dynamics)`

For every dynamics type they returned the identical field:

| Dynamics | both accessors returned |
|----------|-------------------------|
| `AnelasticDynamics`    | `reference_state.pressure` |
| `CompressibleDynamics` | `dynamics.pressure` |
| `ParcelDynamics`       | `dynamics.pressure` |
| `PrescribedDynamics`   | `dynamics.pressure` |

Two exported names, same return value, different call sites, no semantic difference —
a "which one do I call?" trap. This PR collapses them into one, keeping
**`dynamics_pressure`**.

Why `dynamics_pressure` and not `mean_pressure`: the name says which pressure the
dynamical core hands to the equation of state, buoyancy, and the thermodynamic
tendencies, without asserting that the field *is* a mean. That assertion only holds
for the anelastic reference state; under `CompressibleDynamics` the returned field is
the full prognostic pressure with no background/anomaly split, so "mean" is actively
misleading there. It also reads as a pair with the sibling accessor `dynamics_density`,
which follows the same "field appropriate to the formulation" convention.

`pressure_anomaly` and `total_pressure` keep their names and meanings (anomaly =
deviation from the hydrostatic background; total = background + anomaly).

## Changes

- Remove the four identical `mean_pressure` methods and the interface declaration.
- Merge the two docstrings into the surviving `dynamics_pressure` docstring: units,
  what consumes the field, the per-formulation meaning, and `@ref` cross-links to
  `pressure_anomaly` / `total_pressure`.
- Keep the surviving method in each dynamics file grouped under its `##### Pressure
  interface` section alongside `pressure_anomaly` / `total_pressure`, rather than
  stranded next to the density accessors.
- Repoint the two remaining `mean_pressure` call sites (`total_pressure` for
  `AnelasticDynamics`, and the unit test) to `dynamics_pressure`.
- Drop the `mean_pressure` export from `AtmosphereModels`. `Breeze` already exported
  `dynamics_pressure`, so the top-level public API is unchanged.

The unrelated `dynamics_pressure_solver` and `dynamics_pressure_for_potential_temperature`
are **not** touched.

## Behavior

Pure rename/dedup — every removed method returned the same field its `dynamics_pressure`
twin already returned, so there is no behavior change.

## Verification

- Package precompiles and loads clean; `Breeze.dynamics_pressure` is exported and
  resolves all four methods; `isdefined(Breeze, :mean_pressure)` is `false`.
- No `mean_pressure` occurrences remain anywhere in the repo (`.jl`, `.md`, `.toml`, `.yml`).
- CI covers the rest: `quality_assurance` (ExplicitImports, Aqua), `unit_tests`, and doctests.

### DIFF
--- a/src/AnelasticEquations/AnelasticEquations.jl
+++ b/src/AnelasticEquations/AnelasticEquations.jl
@@ -35,7 +35,7 @@ using Oceananigans.Utils: prettysummary, launch!
 
 using Breeze.Thermodynamics: ReferenceState, MoistureMassFractions, mixture_gas_constant,
                              dry_air_gas_constant
-using Breeze.AtmosphereModels: AtmosphereModels, AtmosphereModel, mean_pressure, pressure_anomaly,
+using Breeze.AtmosphereModels: AtmosphereModels, AtmosphereModel, dynamics_pressure, pressure_anomaly,
                                 kernel_time_step
 
 # Import microphysics interface for buoyancy computation
--- a/src/AnelasticEquations/anelastic_dynamics.jl
+++ b/src/AnelasticEquations/anelastic_dynamics.jl
@@ -115,16 +115,6 @@ reference state density ``ρᵣ(z)``.
 """
 AtmosphereModels.dynamics_density(dynamics::AnelasticDynamics) = dynamics.reference_state.density
 
-"""
-$(TYPEDSIGNATURES)
-
-Return the reference pressure field for `AnelasticDynamics`.
-
-For anelastic models, the dynamics pressure is the time-independent
-hydrostatic reference state pressure ``pᵣ(z)``.
-"""
-AtmosphereModels.dynamics_pressure(dynamics::AnelasticDynamics) = dynamics.reference_state.pressure
-
 #####
 ##### Prognostic fields
 #####
--- a/src/AnelasticEquations/anelastic_dynamics.jl
+++ b/src/AnelasticEquations/anelastic_dynamics.jl
@@ -50,9 +50,12 @@ end
 """
 $(TYPEDSIGNATURES)
 
-Return the mean (reference) pressure field for `AnelasticDynamics`, in Pa.
+Return the dynamics pressure field for `AnelasticDynamics`, in Pa.
+
+For anelastic models, this is the time-independent hydrostatic reference state
+pressure ``pᵣ(z)``.
 """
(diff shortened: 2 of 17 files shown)


giordano (approved):
This is fine by me, the two functions were indeed doing the same thing, but just a note: as I indicated with the label, this is a breaking change because it's removing public API, so next version of Breeze will have to bump the minor version (v0.8.0)

glwagner:
is it true that the `pressure_anomaly` always has a zero mean?

kaiyuan-cheng:
> is it true that the `pressure_anomaly` always has a zero mean?

I'm not quite sure I understand this question. What is the purpose of taking the mean of `pressure_anomaly`?

glwagner:
> > is it true that the `pressure_anomaly` always has a zero mean?
> 
> I'm not quite sure I understand this question. What is the purpose of taking the mean of `pressure_anomaly`?

I guess I am wondering if this decomposition is always between a "mean" pressure and the anomaly away from a mean. In general, we often use decompositions into a reference and an anomaly, but the reference does not represent the average of any field. Is the dynamics pressure always guaranteed to be the average of something?

kaiyuan-cheng:
> > > is it true that the `pressure_anomaly` always has a zero mean?
> > 
> > 
> > I'm not quite sure I understand this question. What is the purpose of taking the mean of `pressure_anomaly`?
> 
> I guess I am wondering if this decomposition is always between a "mean" pressure and the anomaly away from a mean. In general, we often use decompositions into a reference and an anomaly, but the reference does not represent the average of any field. Is the dynamics pressure always guaranteed to be the average of something?

No. For `CompressibleDynamics`, the accessor returns the full pressure field and `pressure_anomaly` is 0, so `mean_pressure` is really total pressure there. For `AnelasticDynamics`, it's `reference_state.pressure`. Nothing constrains the horizontal mean of `p′` to vanish at each level. It's reference-vs-anomaly, as you say, not a Reynolds decomposition.

`mean_pressure` is a misleading name indeed. We may want to come up with a better name.

kaiyuan-cheng:
I think we should use `dynamics_pressure`, instead of `mean_pressure`.

glwagner (approved):

Issue #847, opened 2026-07-15
`TKEBasedTurbulenceClosure`


ewquon:
This issue is for discussing TKE-based turbulence closures for `AtmosphereModel`: what a working branch implements today, the literature each piece comes from, and complementary features we propose to add. A PR for the implemented portion follows shortly.

## What's been explored so far

A `TKEBasedDiffusivities` module (mirroring the Oceananigans `TKEBasedVerticalDiffusivities` conventions) providing a scale-adaptive TKE + mass-flux closure — one closure spanning mesoscale-PBL → gray-zone → LES resolutions:

- **Prognostic subgrid TKE** (1.5-order closure; Mellor & Yamada 1982; Deardorff 1980), carried as a `:tke` tracer that `AtmosphereModel` adds automatically through a new generic `closure_required_tracers` hook. Patankar-positive source update, surface-similarity TKE floor.
- **Blended master length scale**: a boundary-layer branch (harmonic wall + Blackadar (1962) asymptotic pair) and an LES branch (Deardorff (1980) grid scale), both limited by the stable-buoyancy length `Cb·√e/N` (Deardorff 1980).
- **Resolution weight** `W = e/(e + C_W·k_r)` with `k_r` the locally diagnosed resolved kinetic energy — a *solution-triggered* partition in the PITM/PANS spirit (Schiestel & Dejoan 2005; Girimaji 2006) rather than a grid-triggered switch, addressing the gray zone / terra incognita (Wyngaard 2004; Honnert et al. 2020) with the lesson of the aerospace DES → SAS progression (Spalart 2009; Menter & Egorov 2010): single-column configurations recover the exact PBL limit, fine 3-D grids the LES limit, with no imposed transition scale.
- **Dissipation** `ε = Ce(W)·e^{3/2}/ℓ` blending the Mellor–Yamada B₁ PBL limit with the Deardorff LES limit; turbulent Prandtl number growing with Ri in the Louis (1979)-family tradition.
[...]

PR #875, opened 2026-07-29
Add TKEBasedTurbulenceClosure, a CATKE-style prognostic-TKE closure with constant stability functions

### LINKED ISSUE #870: Closure-field halos are never filled, halving the vertical viscous stress in the first column
`AtmosphereModel` never calls `fill_halo_regions!` on `model.closure_fields`. Every Oceananigans
model does so immediately after computing diffusivities
(`update_nonhydrostatic_model_state.jl:50`); Breeze's `compute_auxiliary_variables!` omits it.

`νₑ` is stored at one location and consumed at several, so most accessors reach it **horizontally**
and therefore read the halo. The exact stencil depends on `viscosity_location`: for a closure
storing `νₑ` at `(Center, Center, Center)` — `SmagorinskyLilly`, `AnisotropicMinimumDissipation` —
the vertical stress on `u` uses `νᶠᶜᶠ = ℑxzᶠᵃᶠ(νₑ)`, a four-point average that includes
`ν[i-1, j, k-1]` and `ν[i-1, j, k]`; the horizontal stress `𝒯_ux` reads `ν[i, j, k]` directly but is
then differenced in `x`, so the tendency at `i = 1` still reaches `ν[0, j, k]`. `νᶠᶠᶜ`, `κᶠᶜᶜ` and
`κᶜᶠᶜ` interpolate horizontally too.

A zero halo therefore under-weights the stress in the first interior column, and at every rank
boundary under MPI. This affects existing runs and is not specific to any one closure.

## Reproduction

```julia
using Breeze, Oceananigans

grid = RectilinearGrid(CPU(); size=(8, 8, 8), extent=(1, 1, 1))

for closure in (SmagorinskyLilly(), AnisotropicMinimumDissipation())
    model = AtmosphereModel(grid; closure)
    θ₀ = model.dynamics.reference_state.potential_temperature
    set!(model; θ=θ₀, ρu=(x, y, z) -> sin(2π*x) * cos(2π*y) + z)
    time_step!(model, 1e-3)

    νₑ = model.closure_fields.νₑ
    Hx = νₑ.grid.Hx
    @show nameof(typeof(closure)),
          maximum(abs, interior(νₑ)),
          maximum(abs, parent(νₑ)[1:Hx, :, :])   # west halo
end
```

On Breeze `2351319c` (v0.7.4):

| closure | interior `maximum(abs, νₑ)` | west halo |
|---|---|---|
| `SmagorinskyLilly` | 0.001745 | **0.0** |
| `AnisotropicMinimumDissipation` | 0.002829 | **0.0** |

Filling the halo makes the two agree.

## Why it went unnoticed

No test inspects a halo value of `νₑ`. The interior is correct, so every assertion written against
`interior(...)` passes.

## Fix

One line in `compute_auxiliary_variables!`
(`src/AtmosphereModels/update_atmosphere_model_state.jl`), a no-op when
`closure_fields === nothing`:

```julia
    compute_closure_fields!(model.closure_fields, model.closure, model)
    fill_halo_regions!(model.closure_fields; only_local_halos=true)
```

Closures storing their fields in a struct rather than a `NamedTuple` need a
`BoundaryConditions.fill_halo_regions!` method, as CATKE has.

Suggested regression test: assert the halo of `νₑ` matches the interior after `update_state!`, for a
closure with spatially varying viscosity.

### LINKED ISSUE #871: Vertically-implicit diffusion is not mass-flux weighted, giving a spurious `c ∝ 1/ρ` steady state
Breeze's prognostic scalars are density weighted (`q = ρc`). The **explicit** flux divergence forms
the mass-flux operator on the specific variable, `∂t(ρc) = ∂z(ρ κ ∂z c)` — `Jᶜz` in
`src/TurbulenceClosures/TurbulenceClosures.jl` weights the kinematic flux by `ℑz(ρ)`. The
**vertically-implicit** solve instead applies `∂t q = ∂z(κ ∂z q)` directly to `q`, because the
tridiagonal coefficients it is built from — `ivd_upper_diagonal`, `ivd_lower_diagonal` and
`ivd_diagonal` in `vertically_implicit_diffusion_solver.jl:66-118` — carry no density, and Breeze
forwards `density` only to the *advection* half of the combined coefficients.

Mixing should drive the *specific* variable to a constant, `∂z c = 0`. The unweighted operator
instead drives the *density-weighted* one to a constant, `ρc = const`, so the specific variable
grows with height as `c ∝ 1/ρ` — the two operators agree only for constant `ρ`. For potential
temperature, a well-mixed layer would be reported as stably stratified.

## Reproduction

4 km column (ρ varies ≈ 40%), zero surface flux, run to steady state:

```julia
grid = RectilinearGrid(CPU(); size=(1, 1, 32), x=(0, 100), y=(0, 100), z=(0, 4000),
                       topology=(Periodic, Periodic, Bounded))
closure = VerticalScalarDiffusivity(VerticallyImplicitTimeDiscretization(); κ=200)
model = AtmosphereModel(grid; closure, advection=nothing, tracers=:ρc)
set!(model; ρc = (x, y, z) -> 1 + cos(π * z / 4000))
for _ in 1:200; time_step!(model, 2000); end
```

Observed: `ρc` relaxes to uniform (spread 3×10⁻¹⁵) while `c = ρc/ρ` is left **36% non-uniform**.
Expected the reverse. Implicit-vs-explicit on the same column differ by ≈ 6% — against an
estimated ≈ 0.3% from the backward-vs-forward Euler difference alone (`κk²Δt` compounded over
the run), so the discrepancy is not a time-discretization artifact.

## Why it went unnoticed

`test/vertical_diffusion.jl` runs on `Lz = 100 m`, where ρ varies < 1% — inside every tolerance in
the file. Its metric is an L2-norm *ratio*, which cancels most of the profile error. And every case
uses `advection = nothing`, so `needs_implicit_solver(advection)` is `false` and the solve takes the
no-density `implicit_step!` branch.

## Fix

Each coefficient is the upstream one times a density ratio, so this composes with closure tuples
(whose κ-sum happens inside `_ivd_*_diagonal`) and with `closure === nothing`:

```
du(k)  = -Δt κₖ₊₁ (ρᶠₖ₊₁ / ρᶜₖ₊₁) Δz⁻¹ᶜₖ Δz⁻¹ᶠₖ₊₁
dl(k′) = -Δt κₖ   (ρᶠₖ   / ρᶜₖ₋₁) Δz⁻¹ᶜₖ Δz⁻¹ᶠₖ,      k = k′ + 1
d(k)   =  1 + Δt Δz⁻¹ᶜₖ [κₖ₊₁ ρᶠₖ₊₁ Δz⁻¹ᶠₖ₊₁ + κₖ ρᶠₖ Δz⁻¹ᶠₖ] / ρᶜₖ
```

Two traps:

- The diagonal **must be written out explicitly**, not as `1 - du - dl`. `ρᶜ` belongs to the
  *column*, not the row — `du(k)` divides by `ρᶜₖ₊₁` while the diagonal divides by `ρᶜₖ` — so
  `1 - du - dl` breaks conservation of `Σ Δz ρc`.
- Call sites **cannot** be keyed on `needs_implicit_solver(advection)`, which is `false` for both
  `advection = nothing` and ordinary WENO. Keying on it drops the weighting for every closure
  carrying `VerticallyImplicitTimeDiscretization` that is not also using adaptive-implicit
  advection. That predicate was introduced (Oceananigans `7fd634a9d`, #5604) purely to decide at
  *construction* whether to allocate a solver the closure alone would not need — an allocation
  question about the advection scheme — so it is the wrong question to ask per substep. Guarding on
  `implicit_solver !== nothing` asks the right one.

`ρw` (z-Face) is separate: upstream's stencil evaluates `ν` at a single center for both
off-diagonals of a row (`vertically_implicit_diffusion_solver.jl:95-112`), so there is no consistent
center at which to evaluate `ρ`. Best resolved upstream first.

## Suggested regression tests

On a deep column (≥ 2 km), three checks — five assertions in total:

1. implicit vs explicit agree, separately for `κ` and `ν` (2);
2. the steady state is well mixed in the **specific** variable, and *not* in the density-weighted
   one (2);
3. `Σ Δz ρc` is conserved under zero surface flux, on a **stretched** grid (1).

Four of the five fail against the unfixed code. Only conservation passes either way — the
unweighted operator is still a flux divergence, so it conserves `Σ Δz ρc` too; that assertion
guards the diagonal's form, not the weighting.

ewquon:
Implements the eddy-diffusivity portion of #847, in the spirit of CATKE ([Wagner et al. 2025](https://doi.org/10.1029/2024MS004522)).

Adds `TKEBasedTurbulenceClosure`, a vertical eddy-diffusivity closure carrying one prognostic equation for the subgrid turbulent kinetic energy — the tracer `ρe`, which the closure appends to the model:

```
Kᵘ = Sᵘ ℓ √e,   Kᶜ = Sᶜ ℓ √e,   Kᵉ = Sᵉ ℓ √e,   ε = Sᴰ e^{3/2} / ℓ
ℓ  = min(z, Cᴺ √e / N)
∂ₜ(ρe) + ∇·(ρ u e) = ∂z(ρ Kᵉ ∂z e) + ρ (P + B − ε),   P = Kᵘ S²,   B = −Kᶜ N²
```

Each diffusivity is a *stability function* times one primary mixing length — `ℓᵘ = Sᵘ ℓ`, `ℓᶜ = Sᶜ ℓ`, `ℓᵉ = Sᵉ ℓ` — and the dissipation length is `ℓᴰ = ℓ / Sᴰ`. In this first version the stability functions are constants, held in `ConstantStabilityFunctions(Cᵘ, Cᶜ, Cᵉ, Cᴰ)` and dispatched so that Richardson-number-dependent functions, as in CATKE, are a new type plus four methods. The mixing length is hard-coded: the height above the surface — carrying no coefficient of its own — limited by the stratification length `Cᴺ √e / N`, which is infinite in neutral and unstable air. Nothing inside the closure reads a surface flux, a friction velocity or a roughness length.

The defaults are the MYNN coefficients with the von Kármán constant absorbed — `Cᵘ = 0.196`, `Cᶜ = 0.265`, `Cᵉ = 0.392`, `Cᴰ = 0.295`, `Cᴺ = 0.76` — implying `κ = (Cᵘ³/Cᴰ)^¼ = 0.40`, `e/u★² = 1/√(Cᵘ Cᴰ) = 4.2`, `Pr = Cᵘ/Cᶜ = 0.74` and a stratified steady state at `Ri† = Cᵘ Cᴺ²/(Cᶜ Cᴺ² + Cᴰ) = 0.25`. They are placeholders for calibration.

**Numerics.** These follow CATKE. The sinks — the dissipation rate `√|e|/ℓᴰ`, or the damping rate `1/τ` where advection has driven `e` negative, plus the negative part of the buoyancy flux, Patankar-gated by `e > minimum_tke` — form a linear implicit coefficient that enters every stage's vertically implicit tridiagonal solve together with the vertical diffusion of `e`, so `e` stays positive for any time step and is never clipped. The sources — shear production and the positive part of the buoyancy flux — enter the stage tendency through a new `compute_closure_tendencies!` hook that both time steppers call alongside the flux boundary conditions. The surface TKE condition is zero-flux for now.

**Scope.** Eddy diffusivity only: every subgrid flux is downgradient, so a convective boundary layer keeps a superadiabatic interior and stable boundary layers run deep — limitations of the formulation, not the implementation. Follow-ups in their own PRs: Richardson-dependent stability functions; a convective/entrainment length driven by the (time-filtered) surface buoyancy flux, together with a CATKE-style surface TKE flux; a `non_local_flux` slot; calibration.

**Docs and validation.** A new "Turbulence closures" page and a single-column example running three published cases: GABLS1, the Shin et al. conventionally neutral boundary layer, and the Han–Bretherton dry convective boundary layer. Scores against all three are in #847, latest table [here](https://github.com/NumericalEarth/Breeze.jl/issues/847#issuecomment-5488316772) — the known signatures of the constant-stability-function starting point (deep stable layer, `ℓ = z` uncapped in neutral columns, over-entrainment without a convective length) are discussed there and motivate the follow-ups.

The two vertically-implicit-diffusion fixes this PR originally carried (#870, #871) merged separately in #925, and the static-energy rename that frees `e`/`ρe` for turbulent kinetic energy merged in #926.


docs/src/appendix/notation.md
```diff
@@ -130,7 +130,29 @@ The following table also uses a few conventions that suffuse the source code and
...
+| ``\ell^b``                          | `ℓᵇ`   |                                     | Buoyancy branch of the mixing length, ``ℓᵇ = Cᵇ q / N``                         |
+| ``\ell^r``                          | `ℓʳ`   | `roughness_length`                  | Surface roughness length for momentum, m. Superscript ``r`` elsewhere denotes *rain* (``qʳ``); the two never appear together |
+| ``\ell^{rh}``                       | `ℓʳʰ`  | `scalar_roughness_length`           | Surface roughness length for heat and moisture; defaults to ``ℓʳ/7.3`` |
+| ``\kappa``                          | `κ`    |                                     | von Kármán constant. Never used for a diffusivity in Breeze's own code — the Oceananigans accessors literally named `κᶠᶜᶜ`, `κᶜᶠᶜ`, `κᶜᶜᶠ` are the one exception, and they are thin one-liners |
+| ``\nu``                             | `ν`    |                                     | Eddy viscosity, m² s⁻¹                                                         |
+| ``K``                               | `K`    |                                     | Eddy diffusivity, ``K = ν/\mathrm{Pr}``, m² s⁻¹                                |
```
glwagner: can we use the same notation as Oceananigans, so K^u for momentum, K^c for scalars (potential temperature and humidity, etc), and K^e for TKE? what do you think?

docs/src/appendix/notation.md
```diff
@@ -130,7 +130,29 @@ The following table also uses a few conventions that suffuse the source code and
...
+| ``C^K``                             | `Cᴷ`   | `diffusivity_coefficient`           | Diffusivity coefficient in ``ν = Cᴷ ℓ \sqrt{e}``                               |
+| ``C^\mu``                           | `Cμ`   |                                     | ``k``–``ε`` coefficient in ``ν = Cμ e²/ε``; sets the neutral surface-layer turbulence level ``e/u_\star² = (Cμ)^{-1/2}`` |
+| ``C^\varepsilon``                   | `Cᵋ`   | `dissipation_coefficient`           | Dissipation coefficient in ``ε = Cᵋ e^{3/2}/ℓ``; *derived*, ``Cᵋ = Cμ/Cᴷ``      |
+| ``C^\mathrm{sfc}``                  |        | `surface_tke_coefficient`           | Surface TKE level ``e/u_\star²``; *derived*, ``(Cμ)^{-1/2}``                    |
+| ``C^s``                             | `Cˢ`   | `stress_coefficient`                | Stress coefficient; *derived*, ``Cᴷ/(Cμ)^{1/4}``. Equals 1 on the neutral log-law locus ``Cμ = Cᴷ⁴`` |
+| ``C^t``, ``C^b``                    | `Cᵗ`, `Cᵇ` |                                 | Coefficients of the turbulence and buoyancy mixing-length branches             |
```
glwagner: probably, we shouldn't put free parameters in the notation list, just because there may eventually be an excessive number of them

giordano:
Is this related to #847?

docs/src/appendix/notation.md
ewquon: Done in fe6f6c74 — `νₑ → Kᵘ`, `κₑ → Kᶜ`, `νₑᵗ → Kᵉ`. That third old name was the tell that the ν/κ pair had no room for a TKE diffusivity, so it had to be spelled as a decorated viscosity.

One thing to confirm: CATKE spells these `κu`, `κc`, `κe` (`CATKEClosureFields`), but Breeze reserves `κ` for the von Kármán constant, which this closure leans on (`GeometricLengthScale.κ`, `von_karman_constant`), and `notation.md` already records that reservation. So I took the structure and kept your letter `K` rather than matching Oceananigans character-for-character. Happy to switch to `κu/κc/κe` if you would rather have exact parity.

`viscosity` and `diffusivity` are the accessors Oceananigans dispatches on, so nothing outside the closure depended on the field names.

docs/src/appendix/notation.md
ewquon: Agreed — removed in 641cb685: `Cᴷ`, `Cμ`, `Cᵋ`, `Cˢᶠᶜ`, `Cˢ`, `Cᵗ`, `Cᵇ`. They are documented on the closure's own docstring and page, which is where they scale as more schemes land.

Where a coefficient still appears inside a formula in another row (`ℓᵗ = Cᵗ ∫qz dz / ∫q dz`, `ℓᵇ = Cᵇ q / N`) I left it: the row defines the length scale, not the coefficient.

ewquon:
@giordano yes — #847 is the design discussion for the closure family, and this PR is the eddy-diffusivity portion of it. I have added the cross-reference to the description, and posted single-column validation there against GABLS1, a conventionally neutral boundary layer, and a dry convective boundary layer.

docs/src/turbulence_closures.md
```diff
@@ -0,0 +1,115 @@
+# Turbulence closures
+
+A turbulence closure supplies the subgrid fluxes of momentum and of every tracer — the transport
+carried by motions the grid does not resolve. Breeze's closures are Oceananigans
+`AbstractScalarDiffusivity`s, so they compose with the rest of the model through the same
+diffusivity interface as Oceananigans' own. A closure may be either three-dimensional, as a
+subgrid-scale (or subfilter-scale) model for large-eddy simulation; or vertical only, commonly
+referred to as a boundary-layer scheme.
```
ewquon:
```suggestion
referred to as a planetary boundary-layer scheme.
```

docs/src/turbulence_closures.md
```diff
@@ -0,0 +1,115 @@
...
+
+In addition to the [`TKEBasedTurbulenceClosure`](@ref) described here, Breeze also inherits
+[Oceananigans' own turbulence closures](https://clima.github.io/OceananigansDocumentation/stable/physics/turbulence_closures).
```
ewquon:
```suggestion
!!! tip "Other available turbulence closures"

    In addition to the [`TKEBasedTurbulenceClosure`](@ref) described here, Breeze also inherits
    [Oceananigans' own turbulence closures](https://clima.github.io/OceananigansDocumentation/stable/physics/turbulence_closures).
```

docs/src/turbulence_closures.md
```diff
@@ -0,0 +1,115 @@
...
+a wall: writing ``x`` for the ratio of a branch to the smallest one, the blend sits below that
+smallest branch by ``x^p/p``, so at ``p = 1`` an outer scale contaminates the surface layer at
+first order in height. [Mason and Thomson (1992)](@cite MasonThomson1992) match a wall scale to an
+outer scale in exactly this form and recommend ``p = 2``.
+
+Two structural departures from MYNN, both toward smoothness: ``ℓᵇ`` is a branch rather than the
+hard realizability clip ``ℓ/q ≤ 1/N``, and ``N²`` enters through a smooth positive part, so the
+stable limit engages continuously instead of switching on. Two further pieces of MYNN are not yet
+implemented — the stability correction on the surface branch (their Eq. 53) and the convective
+enhancement of ``ℓᵇ`` (their Eq. 55).
```
ewquon:
```suggestion
Two structural departures from MYNN. ``N²`` enters ``ℓᵇ`` through a smooth positive part, so
the stable limit engages continuously rather than switching on at ``∂Θᵥ/∂z > 0`` as in their
Eq. 55. And their separate realizability restriction ``ℓ/q ≤ 1/N`` (their Eq. 56) is dropped:
every blending rule here is bounded by its smallest branch, so ``ℓ ≤ ℓᵇ = Cᵇ q/N`` already
satisfies it whenever ``Cᵇ ≤ 1``, which holds for both the default 0.53 and MYNN's own 1.

MYNN's two remaining pieces are implemented but off by default: the stability correction on
the surface branch (their Eq. 53) is [`SurfaceLayerLengthScale`](@ref), and the convective
enhancement of ``ℓᵇ`` (their Eq. 55) is the ``Cᶜᵇ`` coefficient of
[`BuoyancyLengthScale`](@ref), zero unless set. [`NakanishiNiinoLengthScale`](@ref) assembles
both with MYNN's own coefficients under their harmonic blend.
```

docs/src/turbulence_closures.md
```diff
@@ -0,0 +1,115 @@
...
+layer with ``ℓ = κ z`` the closure collapses onto Prandtl's mixing-length model, but with an
+effective length ``Cˢ κ z`` rather than ``κ z``, where
+
+```math
+Cˢ ≡ Cᴷ / (Cμ)^{1/4}
+```
+
+is the stress coefficient. The model's own velocity gradient is then ``u_\star / (Cˢ κ z)``, so
+regressing ``U/u_\star`` on ``\ln(z/ℓʳ)`` over the constant-stress layer  returns ``Cˢ κ`` rather than
+``κ``. The log law is recovered exactly on ``Cˢ = 1`` or, equivalently, on the locus ``Cμ = Cᴷ⁴``.
```
ewquon:
```suggestion
``κ``. The log law is recovered exactly on ``Cˢ = 1`` or, equivalently, on the locus ``Cμ = (Cᴷ)⁴``.
```

src/AtmosphereModels/implicit_vertical_advection.jl
```diff
@@ -263,6 +277,121 @@ end
...
 
+#####
+##### Mass-flux weighting of the vertically-implicit diffusion coefficients (z-Center fields)
+#####
+##### The prognostic is `q = ρ c`, but the flux the explicit path forms is `ρ κ ∂z c`
+##### (`Jᶜz`/`𝒯_uz` in src/TurbulenceClosures/TurbulenceClosures.jl weight the kinematic flux by
+##### `ℑz(ρ)`). Backward Euler on `∂t q = ∂z(ρ κ ∂z (q/ρ))` gives row k
+#####
+#####   du(k)  = - Δt κₖ₊₁ (ρᶠₖ₊₁ / ρᶜₖ₊₁) Δz⁻¹ᶜₖ Δz⁻¹ᶠₖ₊₁
+#####   dl(k′) = - Δt κₖ    (ρᶠₖ   / ρᶜₖ₋₁) Δz⁻¹ᶜₖ Δz⁻¹ᶠₖ,          k = k′ + 1
```
ewquon:
```suggestion
#####   dl(k′) = - Δt κₖ   (ρᶠₖ   / ρᶜₖ₋₁) Δz⁻¹ᶜₖ Δz⁻¹ᶠₖ,          k = k′ + 1
```

src/AtmosphereModels/update_atmosphere_model_state.jl
```diff
@@ -214,8 +214,16 @@ function compute_auxiliary_variables!(model)
...
 
-    # Compute diffusivities
+    # Compute diffusivities, then fill their halos. A diffusivity stored at (Center, Center, Face)
+    # is interpolated *horizontally* to reach the momentum faces — `νᶠᶜᶠ` averages `ν[i-1, j, k]`
+    # and `ν[i, j, k]` — so an unfilled halo halves the vertical viscous stress in the first column
+    # and at every rank boundary. Every Oceananigans model fills these halos here
+    # (`update_nonhydrostatic_model_state.jl:50`). The omission was latent because no test ever
+    # inspected a halo value of `νₑ` — not because it was harmless: `SmagorinskyLilly`,
+    # `DynamicSmagorinsky` and `AnisotropicMinimumDissipation` all produce a spatially varying `νₑ`
+    # and were all affected. A no-op when `closure_fields === nothing`.
```
ewquon:
```suggestion
    # and `ν[i, j, k]` — an unfilled halo halves the vertical viscous stress in the first column
    # and at every rank boundary. A no-op when `closure_fields === nothing`.
    # Every Oceananigans model fills these halos here (`update_nonhydrostatic_model_state.jl:50`). 
```

src/TurbulenceClosures/blended_mixing_length.jl
```diff
@@ -0,0 +1,536 @@
...
+
+Fields
+======
+
+$(TYPEDFIELDS)
+"""
+Base.@kwdef struct GeometricLengthScale{FT} <: AbstractLengthScale
+    "von Kármán constant, setting the slope ``ℓᵍ = κ(z + ℓʳ)``"
+    κ :: FT = 0.4
+    "roughness length ``ℓʳ``, which keeps ``ℓᵍ`` finite at the surface, m"
```
ewquon:
```suggestion
    "roughness length ``ℓʳ``, which keeps ``ℓᵍ > 0`` at the surface, m"
```

src/TurbulenceClosures/blended_mixing_length.jl
```diff
@@ -0,0 +1,536 @@
...
+    κ :: FT = 0.4
+    "roughness length ``ℓʳ``, m"
+    ℓʳ :: FT = 0.1
+    "strongly-stable divisor, MYNN's 3.7"
+    Cⁿ :: FT = 3.7
+    "weakly-stable slope, MYNN's 2.7"
+    Cˢ :: FT = 2.7
+    "unstable coefficient, MYNN's 100"
+    Cᶜ :: FT = 100
+    "unstable exponent, MYNN's 0.2"
```
ewquon:
```suggestion
    "strongly-stable divisor"
    Cⁿ :: FT = 3.7
    "weakly-stable slope"
    Cˢ :: FT = 2.7
    "unstable coefficient"
    Cᶜ :: FT = 100
    "unstable exponent"
```

src/TurbulenceClosures/blended_mixing_length.jl
```diff
@@ -0,0 +1,536 @@
...
+    nᶜ :: FT = 0.2
+    "floor on ``ζ``, bounding the unstable branch. Nakanishi (2001)'s LES spans ``ζ ∈ [-3.13, 0.44]``
+     (his Table I and Fig. 3), and the unstable branch is the only piece of Eq. 53 that diverges
+     rather than saturating — at zero shear it would otherwise reach ``ζ ~ -10¹⁴``. Because
+     ``ℓˢ ∝ |ζ|^{nᶜ}``, doubling the ceiling would take a 32-fold change here, so the exact value is
+     second order; ``-4`` sits just beyond the data as MYNN's ``ζ = 1`` sits beyond theirs"
```
ewquon:
```suggestion
     rather than saturating"
```

src/TurbulenceClosures/blended_mixing_length.jl
```diff
@@ -0,0 +1,536 @@
...
+#####
+##### Blends
+#####
+
+"""
+$(TYPEDEF)
+
+``ℓ = \\min(ℓ₁, ℓ₂, …)`` — the most restrictive branch wins outright. The simplest rule, and the
+only one with no free parameter, but it is ``C⁰``: the derivative with respect to a branch that is
+not currently smallest is exactly zero, which matters for gradient-based calibration.
```
ewquon:
```suggestion
``ℓ = \\min(ℓ₁, ℓ₂, …)`` — the simplest rule: the most restrictive branch wins outright.
```

src/TurbulenceClosures/blended_mixing_length.jl
```diff
@@ -0,0 +1,536 @@
...
+where one scale is meant to dominate — the surface layer, say — the others still contaminate it
+proportionally. [`PowerBlend`](@ref) is the same rule with that order as a parameter.
+"""
+struct HarmonicBlend <: AbstractLengthScaleBlend end
+
+"""
+$(TYPEDEF)
+
+``ℓ^{-p} = ℓ₁^{-p} + ℓ₂^{-p} + …``, the ``p``-norm blend of
+[Mason and Thomson (1992)](@cite MasonThomson1992), who match a wall scale ``κ(z + z₀)`` to an
```
ewquon:
```suggestion
[Mason and Thomson (1992)](@cite MasonThomson1992), who match a wall scale ``κ(z + ℓʳ)`` to an
```

src/TurbulenceClosures/blended_mixing_length.jl
```diff
@@ -0,0 +1,536 @@
...
+outer scale with ``n = 2``.
+
+``p = 1`` reproduces [`HarmonicBlend`](@ref) and ``p → ∞`` approaches [`MinimumBlend`](@ref).
+Writing ``x`` for the ratio of a branch to the smallest one, the blend sits below that smallest
+branch by ``x^p / p``, so larger ``p`` confines each branch's influence to the region where it is
+actually the relevant scale. ``p = 2`` is both the literature value and the cheapest: it needs one
+reciprocal square root rather than a general power.
```
ewquon:
```suggestion
``p = 2`` is both the literature value and the cheapest: it needs one reciprocal square root rather than a general power.
```

src/TurbulenceClosures/blended_mixing_length.jl
```diff
@@ -0,0 +1,536 @@
...
+```julia
+closure = TKEBasedTurbulenceClosure(mixing_length = NakanishiNiinoLengthScale(), Cq = 3)
+```
+
+The blend is not incidental. The unstable coefficient of [`SurfaceLayerLengthScale`](@ref) was
+chosen so that the *blended* length matches the LES, not the branch itself, so pairing these
+branches with a sharper rule leaves ``ℓ`` systematically long. Their companion setting is
+``C^q = 3`` (their Eq. 67), which is deliberately left to the caller rather than folded in here:
+it is a modelling choice with its own provenance, and burying it is how it stops being one.
+
```
ewquon:
```suggestion
```

docs/src/appendix/notation.md
```diff
@@ -130,7 +130,24 @@ The following table also uses a few conventions that suffuse the source code and
...
 | ``Cᵛ``                              | `Cᵛ`   |                                     | Surface vapor transfer coefficient (Dalton number)                             |
-| ``\ell``                            | `ℓ`    |                                     | Surface roughness length, m                                                    |
+| ``\ell``                            | `ℓ`    |                                     | A **length scale**, m; the superscript says which one. Bare ``ℓ`` is used locally where only one length scale is in play (the master mixing length in `TKEBasedTurbulenceClosure`, the divergence-damping scale in `CompressibleEquations`) |
+| ``\ell^g``                          | `ℓᵍ`   |                                     | Geometric branch of the mixing length, ``ℓᵍ = κ(z + ℓʳ)``                       |
+| ``\ell^t``                          | `ℓᵗ`   |                                     | Turbulence branch of the mixing length, ``ℓᵗ = Cᵗ ∫qz\,dz / ∫q\,dz``            |
+| ``\ell^b``                          | `ℓᵇ`   |                                     | Buoyancy branch of the mixing length, ``ℓᵇ = Cᵇ q / N``                         |
```
glwagner, line 136: I propose we avoid defining the new symbol `q`, which we use for microphysical variables / mass fractions (eg `q^v` for specific humidity). Instead we can use `sqrt(e)` here

docs/src/turbulence_closures.md
```diff
@@ -0,0 +1,126 @@
...
+### The mixing length
+
+The length scale ``ℓ`` is supplied by a dispatched component, by default a
+[`BlendedMixingLength`](@ref): a tuple of independent length-scale branches reduced to one master
+length by a blending rule. The three default branches follow MYNN
+([Nakanishi and Niino 2009](@cite NakanishiNiino2009), their Eqs. 53–55), with the surface
+branch reduced to its neutral limit,
+
+```math
+ℓᵍ = κ (z + ℓʳ), \qquad ℓᵗ = Cᵗ \frac{∫ q z \, dz}{∫ q \, dz}, \qquad ℓᵇ = Cᵇ q / N,
```
glwagner, line 47: does it matter to add `ℓʳ` ? I know at least over the ocean `ℓʳ` is something like 1e-3. The big simplification of omitting it is that the turbulence closure becomes decoupled from the flux implementation. We may want to explore novel / learned forms for `ℓʳ`, so requiring that it is passed into the turbulence closure presents a significant complication

docs/src/turbulence_closures.md
```diff
@@ -0,0 +1,126 @@
...
+Eliminating ``ℓ`` between the viscosity and the dissipation gives the ``k``–``ε`` eddy-viscosity
+relation, whose coefficient is the product of the other two:
+
+```math
+K^u = Cᴷ ℓ \sqrt{e} = (Cᴷ Cᵋ) \frac{e²}{ε} ≡ Cμ \frac{e²}{ε}, \qquad Cμ ≡ Cᴷ Cᵋ.
+```
+
+So the closure stores ``Cᴷ`` and ``Cμ`` rather than ``Cᴷ`` and ``Cᵋ``, and the dissipation
+coefficient ``Cᵋ = Cμ/Cᴷ``, the surface turbulence level and the stress coefficient all derive from
+the stored pair.
```
glwagner, line 93: I propose avoiding mention of k-epsilon here. While I understand the closure can be derived from a 2-equation relation, I think it is more intellectually honest to simply propose the diffusivity as an empirical closure rather than claim there is a deeper theoretical reason for using this form. This gives us wider latitude for subsequent modifications, such as stability functions, non-local terms, learned mixing lengths, etc.

docs/src/turbulence_closures.md
```diff
@@ -0,0 +1,126 @@
...
+### Coefficients
+
+Eliminating ``ℓ`` between the viscosity and the dissipation gives the ``k``–``ε`` eddy-viscosity
+relation, whose coefficient is the product of the other two:
+
+```math
+K^u = Cᴷ ℓ \sqrt{e} = (Cᴷ Cᵋ) \frac{e²}{ε} ≡ Cμ \frac{e²}{ε}, \qquad Cμ ≡ Cᴷ Cᵋ.
+```
+
+So the closure stores ``Cᴷ`` and ``Cμ`` rather than ``Cᴷ`` and ``Cᵋ``, and the dissipation
```
glwagner, line 91: Following CATKE, one way to think about the "coefficients" is that there need to be different mixing lengths for momentum vs tracers vs TKE. This leads to a notation:

```math
K^u = \ell^u \sqrt{e}
```

where ``\ell^u`` may be expressed in terms of a "primary scale" ``\ell``:

```math
\ell^u = S^u \ell
```

where ``S^u`` is a "stability function" for ``u``. Then the form above uses a constant stability function,

```math
S^u = Cᴷ
```

I propose we write the closure this way to avoid confusion later on about whether what we write is compatible or not with MY, and also to promote a more straightforward empirical interpretation of the closure.

glwagner: We can also avoid defining ``C \mu``

src/AtmosphereModels/implicit_vertical_advection.jl
```diff
@@ -263,6 +277,121 @@ end
...
+# Breeze-owned wrapper routing a z-Center prognostic's implicit solve to the mass-flux-weighted
+# coefficients below. Like `VerticalMomentumImplicitAdvection`, wrapping the scheme puts a
+# Breeze-owned type in the `get_coefficient` signature, so these methods are neither type piracy
+# nor ambiguous with Oceananigans' own `AIVA` and fallback methods.
+struct MassWeightedImplicitDiffusion{A}
+    scheme :: A
+end
+
+Adapt.adapt_structure(to, a::MassWeightedImplicitDiffusion) =
+    MassWeightedImplicitDiffusion(adapt(to, a.scheme))
```
glwagner, line 318: I think these function belong in a new file, rather than in the "implicit vertical advection" file.

src/AtmosphereModels/update_atmosphere_model_state.jl
```diff
@@ -214,8 +214,13 @@ function compute_auxiliary_variables!(model)
     # Dispatch on dynamics type (computes pressure for compressible dynamics)
     compute_auxiliary_dynamics_variables!(model)
 
-    # Compute diffusivities
+    # Compute diffusivities, then fill their halos. A diffusivity stored at (Center, Center, Face)
+    # is interpolated *horizontally* to reach the momentum faces — `νᶠᶜᶠ` averages `ν[i-1, j, k]`
+    # and `ν[i, j, k]` — an unfilled halo halves the vertical viscous stress in the first column
+    # and at every rank boundary. A no-op when `closure_fields === nothing`.
+    # Every Oceananigans model fills these halos here (`update_nonhydrostatic_model_state.jl:50`).
```
glwagner, line 221: I think this comment can be omitted.

src/TimeSteppers/ssp_runge_kutta_3.jl
```diff
@@ -119,24 +126,30 @@ function ssp_rk3_substep!(model, Δt, α)
...
-        # - indices 1, 2, 3 are momentum (ρu, ρv, ρw)
-        # - indices 4+ are scalars (ρθ/ρe, ρqᵗ, microphysics, tracers)
-        # For scalars, we use Val(i - 3) to get Val(1), Val(2), etc.
-        field_index = Val(i - 3)
+        # Dynamics-specific prognostics such as compressible dry density are advanced explicitly,
+        # but are not closure scalars and must not enter the implicit diffusion solve.
+        i <= n_dynamics && continue
+
+        momentum = names[i] === :ρu || names[i] === :ρv || names[i] === :ρw
+        field_index = momentum ? nothing : closure_scalar_index(model, i)
```
glwagner, line 139: don't we need implicit diffusion for momentum and potential temperature?

I think there is an assumption embedded in `i <= n_dynamics` but its hard to understand. Can this be made more explicit? Something like `skip_vertical_diffusion(field_index, closure)` ?

src/TurbulenceClosures/blended_mixing_length.jl
```diff
@@ -0,0 +1,533 @@
...
+    closure_ij = getclosure(i, j, closure)
+    Cᵗ = turbulence_length_coefficient(closure_ij.mixing_length)
+    eᵐⁱⁿ = closure_ij.eᵐⁱⁿ
+    qᵐⁱⁿ = sqrt(2 * eᵐⁱⁿ)
+
+    FT = eltype(grid)
+
+    if isnothing(Cᵗ)
+        @inbounds ℓᵗ[i, j, 1] = FT(Inf)
```
glwagner, line 510: why do we need this branch?

src/TurbulenceClosures/blended_mixing_length.jl
```diff
@@ -0,0 +1,533 @@
...
+    else
+        ∫qz = zero(grid)
+        ∫q  = zero(grid)
+
+        for k in 1:size(grid, 3)
+            eᵢ = @inbounds e[i, j, k]
+            q  = sqrt(2 * max(eᵐⁱⁿ, eᵢ))
```
glwagner, line 517: the `sqrt(2)` falls out of the integral and doesn't need to be included.

src/TurbulenceClosures/blended_mixing_length.jl
```diff
@@ -0,0 +1,533 @@
...
+`RectilinearGrid`, floored and unfloored give identical `ℓ` and `Kᵘ`.
+
+What the unfloored form buys is immersed boundaries, where `z_bottom` varies horizontally and the
+floor can bind at *interior* faces that no mask reaches. Those are out of scope for now, so treat
+this as keeping the door open, not as fixing a present defect.
+
+These are *new functions in this module*, not methods added to Oceananigans'. Extending the upstream
+ones is not possible without type piracy — their signature `(i, j, k, grid)` contains no
+Breeze-owned type — and would silently change the mixing length of every Oceananigans closure that
+calls them, CATKE included.
```
glwagner, line 439: huge docstring for a minor function. I don't really understand why `z_node - z_bottom` doesn't work generically. It seems simple enough...

src/TurbulenceClosures/blended_mixing_length.jl
```diff
@@ -0,0 +1,533 @@
...
+
+`mixing_lengthᶜᶜᶠ` at cell centers, for the dissipation ``ε = Cᵋ e^{3/2}/ℓ``, which lives with ``e``
+at centers.
+
+Evaluating the geometric branch here rather than interpolating ``ℓ`` down from the faces matters in
+the first cell: the face value at the surface is masked to zero, so an interpolated ``ℓ`` would be
+halved (or, ignoring the masked face, doubled) exactly where the log-layer balance is judged.
+"""
+@inline function mixing_lengthᶜᶜᶜ(i, j, k, grid, ml::BlendedMixingLength, q, N², state)
+    ℓs = map(branch -> length_scaleᶜᶜᶜ(i, j, k, grid, branch, q, N², state), ml.branches)
```
glwagner, line 411: invoking `map` like this may be problematic for compile time and reactant differentiation

src/TurbulenceClosures/blended_mixing_length.jl
```diff
@@ -0,0 +1,533 @@
...
+    Jᵇ = @inbounds state.Jᵇ[i, j, 1]
+    ℓᵍ = ℓ.κ * (z + ℓ.ℓʳ)
+
+    # ζ = z/L with L = -u★³/(κ Jᵇ), formed as a product so that u★ → 0 — free convection, where the
+    # shear scale is irrelevant rather than zero — sends ζ → -∞ and the branch to infinity.
+    u★³ = u★² * sqrt(u★²)
+    ζ = -ℓ.κ * z * Jᵇ / max(u★³, eps(typeof(ℓᵍ)))
+
+    ζ⁺ = max(ζ, zero(ζ))
+    ζ⁻ = clamp(ζ, ℓ.ζᵐⁱⁿ, zero(ζ))
```
glwagner, line 376:
```suggestion
    ζ⁺ = max(ζ, 0)
    ζ⁻ = clamp(ζ, ℓ.ζᵐⁱⁿ, 0)
```

src/TurbulenceClosures/blended_mixing_length.jl
```diff
@@ -0,0 +1,533 @@
...
+"""
+$(TYPEDSIGNATURES)
+
+MYNN Eq. 53. The three pieces are formed unconditionally and selected afterwards, because `ifelse`
+evaluates both arms: the unstable form raises ``1 - Cᶜ ζ`` to a fractional power, which is a domain
+error for ``ζ > 1/Cᶜ``, so its argument is clipped to the side of ``ζ = 0`` on which it applies.
+"""
+@inline function surface_layer_length(ℓ::SurfaceLayerLengthScale, z, i, j, state)
+    u★² = @inbounds state.u★²[i, j, 1]
+    Jᵇ = @inbounds state.Jᵇ[i, j, 1]
```
glwagner, line 367: as we have seen with CATKE, it can be problematic to use the surface flux directly in the length scale (which can fluctuate very rapidly). For CATKE we smooth the buoyancy flux using some timescale. For this PR though, I think it would be better (less likely to introduce problems in the future) if we just omit this contribution completely. Perhaps we do not need this surface layer mixing length and can rely on stability functions instead? They play an overlapping role.

src/TurbulenceClosures/blended_mixing_length.jl
```diff
@@ -0,0 +1,533 @@
...
+``S_M(G_M, G_H)`` and ``S_H(G_M, G_H)`` of the Mellor–Yamada hierarchy are replaced in
+[`TKEBasedTurbulenceClosure`](@ref) by a constant ``Cᴷ`` and a Richardson-dependent Prandtl number,
+which makes it a ``k``–``ℓ`` closure borrowing their parameterizations rather than a level-2.5
+model. What is genuinely theirs is the length scale, and that is what this constructor names.
+
+``Cᵇ`` and ``Cᶜᵇ`` are given explicitly because they are where MYNN differs from this package's own
+branch defaults — Deardorff's ``Cᵇ = 0.53`` and no convective enhancement. For anything else, build
+a [`BlendedMixingLength`](@ref) directly.
+"""
+NakanishiNiinoLengthScale(; ℓʳ = 0.1, Cᵇ = 1, Cᶜᵇ = 5) =
```
glwagner, line 289: isn't a constant roughness length inconsistent with a general flux application? eg in NumericalEarth the roughness is state dependent.

src/TurbulenceClosures/blended_mixing_length.jl
```diff
@@ -0,0 +1,533 @@
...
+Fields
+======
+
+$(TYPEDFIELDS)
+"""
+struct BlendedMixingLength{B, L}
+    "the rule reducing the branches to one length"
+    blend :: B
+    "tuple of length-scale branches"
+    branches :: L
```
glwagner, line 263: I think it may be simpler to hard code the lengths we can consider, so that we don't have to unroll the `branches` tuple at compile time. I'm also worried this could make reactant compilation a headache. Maybe we can consider this kind of generic abstraction if we find that it is very important to consider an arbitrary number of length scales, in the future?

src/TurbulenceClosures/blended_mixing_length.jl
```diff
@@ -0,0 +1,533 @@
...
+
+"""
+$(TYPEDSIGNATURES)
+
+Smooth positive part of `x` with transition scale `δ > 0`: `(x + √(x² + δ²)) / 2`. Agrees with
+`max(x, 0)` to within `δ/2` and is differentiable everywhere, unlike `max`.
+"""
+@inline smooth_positive(x, δ) = (x + sqrt(x^2 + δ^2)) / 2
+
+@inline length_scaleᶜᶜᶠ(i, j, k, grid, ℓ::GeometricLengthScale, q, N², state) =
```
glwagner, line 328: perhaps `HeightAboveSurface` instead of `GeometricLengthScale`? I was confused at first what "geometry" meant (although I understand now this is meant to pertain to the "geometry of a turbulence eddy"). I think perhaps a more direct / prosaic name would be more future proof than imposing an eddy-size-based interpretation (which may not be entirely accurate)?

src/TurbulenceClosures/blended_mixing_length.jl
```diff
@@ -0,0 +1,533 @@
...
+    Cˢ :: FT = 2.7
+    "unstable coefficient"
+    Cᶜ :: FT = 100
+    "unstable exponent"
+    nᶜ :: FT = 0.2
+    "floor on ``ζ``, bounding the unstable branch. Nakanishi (2001)'s LES spans ``ζ ∈ [-3.13, 0.44]``
+     (his Table I and Fig. 3), and the unstable branch is the only piece of Eq. 53 that diverges
+     rather than saturating"
+    ζᵐⁱⁿ :: FT = -4
+end
```
glwagner, line 159: If we omit this length scale, we can avoid difficulties associated with inconsistency between the flux formulation and the closure formulation

src/TurbulenceClosures/blended_mixing_length.jl
```diff
@@ -0,0 +1,533 @@
...
+with MYNN's correction applied, and the two agree exactly in neutral conditions.
+
+Fields
+======
+
+$(TYPEDFIELDS)
+"""
+Base.@kwdef struct GeometricLengthScale{FT} <: AbstractLengthScale
+    "von Kármán constant, setting the slope ``ℓᵍ = κ(z + ℓʳ)``"
+    κ :: FT = 0.4
```
glwagner, line 52: I think this can be omitted. With a stability-function formulation, one of the length scales does not need a parameter (otherwise the number of parameters is overspecified)

src/TurbulenceClosures/TurbulenceClosures.jl
```diff
@@ -1,6 +1,32 @@
...
+       NakanishiNiinoLengthScale,
+       GeometricLengthScale,
+       SurfaceLayerLengthScale,
+       TurbulenceLengthScale,
+       BuoyancyLengthScale,
+       MinimumBlend,
+       HarmonicBlend,
+       PowerBlend,
+       diffusivity_coefficient,
+       dissipation_coefficient,
```
glwagner, line 14: I propose introducing a dissipation length scale with Ri-dependent stability function (like CATKE). In this case, we do not have a single "dissipation coefficient". There is an effective coefficient but it is state-dependent.

src/TurbulenceClosures/TurbulenceClosures.jl
```diff
@@ -1,6 +1,32 @@
...
+       BlendedMixingLength,
+       NakanishiNiinoLengthScale,
+       GeometricLengthScale,
+       SurfaceLayerLengthScale,
+       TurbulenceLengthScale,
+       BuoyancyLengthScale,
+       MinimumBlend,
+       HarmonicBlend,
+       PowerBlend,
+       diffusivity_coefficient,
```
glwagner, line 13: same comment here as below for the dissipation coefficient (I think)

docs/src/appendix/notation.md
```diff
@@ -130,7 +130,24 @@ The following table also uses a few conventions that suffuse the source code and
...
+| ``\rho e``                          | `ρtke` |                                     | Prognostic TKE density, ``ρ e``. Named `ρtke` because `ρe` is the static-energy density |
+| ``\mathrm{Pr}``                     | `Pr`   |                                     | Turbulent Prandtl number, ``\mathrm{Pr} = K^u/K^c``; ``\mathrm{Pr₀}`` is its neutral value |
+| ``Ri``                              | `Ri`   |                                     | Gradient Richardson number, ``Ri = N²/S²``                                      |
+| ``S``                               | `S`    |                                     | Vertical shear magnitude, ``S² = (∂_z u)² + (∂_z v)²``, s⁻¹                     |
+| ``N^2``                             | `N²`   |                                     | Squared Brunt–Väisälä frequency, s⁻². (Bare ``N`` is the acoustic substep count, below) |
+| ``u_\star``                         | `u★`   |                                     | Friction velocity, m s⁻¹                                                       |
```
glwagner, line 150: ideally the friction velocity does not enter into the closure. Note that it is generally the case that ``e \sim u_\star^2``, so that ``e`` at the surface can be used as a proxy / stand-in for the friction velocity in a parameterization context.

src/AtmosphereModels/update_atmosphere_model_state.jl
ewquon:
```suggestion
    # Compute diffusivities
```

docs/src/turbulence_closures.md
ewquon: It's a form I've seen in some papers and I agree that it's a trivial amount. It's there to keep `ℓᵍ` bounded but that's not strictly necessary. Will remove

src/TurbulenceClosures/blended_mixing_length.jl
ewquon: This is just there because of the decision I made regarding `ℓᵍ` and agree that it's not correct in general. Will remove

src/TurbulenceClosures/blended_mixing_length.jl
ewquon: I don't think "geometry of a turbulent eddy" is strictly correct either. I interpret the "geometric" qualifier to indicate that it's a physical distance as opposed to a derived length scale that is representative of some physical phenomenon. I'm supportive of calling it `WallDistance`, `HeightAboveGroundLevel` or `HeightAboveSurface`. Wall distance recalls other RANS turbulence modeling terminology, but what we have here is a 1-D approximation to that.

glwagner:
Update on where this is going, after sitting down with the CATKE sources and this PR.

**Direction.** Align the closure with CATKE's design, starting with constant stability functions: `Kᵘ = Sᵘ ℓ √e`, `Kᶜ = Sᶜ ℓ √e`, `Kᵉ = Sᵉ ℓ √e`, `ε = Sᴰ e^{3/2}/ℓ` (i.e. `ℓᴰ = ℓ/Sᴰ`), one primary length `ℓ = min(z, Cᴺ √e/N)` with the height above the surface carrying no coefficient, and the four `S` held as constants `Cᵘ, Cᶜ, Cᵉ, Cᴰ` in `ConstantStabilityFunctions`, dispatched so that CATKE's Richardson-number-dependent functions are a new type and four methods later. Everything that read the surface fluxes or a roughness length inside the closure goes — the surface-layer length, `Jᵇ`, the `u★²` field and surface TKE floor — as do `q = √(2e)`, `Cμ` and its derived coefficients, the branch/blend machinery, the turbulence length `ℓᵗ`, and the Richardson-dependent Prandtl number (`Pr = Cᵘ/Cᶜ` for now). The closure is presented as empirical rather than derived from k–ε.

**Pushed here**: c37aa300 (on top of a merge of `main`, which also fixed the red 1.11 GPU job — it was the Reactant pin). Rewritten closure, docs page, notation rows, tests and example. Defaults are the MYNN coefficients with κ absorbed (`Cᵘ = 0.196, Cᶜ = 0.265, Cᵉ = 0.392, Cᴰ = 0.295, Cᴺ = 0.76`; κ = 0.40, e/u★² = 4.2, Pr = 0.74, Ri† = 0.25) — placeholders for calibration. The TKE tracer is now `ρe`. The local TKE terms follow CATKE's split (dissipation and the negative buoyancy flux implicit in the rate, production and the positive buoyancy flux explicit, negative TKE damped on a time scale, no clips); the surface TKE is zero-flux for now.

**Split out**: #925 carries the #870/#871 fixes, with the implicit-solve routing keyed on the prognostic's name rather than its position (the `Val(i - 3)` question above) and `DrySurface` promoted from the example into `Thermodynamics`; #926 renames the static-energy symbol `e → s` so that `ρe` is free for TKE. Both should merge before this one; I'll merge `main` here again afterwards.

**Next**: move dissipation and the negative buoyancy flux into the per-stage tridiagonal solve through `implicit_linear_coefficient`, with the sources in the stage tendency (CATKE's structure); rerun the GABLS1 / CNBL / CBL example and refresh the #847 table. Follow-ups in their own PRs: Richardson-dependent stability functions, the convective length scale and a CATKE-style surface TKE flux, a `non_local_flux` slot, calibration.

I'll note on each review thread what happened to it.

docs/src/appendix/notation.md
glwagner: K it is — thanks.

docs/src/appendix/notation.md
glwagner: `√e` throughout now; the `q` notation row is gone (c37aa300).

docs/src/turbulence_closures.md
glwagner: Removed from the length scale: ℓ = min(z, Cᴺ √e/N), no roughness length and no κ (c37aa300).

docs/src/turbulence_closures.md
glwagner: Rewritten as an empirical closure in CATKE's form; no k–ε, no Cμ (c37aa300).

docs/src/turbulence_closures.md
glwagner: Adopted: Kᵘ = Sᵘ ℓ √e, Kᶜ = Sᶜ ℓ √e, Kᵉ = Sᵉ ℓ √e, with the constants in `ConstantStabilityFunctions(Cᵘ, Cᶜ, Cᵉ, Cᴰ)`; a Richardson-dependent variant is a new type plus four methods (c37aa300).

glwagner: Cμ and its derived coefficients are gone (c37aa300).

src/TimeSteppers/ssp_runge_kutta_3.jl
glwagner: Yes — momentum and θ do take the implicit solve; only the dynamics-specific prognostics (ρᵈ, or ρ for the kinematic driver) sit it out. The predicates are now keyed on the prognostic's name (`closure_scalar_index`, `skip_vertical_diffusion`, `acoustic_prognostic_names`) and shared by both steppers — split out into #925.

src/TurbulenceClosures/blended_mixing_length.jl
glwagner: Gone together with the turbulence length (c37aa300).

src/TurbulenceClosures/blended_mixing_length.jl
glwagner: Gone together with `q` (c37aa300).

src/TurbulenceClosures/blended_mixing_length.jl
glwagner: It was `znode − z_bottom`; upstream's floor binds only at the masked k = 1 face on these grids, so the closure now imports Oceananigans' `height_above_bottomᶜᶜᶠ/ᶜᶜᶜ` instead of defining its own (c37aa300).

src/TurbulenceClosures/blended_mixing_length.jl
glwagner: No tuple and no `map` any more: ℓ = min(z, Cᴺ √e/N) is hard-coded (c37aa300).

src/TurbulenceClosures/blended_mixing_length.jl
glwagner: Removed, together with `Jᵇ` and the convective enhancement; the stability functions take that role (c37aa300).

src/TurbulenceClosures/blended_mixing_length.jl
glwagner: `NakanishiNiinoLengthScale` is gone (c37aa300).

src/TurbulenceClosures/blended_mixing_length.jl
glwagner: Done: one primary length ℓ = min(z, Cᴺ √e/N), no branches (c37aa300).

src/TurbulenceClosures/blended_mixing_length.jl
glwagner: The type itself is gone now that ℓ is hard-coded (c37aa300).

src/TurbulenceClosures/blended_mixing_length.jl
glwagner: Removed (c37aa300).

src/TurbulenceClosures/blended_mixing_length.jl
glwagner: Removed: the height above the surface carries no coefficient, and κ = (Cᵘ³/Cᴰ)^{1/4} = 0.40 is derived from the stability functions (c37aa300).

src/TurbulenceClosures/TurbulenceClosures.jl
glwagner: ℓᴰ = ℓ/Sᴰ with a constant Cᴰ for now; the Richardson-dependent Sᴰ arrives with the other stability functions (c37aa300).

src/TurbulenceClosures/TurbulenceClosures.jl
glwagner: Gone (c37aa300).

docs/src/appendix/notation.md
glwagner: u★ no longer enters the closure: the surface floor, the `u★²` field and the surface-layer length are removed, and the surface TKE is zero-flux for now; a CATKE-style flux boundary condition is the follow-up (c37aa300).

src/TurbulenceClosures/tke_based_turbulence_closure.jl
```diff
@@ -0,0 +1,568 @@
...
+scale of each diffusivity.
+
+Fields
+======
+
+$(TYPEDFIELDS)
+"""
+Base.@kwdef struct TKEMixingLength{FT}
+    "coefficient of the stratification length ``ℓᴺ = Cᴺ \\sqrt{e} / N``; the default is
+     Deardorff's ``0.76 \\sqrt{e} / N`` ([Deardorff 1980](@cite Deardorff1980))"
```
glwagner: let's remove this in favor of an in-line comment

src/TurbulenceClosures/tke_based_turbulence_closure.jl
```diff
@@ -0,0 +1,568 @@
...
+    "upper bound on ``Kᵘ``, m² s⁻¹"
+    maximum_viscosity :: FT
+    "upper bound on ``Kᶜ``, m² s⁻¹"
+    maximum_tracer_diffusivity :: FT
+    "upper bound on ``Kᵉ``, m² s⁻¹"
+    maximum_tke_diffusivity :: FT
+    "floor on ``e`` inside ``\\sqrt{e}``, m² s⁻²"
+    minimum_tke :: FT
+    "time scale on which negative ``e`` is damped, s"
+    negative_tke_damping_time_scale :: FT
```
glwagner: let's use in-line comments here instead of strings

src/TurbulenceClosures/tke_based_turbulence_closure.jl
```diff
@@ -0,0 +1,568 @@
...
+"""
+function TKEBasedTurbulenceClosure(time_discretization::TD = VerticallyImplicitTimeDiscretization(),
+                                   FT = Oceananigans.defaults.FloatType;
+                                   mixing_length = TKEMixingLength(),
+                                   stability_functions = ConstantStabilityFunctions(),
+                                   maximum_viscosity = Inf,
+                                   maximum_tracer_diffusivity = Inf,
+                                   maximum_tke_diffusivity = Inf,
+                                   minimum_tke = 1e-6,
+                                   negative_tke_damping_time_scale = 1minute) where TD
```
glwagner, line 183: this time-scale is taken from CATKE, but a shorter time-scale may be more appropriate for Breeze.

src/TurbulenceClosures/tke_based_turbulence_closure.jl
```diff
@@ -0,0 +1,568 @@
...
+    Kᵉ :: K
+    "the primary mixing length ``ℓ``, at (Center, Center, Face); a diagnostic"
+    ℓ :: K
+    "the linear implicit coefficient of the TKE equation, ``∂_t e = Lᵉ e + ⋯``, at (Center, Center, Center):
+     the dissipation rate, the negative part of the buoyancy flux and the damping of negative TKE"
+    Le :: L
+    "per-tracer diffusivity lookup, indexed by the tracer's position among the closure scalars"
+    tupled_tracer_diffusivities :: KC
+    "per-tracer implicit linear coefficient lookup: `Le` for the TKE tracer, zero for every other"
+    tupled_implicit_linear_coefficients :: LC
```
glwagner: as above let's convert this to in-line comments

src/TurbulenceClosures/tke_based_turbulence_closure.jl
```diff
@@ -0,0 +1,568 @@
...
+                     adapt(to, fields.ℓ),
+                     adapt(to, fields.Le),
+                     adapt(to, fields.tupled_tracer_diffusivities),
+                     adapt(to, fields.tupled_implicit_linear_coefficients))
+
+BoundaryConditions.fill_halo_regions!(fields::TKEClosureFields, args...; kw...) =
+    fill_halo_regions!((fields.Kᵘ, fields.Kᶜ, fields.Kᵉ, fields.ℓ), args...; kw...)
+
+function Oceananigans.TurbulenceClosures.build_closure_fields(grid, clock, tracer_names, bcs,
+                                                      closure::FlavorOfTKEClosure)
```
glwagner:
```suggestion
function Oceananigans.TurbulenceClosures.build_closure_fields(grid, clock, tracer_names, bcs, closure::FlavorOfTKEClosure)
```

src/TurbulenceClosures/tke_based_turbulence_closure.jl
```diff
@@ -0,0 +1,567 @@
...
+
+function Oceananigans.TurbulenceClosures.build_closure_fields(grid, clock, tracer_names, bcs, closure::FlavorOfTKEClosure)
+    face_bcs = FieldBoundaryConditions(grid, (Center(), Center(), Face()))
+    default_bcs = (Kᵘ = face_bcs, Kᶜ = face_bcs, Kᵉ = face_bcs, ℓ = face_bcs)
+    bcs = merge(default_bcs, bcs)
+
+    Kᵘ = ZFaceField(grid, boundary_conditions=bcs.Kᵘ)
+    Kᶜ = ZFaceField(grid, boundary_conditions=bcs.Kᶜ)
+    Kᵉ = ZFaceField(grid, boundary_conditions=bcs.Kᵉ)
+    ℓ  = ZFaceField(grid, boundary_conditions=bcs.ℓ)
```
glwagner: Interesting... is it necessary to allocate memory for the mixing length? I don't think CATKE does this.

src/TurbulenceClosures/tke_based_turbulence_closure.jl
```diff
@@ -0,0 +1,567 @@
...
+    end
+end
+
+"""
+$(TYPEDSIGNATURES)
+
+Shear production ``Kᵘ S²`` at (Center, Center, Face).
+"""
+@inline shear_productionᶜᶜᶠ(i, j, k, grid, Kᵘ, u, v) =
+    @inbounds Kᵘ[i, j, k] * shearᶜᶜᶠ(i, j, k, grid, u, v)
```
glwagner:
```suggestion
@inline shear_productionᶜᶜᶠ(i, j, k, grid, Kᵘ, u, v) = @inbounds Kᵘ[i, j, k] * shearᶜᶜᶠ(i, j, k, grid, u, v)
```

src/TurbulenceClosures/tke_based_turbulence_closure.jl
```diff
@@ -0,0 +1,566 @@
...
+"""
+@inline shear_productionᶜᶜᶠ(i, j, k, grid, Kᵘ, u, v) = @inbounds Kᵘ[i, j, k] * shearᶜᶜᶠ(i, j, k, grid, u, v)
+
+"""
+$(TYPEDSIGNATURES)
+
+Buoyancy production ``-Kᶜ N²`` at (Center, Center, Face); negative in stable stratification.
+"""
+@inline buoyancy_productionᶜᶜᶠ(i, j, k, grid, Kᶜ, buoyancy, tracers) =
+    @inbounds -Kᶜ[i, j, k] * ∂z_b(i, j, k, grid, buoyancy, tracers)
```
glwagner:
```suggestion
@inline buoyancy_productionᶜᶜᶠ(i, j, k, grid, Kᶜ, buoyancy, tracers) = @inbounds -Kᶜ[i, j, k] * ∂z_b(i, j, k, grid, buoyancy, tracers)
```

glwagner (approved):

src/TurbulenceClosures/tke_based_turbulence_closure.jl
glwagner: Done in e957f964 — the field string is an inline comment now, and the Deardorff reference moved into the type's docstring.

src/TurbulenceClosures/tke_based_turbulence_closure.jl
glwagner: Converted to inline comments in e957f964 (and dropped the now-empty `Fields` section from the docstring).

src/TurbulenceClosures/tke_based_turbulence_closure.jl
glwagner: Noted with an inline comment in e957f964; kept 1 minute for now so the change stays mechanical. Happy to shorten it here if you have a value in mind — otherwise it can fall out of the calibration round.

src/TurbulenceClosures/tke_based_turbulence_closure.jl
glwagner: It isn't — removed in e957f964. Like CATKE the mixing length is now computed on the fly only; the tests and the example diagnose it from Kᵘ = Cᵘ ℓ √e where they want to look at it.

glwagner: Follow-up in bb73c4c8: the diagnostic now evaluates the closure's own `mixing_lengthᶜᶜᶠ` directly in a `KernelFunctionOperation` — the function takes the specific TKE field `e` and reconstructs √e itself, so the diffusivity kernel and scripts share one function (tests and the example construct the KFO with `e = ρe/ρ`). No wrapper API. One thing this surfaced: the unmasked surface-face ℓ is order Δz, not zero — `height_above_bottomᶜᶜᶠ` floors the wall distance at the thickness of the cell below, as CATKE intends; the stored field only read zero there through `mask_diffusivity`. The test now asserts that correctly.

src/TurbulenceClosures/tke_based_turbulence_closure.jl
```diff
@@ -0,0 +1,537 @@
...
+
+#####
+##### Mixing length: ℓ = min(z, Cᴺ √e / N), evaluated where the caller supplies √e
+#####
+
+"""
+$(TYPEDSIGNATURES)
+
+The stratification length ``ℓᴺ = Cᴺ \\sqrt{e} / N`` at (Center, Center, Face), given the
+turbulent velocity ``w★ = \\sqrt{e}`` there; infinite where ``N² ≤ 0``.
```
ewquon: I don't like this definition of `w★` -- this should be reserved for the convective velocity scale. Recommend renaming throughout, then this is good to merge.

glwagner: Agreed — the name had come along with CATKE's `turbulent_velocityᶜᶜᶜ`, and it collides with Deardorff's convective velocity scale. Gone in c17b9487, without a replacement symbol: the stratification length takes `e` and reconstructs √e itself, and the diffusivity kernel forms the one product ℓ √e that the stability functions scale into Kᵘ, Kᶜ and Kᵉ. √e is still `sqrt(max(eᵐⁱⁿ, e))`, as before.

bischtob:
NUM-217

PR #877, opened 2026-07-30
 Fix the upper sponge: damp total vertical momentum, not the acoustic perturbation

### LINKED ISSUE #874: Mountain wave example does not look right after v0.5
v0.5

Image

After v0.5, including the latest code

Image

kaiyuan-cheng:
UpperSponge damped only the acoustic perturbation (ρw)′, which is the change over a Runge-Kutta stage rather than the wave itself, so a quasi-steady gravity wave passed through the layer untouched and reflected off the rigid lid. The term has been inert since d602e3ed (#622).

Split the damping the way WRF does (Klemp, Dudhia & Hassiotis 2008): the stage-entry part -γρwᴸ now rides in the slow tendency Gˢρw, scaled by the signed Δτ so backward integration stays dissipative, while -γ(ρw)′ keeps its Crank-Nicolson treatment in the column tridiagonal. The damped variable is Cartesian ρw, not the contravariant ρw̃ the substep loop advances: since w̃ = w - slope·u, damping ρw̃ would drive ρw → slope·ρu and force the mean flow onto the coordinate surfaces.

Add GaussianRamp, a sponge profile with no hard base, and switch the mountain-wave example to it. The fix alone leaves a downstream error larger than the signal it measures; that residual is a sponge-profile effect, not a solver one, and UpperSponge previously shipped only clamped ramps.

Add regression tests for what the operator relaxes toward, which is where both this bug and the ρw̃ trap live, and correct the docstring and terrain docs, which described the perturbation-only behaviour as intended.

Closes #874


src/CompressibleEquations/time_discretizations.jl
```diff
@@ -439,21 +439,56 @@ end
...
+The tail does apply weak damping below the nominal layer, but for a steadily forced wave the
+relevant comparison is not `exp(-rate × ramp × t)`: the response equilibrates against
+advection, so the cost scales with `rate × ramp(z) / (U k)`. On the Schär case that is 10% at
+`z = H/2`, falling to 0.07% at `H/4`, and the measured rms `w` over `0.5 < z < 10` km differs
+from the clamped ramp by 0.4%. Prefer a clamped ramp when the dynamics just below the sponge
+must be provably untouched, and this one when a quiet far field matters more.
+"""
+struct GaussianRamp <: AbstractRamp end
+
+@inline (::GaussianRamp)(z, sponge_top, depth) = exp(-(z - sponge_top)^2 / depth^2)
```
glwagner, line 470: I'm worried this will be a bit expensive, especially because it is called very frequently within acoustic substepping. Is there a cheaper shape we can use involving only polynomials? Perhaps a cubic? We can also pre-evaluate this, since it is a 1D profile it should be relatively cheap to use I think.

kaiyuan-cheng: GaussianRamp absorbs the waves more effectively and your suggestion regarding pre-evaluation was a good fix and now it has been implemented.

PR #882, opened 2026-08-01
[test] Cap Reactant and use dev version of Oceananigans

giordano:
With Reactant@0.2.275 we get https://github.com/NumericalEarth/Breeze.jl/actions/runs/30322643579/job/90161546812#step:14:1280
```
Error During Test at /__w/Breeze.jl/Breeze.jl/test/reactant/lat_lon_compilation.jl:95
  Got exception outside of a @test
  CompilationError: MLIR pass pipeline "all" failed
  
  error: invalid init for iterArg: <block argument> of type 'f64' at index: 2
    note: see current operation: 
  %56:2 = "affine.for"(%34, %37) <{lowerBoundMap = affine_map<() -> (0)>, operandSegmentSizes = array<i32: 0, 0, 2>, step = 1 : index, upperBoundMap = affine_map<() -> (3)>}> ({
  ^bb0(%arg12: index, %arg13: f64, %arg14: f64):
    %63 = "affine.load"(%arg4, %arg12, %arg8, %arg9) <{map = affine_map<(d0, d1, d2) -> (d0 + 5, d1 + 5, d2 + 5)>}> : (memref<14x16x18xf64, 1>, index, index, index) -> f64
    %64 = "affine.load"(%arg2, %arg12, %arg8, %arg9) <{map = affine_map<(d0, d1, d2) -> (d0 + 5, d1 + 5, d2 + 5)>}> : (memref<14x16x18xf64, 1>, index, index, index) -> f64
    %65 = "arith.mulf"(%63, %64) <{fastmath = #arith.fastmath<none>}> {fastmathFlags = #llvm.fastmath<none>} : (f64, f64) -> f64
    %66 = "affine.load"(%arg3, %arg12, %arg8, %arg9) <{map = affine_map<(d0, d1, d2) -> (d0 + 6, d1 + 5, d2 + 5)>}> : (memref<14x16x18xf64, 1>, index, index, index) -> f64
    %67 = "affine.load"(%arg3, %arg12, %arg8, %arg9) <{map = affine_map<(d0, d1, d2) -> (d0 + 5, d1 + 5, d2 + 5)>}> : (memref<14x16x18xf64, 1>, index, index, index) -> f64
    %68 = "arith.addf"(%66, %67) <{fastmath = #arith.fastmath<none>}> {fastmathFlags = #llvm.fastmath<none>} : (f64, f64) -> f64
[...]

PR #884, opened 2026-08-02
Require new version of Reactant

giordano:
Let's see if this resolved the error in https://github.com/NumericalEarth/Breeze.jl/pull/882#issue-5038255833

***Edit***: this requires https://github.com/CliMA/Oceananigans.jl/pull/5842.


giordano:
Oh wow, the Reactant regressions are significant and probably real (but would need more runs to confirm), but we can't do much right now, we need the new version for correctness.  CC @Pangoraw.

giordano:
The performance drops happens also on `main`: https://numericalearth.github.io/BreezeBenchmarks/

Pangoraw:
i'll investigate what the root cause can be. there were changes about the loop raising that can be the source.

Pangoraw:
hmm the only things i could find about the regression would be this change within XLA https://github.com/openxla/xla/pull/45680. As I think most of the regression is only in BF16 raise=true?

giordano:
Also the Reactant backend of [CBL; Dynamics: compressible_explicit; Microphysics: 1M_MixedNonEquilibrium [Float32]](https://numericalearth.github.io/BreezeBenchmarks/#benchmark-cbl--dynamics--compressible-explicit--microphysics--1m-mixednonequilibrium--float32-) had a significant regression

PR #885, opened 2026-08-02
[examples] Increase compression rate of MP4 videos

giordano:

PR #886, opened 2026-08-03
[examples] Compress all JLD2 files

giordano:
Along the same lines of #885.  For the `dry_thermal_bubble.jl` example, the output file goes from 66 MiB to 31 MiB.  I'm hoping we get similar saving for the other examples as well.

### DIFF
--- a/examples/acoustic_wave.jl
+++ b/examples/acoustic_wave.jl
@@ -155,6 +155,7 @@ outputs = (; ρ′, u′, w, U, R, W²)
 
 simulation.output_writers[:jld2] = JLD2Writer(model, outputs; filename,
                                               schedule = TimeInterval(0.01),
+                                              jld2_kw = Dict(:compress => true),
                                               overwrite_existing = true)
 
 run!(simulation)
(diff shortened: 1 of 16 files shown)


navidcy (commented):
Happy for this! But also, these .jld2 files are not pushed or published anywhere, right?

giordano:
> Happy for this! But also, these .jld2 files are not pushed or published anywhere, right?

They aren't part of the website, no, but they are uploaded in the build artifacts (so that people can do post-processing, for example to tweak the viz, without having to rerun the simulations themselves all over again), [which is currently at 1.54 GB of size](https://github.com/NumericalEarth/Breeze.jl/actions/runs/30772140920), due to the JLD2 files.

navidcy:
Gotcha! Happy then!
Was wondering if for the purposes of the example it adds an unnecessary complication and whether a `# hide` would be appropriate? But happy either way!

giordano:
> Was wondering if for the purposes of the example it adds an unnecessary complication and whether a # hide would be appropriate?

I feel like compression is something that should be promoted, rather than hidden: it adds no noticeable overhead, but saves some storage on disk.

glwagner:
It might be better to make this default for `JLD2Writer` than pollute the examples.

For example --- is this what we recommend users do? If so it should be default. If not, we have to explain with a comment why we do something not recommended in the example.

giordano:
Compression enabled upstream in https://github.com/CliMA/Oceananigans.jl/pull/5851, this can be closed. I still want to investigate zstd compression as default filter.

PR #890, opened 2026-08-03
Include the upwelling shortwave flux in the radiative flux divergence

kaiyuan-cheng:
## Summary

`compute_radiation_flux_divergence!` built the net radiative flux from three streams
instead of four:

```julia
F_k = ℐ_lw_up[i, j, k] + ℐ_lw_dn[i, j, k] + ℐ_sw_dn[i, j, k]
```

The upwelling shortwave was missing, so the sunlight scattered back upward by air and
clouds, plus the fraction reflected by the surface, never entered the flux divergence.
Since `flux_divergence` feeds the `ρe` and `ρθ` tendencies directly, every clear-sky and
all-sky simulation ran with a biased radiative heating rate.

RRTMGP's two-stream shortwave solver already computes this stream as
`sws.flux.flux_up`; Breeze simply never read it. There was no field to store it in, so
this PR adds `upwelling_shortwave_flux` to `RadiativeTransferModel` alongside the three
existing flux fields and includes it in the net flux.

### Magnitude

In the clear-sky single-column test (8 levels over 10 km, noon at the June solstice at
45°N, surface albedo 0.1), the column-integrated radiative heating goes from
79.9 W/m² to 109.5 W/m². The old value was 27% low.

## Changes

**Source**

* `src/AtmosphereModels/radiation_interface.jl`: add `upwelling_shortwave_flux` to
  `RadiativeTransferModel`.
* `ext/BreezeRRTMGPExt/rrtmgp_shared_utilities.jl`: copy `sws.flux.flux_up` into the new
  field and include it in `_compute_radiation_flux_divergence!`.
* `ext/BreezeRRTMGPExt/{gray,clear_sky,all_sky}_radiative_transfer_model.jl`: allocate the
  field in all three constructors.
* `ext/BreezeRRTMGPExt/gray_radiative_transfer_model.jl`: write zero into the new field,
  since RRTMGP's `rte_sw_noscat!` computes no upward stream. Also zero
  `shortwave_solver.flux` once at construction: the non-scattering solver only ever writes
  `flux_dn_dir`, leaving `flux_up` and `flux_dn` as allocated-but-unwritten memory whenever
  the sun is up. Breeze never read those arrays, but leaving `undef` memory reachable
  contradicts the "no upwelling shortwave" contract for anyone checking it.

**Docs and examples**

* `examples/single_column_radiation.jl`: the four `ℐ_net_*` diagnostics had the same
  omission, so the "Net flux" panel disagreed with the model's own net flux for clear-sky,
  2×CO₂, and all-sky. Fixed, renamed `ℐ_sw_*` to `ℐ_sw_dn_*` now that an up counterpart
  exists, and plotted SW↑ as dashed lines on the shortwave panel.
* `docs/src/radiative_transfer.md`: the page listed three flux fields and stated "There is
  no diffuse shortwave or upwelling shortwave in this model". Rewritten to list all four
  fields with their sign convention, to distinguish gray optics (direct beam only,
  SW↑ ≡ 0) from clear-sky and all-sky (total downwelling, SW↑ from scattering and surface
  reflection), and to give the net-flux definition over all four streams explicitly.

**Tests**

* Assert finiteness, sign, and the surface and top-of-atmosphere behavior of the new field
  across gray, clear-sky, and all-sky.
* Assert the new field matches `RRTMGP.sw_flux_up` exactly, so a future refactor cannot
  quietly drop the copy.
* Add a column-budget identity to the clear-sky and gray tests: integrating `-∂ℐ_net/∂z`
  over the column must recover the boundary-flux difference `ℐ_net[1] - ℐ_net[end]`. This
  fails if `flux_divergence` drops any of the four streams. Verified by reintroducing the
  bug locally, which trips the clear-sky assertion with 109.5 vs 79.9.
* Assert `upwelling_shortwave_flux` is identically zero for gray optics.

## Sign convention

Unchanged, and the new field follows it: all four flux fields are stored positive-upward,
with downwelling fluxes negated at copy time. The net flux is therefore the plain sum of
the four, matching RRTMGP's own `flux_net = flux_up - flux_dn`, and `flux_divergence` is
`-(F_{k+1} - F_k) / Δzᶜᶜᶜ`, positive for convergence.

## Not in this PR

Gray optics silently ignores `surface_albedo` in the shortwave: `NoScatSWRTE` receives
`sfc_alb_direct` and `sfc_alb_diffuse`, but `rte_sw_noscat!` reads only `toa_flux` and
`cos_zenith`, so the direct beam is never reflected and the surface absorbs all of it.
Fixing that means swapping in a scattering shortwave solver, which would change every
gray result, so this PR documents the limitation with a warning in
`docs/src/radiative_transfer.md` and points users to `ClearSkyOptics` or `AllSkyOptics`
instead.

## Testing

`gray_radiative_transfer`, `clear_sky_radiative_transfer`, `all_sky_radiative_transfer`,
and `radiation_scheduling` all pass on CPU, in Float64 and with Float32 enabled.
`ExplicitImports` passes with the new import. The docs were not built locally; the two
`@ref` links added resolve through the same `@autodocs` mechanism as the existing
`ClearSkyOptics` reference on the same page.


glwagner (approved):
good catch

kaiyuan-cheng:
@giordano Do you know why CI gets stuck?

giordano:
[GitHub has been on fire for several hours now](https://www.githubstatus.com/incidents/qcvjkzcs7j74)

PR #891, opened 2026-08-03
Convert 1M accretion collision efficiencies to the working precision

jtbuch:
## Problem

The one-moment accretion wrappers forward `option.e` (and `option.coeff_disp` for rain–snow collisions) straight into CloudMicrophysics:

```julia
@inline function cloud_precipitation_accretion(option, cloud, precipitation, velocity, qᶜ, qᵖ, ρ)
    return accretion(cloud, precipitation, velocity, option.e, qᶜ, qᵖ, ρ)
end
```

CloudMicrophysics' low-level accretion kernels are written as `where {FT}` over *every* float argument (`E`, `q_clo`, `q_pre`, `ρ`, and the velocity parameters), so a collision efficiency stored at a different precision than the model fields leaves them with no applicable method. The wrapper body then infers to `Union{}`, which throws a `MethodError` on the CPU and, on the GPU, surfaces as a dynamic call inside `cloud_precipitation_accretion` — the reported symptom is a dynamic `getfield` on `option.e`, which is just the un-lowered `Base.getproperty` left behind once inference gives up.

This bites whenever the microphysics parameters and the model do not agree on a float type — a `Float64` default carried into a `Float32` model, or a single perturbed parameter supplied by a calibration.

## Fix

Convert the collision efficiency and the velocity dispersion coefficient to the working precision before the call. This matches what `condensate_formation_from_option` already does one screen up, where relaxation rates are converted with `convert(typeof(reference), ...)`.

## Tests

New testset `Accretion options stored at another precision`, run over `all_float_types()` with the accretion options deliberately stored at the *other* precision. It exercises all three wrappers (`cloud_precipitation_accretion`, `rain_sink_accretion`, `rain_snow_accretion`) and checks that the result comes back at the model's float type, so an accidental promotion to `Float64` inside a `Float32` model is caught too. The testset errors on `main` and passes here; the rest of `test/cloud_microphysics_1M.jl` passes unchanged on CPU.

## Not addressed here

- A mismatch spanning the *whole* `Microphysics1MParams` (rather than just the process options) still fails, now at `terminal_velocity` in `update_microphysical_auxiliaries!`. `OneMomentCloudMicrophysics(FT)` reads `Oceananigans.defaults.FloatType` at call time and is never reconciled against the grid; a construction-time check would be the right answer, but it is a separate concern.
- `ice_autoconversion_with_supersaturation` unpacks `option.r_ice_snow` without converting, so a mixed-precision `WithSupersaturation` would make its trailing `ifelse` type-unstable. Same defect class, different process, left out to keep this focused.


giordano:
If this is meant to resolve compatibility with CloudMicrophysics v0.37.2, then https://github.com/NumericalEarth/Breeze.jl/blob/cb4f4805d641205dcb2d002659ac5ef5d001859a/Project.toml#L36 and https://github.com/NumericalEarth/Breeze.jl/blob/cb4f4805d641205dcb2d002659ac5ef5d001859a/test/Project.toml#L33 should be relaxed to
```toml
 CloudMicrophysics = "0.37"
```

jtbuch:
@giordano your suggestion may not work as `0.37.2` turned `option` into a singleton type and moved its data into a new `process_params` field.

giordano:
Ok, I may have misunderstood your message elsewhere, I thought this was to address the changes in CloudMicrophysics@v0.37.2. The changes in options are handled in #878.

glwagner (approved):

PR #893, opened 2026-08-04
Resolve the three readings of `surface_pressure`

### LINKED ISSUE #887: `surface_pressure` is read as three different quantities; flat and terrain reference states disagree on the datum
`surface_pressure(dynamics)` is read as three different physical quantities depending on which code path consumes it, and the two reference-state implementations of the *same* concept disagree with each other. On a flat grid whose bottom is at z = 0 all three coincide, which is why this has gone unnoticed.

The most visible consequence is that `HydrostaticallyBalancedDensity`, the documented one-call cold start, produces an initial state that disagrees with its own reference state by O(ρgh) over terrain. That section is below.

## The two reference states use different datums

`_compute_exner_column!` (`reference_states.jl:611-629`) anchors with `Π¹ = π₀_surface - g * Δzᶜ₁ / (2 * cᵖᵐ¹ * θ¹)`, a half step from the grid's **bottom face**. No `znode` appears, so `p₀` is the pressure at the domain bottom, whatever height that is.

`terrain_surface_reference_fields` instead integrates from **absolute z = 0** up to each column's terrain height, and `test/terrain_following_reference_state.jl` asserts `pᵣ` equals the continuous profile from `p₀` at each cell's absolute height.

They agree only when the domain bottom is at 0.

```julia
# Flat grid whose bottom sits at 2000 m. Is p0 interpreted as pressure at z=0, or at the domain bottom?
using Oceananigans
using Oceananigans.Fields: interior
using Breeze
using Breeze: CompressibleDynamics, AtmosphereModel, ExplicitTimeStepping
using Breeze.Thermodynamics: hydrostatic_pressure, ThermodynamicConstants

θ₀, pˢᵗ, p₀ = 288.0, 1e5, 101325.0
constants = ThermodynamicConstants(Float64)

grid = RectilinearGrid(CPU(); size = (4, 40), x = (0, 1e4), z = (2000, 6000),
                       topology = (Periodic, Flat, Bounded))
model = AtmosphereModel(grid; dynamics = CompressibleDynamics(ExplicitTimeStepping();
                            surface_pressure = p₀, reference_potential_temperature = θ₀))

pᵣ1 = interior(model.dynamics.reference_state.pressure)[1, 1, 1]
z₁ = 2000 + (6000 - 2000) / 40 / 2       # first cell centre height
println("first cell centre       = ", z₁, " m")
println("reference pr[1]         = ", round(pᵣ1, digits = 1), " Pa")
println("if p0 were at z = 0     = ", round(hydrostatic_pressure(z₁, p₀, θ₀, pˢᵗ, constants), digits = 1), " Pa")
println("if p0 at domain bottom  = ", round(hydrostatic_pressure(z₁ - 2000, p₀, θ₀, pˢᵗ, constants), digits = 1), " Pa")
```

```
first cell centre       = 2050.0 m
reference pr[1]         = 100727.2 Pa
if p0 were at z = 0     = 78818.4 Pa
if p0 at domain bottom  = 100727.2 Pa
```

The docstrings at `reference_states.jl:19`, `:473` and `:709` describe the z = 0 convention for both, so they are wrong for the flat path on any raised domain.

## Three readings in total

| reading | consumers |
|---|---|
| pressure at the column's bottom face | `_compute_exner_column!` (flat reference), `HydrostaticallyBalancedDensity` via `_compute_exner_reference_3d!` (`set_to_mean.jl:297`) |
| pressure at absolute z = 0 | `terrain_surface_reference_fields` (terrain reference) |
| pressure at the ground | `BulkSensibleHeatFluxFunction` (`BoundaryConditions.jl:369`), `BulkVaporFluxFunction` (`:395`), bulk drag (`bulk_drag.jl:120,131`), stability functions via `surface_virtual_potential_temperature` (`polynomial_bulk_coefficient.jl:655`) |

`compute_hydrostatic_pressure!` applies the first reading on terrain grids, where the bottom face is the terrain surface rather than a common level. It is exported (`Breeze.jl:71,124`, `AtmosphereModels.jl:93`) so a user can call it on a terrain grid today, though there is no in-tree caller outside `test/diagnostics.jl:247`.

The first and second readings are the ones that collide, because on a terrain grid the bottom face is per-column while z = 0 is common.

## HydrostaticallyBalancedDensity disagrees with its own reference over terrain

`HydrostaticallyBalancedDensity` is exported (`Breeze.jl:39`), is not gated on terrain, and is the documented one-call cold start. It routes through `_compute_exner_reference_3d!` to `_compute_exner_column!`, so every column is anchored at `p₀` on **its own** terrain surface. The terrain reference state, built from the same scalar, instead reduces `p₀` from z = 0 to each column's height. The two therefore disagree horizontally by O(ρgh).

```julia
# HydrostaticallyBalancedDensity on terrain: does the cold-start state agree with its own reference?
using Oceananigans
using Oceananigans.Fields: interior
using Oceananigans.Grids: znode, Center
using Breeze
using Breeze: CompressibleDynamics, AtmosphereModel, ExplicitTimeStepping, HydrostaticallyBalancedDensity,
              TerrainFollowingVerticalDiscretization, LinearDecay, materialize_terrain!

h₀, a, pˢᵗ, θ₀ = 2000.0, 20000.0, 1e5, 288.0
hill(x) = h₀ * exp(-(x / a)^2)

z_faces = TerrainFollowingVerticalDiscretization(collect(range(0, 20000, length = 41));
                                                 formulation = LinearDecay())
grid = RectilinearGrid(CPU(); size = (12, 40), x = (-6e4, 6e4), z = z_faces,
                       topology = (Periodic, Flat, Bounded))
materialize_terrain!(grid, hill)

model = AtmosphereModel(grid; dynamics = CompressibleDynamics(ExplicitTimeStepping();
                            surface_pressure = 101325.0, reference_potential_temperature = θ₀))
set!(model; θˡⁱ = θ₀, qᵗ = 0, ρ = HydrostaticallyBalancedDensity(), compute_reference_state = true)

p  = interior(model.dynamics.pressure)[:, 1, 1]
pᵣ = interior(model.dynamics.reference_state.pressure)[:, 1, 1]

println(" i   terrain h [m]   z1 [m]     state p[k=1]   reference p[k=1]   state - ref [Pa]")
for i in 1:2:11
    xi = -6e4 + (i - 0.5) * (1.2e5 / 12)
    z₁ = znode(i, 1, 1, grid, Center(), Center(), Center())
    println(lpad(i,2), lpad(round(hill(xi), digits=1), 14), lpad(round(z₁, digits=1), 10),
            lpad(round(p[i], digits=1), 16), lpad(round(pᵣ[i], digits=1), 19),
            lpad(round(p[i] - pᵣ[i], digits=1), 19))
end
println("\nspread of state p[k=1]     = ", round(maximum(p) - minimum(p), digits=1), " Pa")
println("spread of reference p[k=1] = ", round(maximum(pᵣ) - minimum(pᵣ), digits=1), " Pa")
```

```
 i   terrain h [m]   z1 [m]     state p[k=1]   reference p[k=1]   state - ref [Pa]
 1           1.0     251.0         98361.5            98349.3               12.3
 3          93.5     342.4         98375.1            97282.1             1093.0
 5        1139.6    1375.3         98528.5            85782.9            12745.6
 7        1878.8    2105.3         98637.1            78265.7            20371.4
 9         419.2     664.0         98422.8            93590.7             4832.2
11          12.7     262.5         98363.2            98214.8              148.5

spread of state p[k=1]     = 275.6 Pa
spread of reference p[k=1] = 20083.6 Pa
```

Each column is individually in vertical hydrostatic balance, which is the marker's stated contract. The defect is horizontal: the state carries essentially no surface pressure gradient across a 1.9 km hill (275.6 Pa, which is just the half-cell Δz variation), while the reference built from the identical scalar carries 20.1 kPa of it.

Dropping `compute_reference_state = true` gives identical numbers, so the pairing the docstring recommends does not avoid this. The `surface_pressure` kwarg on the marker is also a scalar, so there is no per-column workaround available to a user.

## What the third reading costs

A flat plateau at 1921 m, true ground pressure 79672 Pa, true sea level 100798 Pa:

```julia
# `surface_pressure` is consumed both as a z = 0 datum (reference state) and as ground pressure
# (bulk surface fluxes). Over terrain those are different numbers, so no single value is correct
# for both. Breeze + Oceananigans only; no data, no downloads.

using Oceananigans
using Oceananigans.Fields: interior
using Oceananigans.Grids: znode, Center
using Breeze
using Breeze: CompressibleDynamics, AtmosphereModel, ExplicitTimeStepping,
              TerrainFollowingVerticalDiscretization, LinearDecay, materialize_terrain!
using Breeze.Thermodynamics: hydrostatic_pressure, potential_temperature_from_temperature,
                             surface_density, ThermodynamicConstants

z_plateau, θᵣ, pˢᵗ, T_ground = 1921.0, 288.0, 1e5, 300.0
p_sea_level = 100798.0
constants = ThermodynamicConstants(Float64)

# Physically correct pressure at the plateau surface, from the correct sea-level datum.
p_ground = hydrostatic_pressure(z_plateau, p_sea_level, θᵣ, pˢᵗ, constants)
θ_ground_true = potential_temperature_from_temperature(T_ground, p_ground, pˢᵗ, constants)
ρ_ground_true = surface_density(p_ground, T_ground, constants)

z_faces = TerrainFollowingVerticalDiscretization(collect(range(0, 20000, length = 51));
                                                 formulation = LinearDecay())
make_grid() = (g = RectilinearGrid(CPU(); size = (8, 50), x = (-1e5, 1e5), z = z_faces,
                                   topology = (Periodic, Flat, Bounded));
               materialize_terrain!(g, x -> z_plateau); g)

println("plateau ", z_plateau, " m;  true ground pressure ", round(p_ground, digits=1),
        " Pa;  true θ_ground ", round(θ_ground_true, digits=2), " K\n")
println(rpad("surface_pressure =", 24), rpad("reference err", 16), rpad("θ₀ err", 12), "ρ₀ err (drag)")

for p₀ in (81159.0, p_sea_level)
    grid = make_grid()
    model = AtmosphereModel(grid; dynamics = CompressibleDynamics(ExplicitTimeStepping();
                                surface_pressure = p₀, reference_potential_temperature = θᵣ))

    # consumer 1: reference state, wants the z = 0 datum
    z₁ = znode(1, 1, 1, grid, Center(), Center(), Center())
    ref_err = interior(model.dynamics.reference_state.pressure)[1, 1, 1] -
              hydrostatic_pressure(z₁, p_sea_level, θᵣ, pˢᵗ, constants)

    # consumer 2: BulkSensibleHeatFluxFunction, wants pressure at the ground
    θ_err = potential_temperature_from_temperature(T_ground, p₀, pˢᵗ, constants) - θ_ground_true
    ρ_err = 100 * (surface_density(p₀, T_ground, constants) - ρ_ground_true) / ρ_ground_true

    println(rpad(string(round(p₀, digits=1), " Pa"), 24),
            rpad(string(round(ref_err, digits=1), " Pa"), 16),
            rpad(string(round(θ_err, digits=2), " K"), 12),
            round(ρ_err, digits=1), " %")
end
```

```
plateau 1921.0 m;  true ground pressure 79672.1 Pa;  true θ_ground 320.11 K

surface_pressure =      reference err   θ₀ err      ρ₀ err (drag)
81159.0 Pa              -16236.2 Pa     -1.69 K     1.9 %
100798.0 Pa             -0.0 Pa         -20.79 K    26.5 %
```

No value zeroes both columns. The density error is the consequential one: bulk drag is `-ρ₀ Cᴰ |U| u`, so 26.5 % in `ρ₀` is 26.5 % in surface momentum drag, and it scales the sensible heat and moisture fluxes too.

## Why this matters now

#863 made the perturbation-form PGF the default, so the reference now enters `∂ₓ(p - pᵣ)` directly. Under the cold start above the model sees `p - pᵣ` of about 20.4 kPa at the summit at t = 0, which is a concrete initial-condition inconsistency rather than a gradual degradation of cancellation.

NumericalEarth/NumericalEarth.jl#488 is a separate caller passing a terrain-level value into the datum slot. It cannot be fixed there, because any single scalar only chooses which consumer to break.

## Constraints on a fix

Recording these because the obvious fix does not work as stated.

1. **A rename alone is not enough, and is not free.** The call site uses the interface `AtmosphereModels.surface_pressure(dynamics)` (`dynamics_interface.jl:222`), implemented by four dynamics types, so disambiguation has to happen at the accessor, not on one field. Renaming `CompressibleDynamics.surface_pressure` to `sea_level_pressure` would also be inaccurate for the flat path unless the flat anchor moves to z = 0, which changes the reference state, the PGF and rest-state discrete balance for every elevated flat domain. `ExnerReferenceState.surface_pressure` would additionally keep the old name for the same number.

2. **The BC path cannot read `pˢ` as constructed today.** BCs are materialized at `atmosphere_model.jl:188`, before `materialize_dynamics` at `:196` builds the reference, so the per-column `pˢ` does not exist yet, and the dependency is circular. `pˢ` is also rebuilt by `reset_reference_state!` and by `set!(model; compute_reference_state=true)`, so a captured copy goes stale.

3. **The reference is an approximation for the BCs anyway.** What the bulk fluxes physically want is the model's own diagnostic pressure at k = 1 extrapolated half a cell down. That is terrain-agnostic, self-updating, and removes the datum question from the BC path entirely, and `getbc` already reads `fields.θ` / `fields.qᵛ`, so it is in idiom. It would need `Oceananigans.fields(model)` (`atmosphere_model.jl:534-538`) to expose the diagnostic pressure, and the same construction-ordering question applies.

Happy to take a cut at any of these once the direction is agreed. Filing rather than PRing because the datum convention looks like a judgement call about intended semantics rather than a bug with an obvious fix.

kaiyuan-cheng:
Closes #887.

`surface_pressure` was read as three different physical quantities depending on the consumer, and the two reference-state implementations of the *same* concept disagreed about which one they meant. On a flat grid whose bottom is at z = 0 all three coincide, which is why it went unnoticed.

This PR resolves all three readings: two by fixing the datum and naming the quantities apart, the third by having the surface fluxes stop reading a construction-time constant at all.

## 1. The datum was not a judgement call

The issue frames the z=0-versus-bottom-face split as ambiguity about intended semantics. It isn't. Counting the reference constructors:

| constructor | anchor |
|---|---|
| `ReferenceState` | `set!(pᵣ, z -> hydrostatic_pressure(z, p₀, …))`, absolute `z` |
| `_compute_hydrostatic_reference!` | `z⁻ = 0`, `p⁻ = p₀`, steps to `znode` |
| `_compute_isothermal_reference!` | `z₁ = znode(1,1,1,…)`, `p₀ exp(-g z₁/(Rᵈ T₀))` |
| `terrain_exner_reference_state` | continuous integration from z = 0 to terrain height |
| **`_compute_exner_column!`** | **`Δzᶜᶜᶜ(i,j,1)/2` from the bottom face, no `znode`** |

Four of five already implemented the z = 0 datum that all three docstrings claim. The outlier is one kernel, and its sibling 80 lines above it in the same file, filling the same struct, already used `znode` — so the isentropic and isothermal constructors of `ExnerReferenceState` disagreed with *each other* on any raised domain.

The datum is now reduced to the column's bottom face on the host, where the reference profile is still a callable, and that anchor is passed into the column kernel. The reduction is `moist_hydrostatic_pressure`, moved into `Thermodynamics` from `terrain_compressible_physics.jl` so the height-coordinate and terrain paths share one implementation instead of two that can drift.

`ExnerReferenceState` retains the anchor as a 2D field: horizontally uniform on a height-coordinate grid, whose bottom face is a single level; the genuine terrain-surface field on a terrain-following grid, where the terrain path already computed it and then discarded it.

## 2. Naming

Keeping `surface_pressure` for the datum would have left the worse name on the wrong quantity: in every other atmospheric context "surface pressure" means the pressure at the ground. Having to know which of two senses each site meant is how the original defect survived.

| name | meaning | type |
|---|---|---|
| `base_pressure` | pressure of the reference atmosphere at z = 0, the datum. A property of the atmosphere, not the grid | scalar |
| `surface_pressure` | pressure at the ground, i.e. at each column's bottom face | 2D `(Center, Center, Nothing)` field, horizontally uniform on a height-coordinate grid |

They are equal, exactly, on the usual domain starting at z = 0, which is why one name served both for so long.

```
   z = 6000 ────────────────────  domain top
   z = 2000 ════════════════════  domain bottom → surface_pressure = 79320 Pa
   z = 0    ─ ─ ─ ─ ─ ─ ─ ─ ─ ─   the datum    → base_pressure    = 101325 Pa
```

`surface_pressure(dynamics)` is extended by each dynamics that carries a materialized reference state, and throws rather than silently falling back to the datum when one does not, so the ambiguity cannot reappear as a quiet wrong answer.

## 3. Consumers stop disagreeing with the reference

**`HydrostaticallyBalancedDensity`** anchored every column at the single scalar, so over a 1.9 km hill the cold start carried 276 Pa of surface pressure gradient while its own reference carried 20.1 kPa. Since #863 made the perturbation-form PGF the default, the model saw that 20.4 kPa at t = 0. It now anchors on the reference's per-column pressure; without a reference state it reduces the datum to each column's bottom face along the current near-surface state, so terrain cold starts work with `reference_state = nothing` too.

**`set_to_mean!`** rewrites the bottom-face anchor when it recomputes the reference: the datum is fixed, but the profile used to reduce it to the ground is not, so re-meaning the state moves the anchor and the bottom boundary values of `pressure`/`density` with it. It reduces the datum with `moist_hydrostatic_pressure`, the same function the constructor anchors on, and dispatches to the per-column integration when the reference fields are per-column rather than a single broadcast column.

Both reference states stay **immutable**, which they must, because the `dynamics` object that owns them is passed directly into GPU kernels and so has to be `isbits` after `Adapt`. The bottom-face pressure and density are 2D fields rewritten in place, and the bottom `ValueBoundaryCondition` of the reference `pressure`/`density` is built once from a 2D aliasing view of them (`surface_boundary_value`), so the boundary value follows a reset with nothing rebuilt.

**`compute_hydrostatic_pressure!`** integrates upward from the bottom face, so it must start from the pressure there. It now diagnoses that from the live model state by extrapolating the first cell centre down half a cell, which makes it both terrain-correct and independent of the datum convention.

## 4. Surface fluxes diagnose their own pressure

The third reading. `BulkDrag`, `BulkSensibleHeatFlux`, `BulkVaporFlux` and `PolynomialCoefficient` used a scalar captured at construction to build `ρ₀`, the surface saturation state, and the stability correction's `θᵥ₀`. Over a 1921 m plateau that scalar is wrong by the full reduction — ~26 % in `ρ₀`, and so in the surface drag and the sensible and moisture fluxes.

Rather than plumbing the reference's ground pressure into the boundary conditions — which is awkward, since they are materialized before `materialize_dynamics` builds the reference — the flux functions now compute it at evaluation time from the model state:

- `surface_pressure_from_cell_center(p, ρ, Δz, g) = p exp(g Δz ρ / 2p)` extrapolates the first-cell pressure down half a cell along the local scale height. An indexed method, `surface_pressure_from_cell_center(i, j, k, grid, p, ρ, g)`, is the single definition all three consumers go through (the surface fluxes, `compute_hydrostatic_pressure!`, and `PrescribedDynamics`), so they cannot drift on which level or which half-cell factor they use.
- `dynamics_thermodynamic_fields(dynamics)` carries `p` and `ρ`, which `boundary_condition_args` passes as a second tuple after the model fields; Breeze's own `getbc` methods merge the two on entry through `surface_layer_state` and read by name from there on. They are `dynamics_pressure` and `total_density` — always actual `Field`s, and the same thermodynamic pressure the rest of the model uses: prognostic under `CompressibleDynamics`, the hydrostatic reference profile under `AnelasticDynamics`. Deliberately not `total_pressure`, whose anelastic form is a lazy sum that would rebuild an `AbstractOperation` on every halo fill, and whose nonhydrostatic anomaly is defined only up to a constant.
- `wall_air_pressure(i, j, k, grid, side, direction, fields, constants)` wraps it: half a cell down under `Bottom`, half a cell up over `Top`, and the plain cell-centre pressure on a vertical wall, where wall and centre sit at the same height. Momentum fluxes interpolate it to the face of the component they act on. (It was `surface_air_pressure`, bottom-only, until merging `main` generalised the bulk fluxes to all six walls.)

The stored pressure is gone from all four; `PolynomialCoefficient` now holds `standard_pressure` and takes the surface pressure as an argument. Because the live `p`/`ρ` are reachable from the surface-layer tuple, the two construction-time hooks that existed only to hand the boundary conditions a reference profile — `boundary_conditions_reference_state` and `boundary_conditions_thermodynamic_state`, the latter of which rebuilt a whole reference state during boundary-condition materialization — are gone too. The result is terrain-agnostic, datum-agnostic, and follows the evolving state instead of freezing the value seen at construction.

Once nothing is captured at materialization time, the state that used to be threaded there is dead. `base_pressure`, `microphysical_fields`, `specific_prognostic_moisture` and `temperature` were still being passed down roughly twenty signatures without a single leaf reading them, so they are removed. That also removes the `materialize_microphysical_fields` call that built a full set of microphysical fields during boundary-condition materialization only to discard them, and `VirtualPotentialTemperature(grid; reference_state, …)`, which was the last place `Diagnostics` reached into reference-state internals.

The surface θᵥ goes through the same `virtual_potential_temperature` as the air-side value, so both sides of the bulk Richardson difference come from one definition rather than two copies of the same algebra on either side of a subtraction.

This also fixes the Monin–Obukhov measurement height on terrain. `evaluation_height` used `znode(i, j, 1, grid, Center(), Center(), Center())`, the height of the first cell center above **z = 0**; over a 1921 m plateau that is ~1921 m rather than the half-cell it is meant to be. It is now `Δzᶜᶜᶜ(i, j, 1, grid) / 2`, the height above the local ground.

The `ρ₀` change supersedes the old Monin–Obukhov argument in `BulkDragFunction`'s docstring, which justified the construction-time constant on the grounds that diagnosing density from the first cell would make `ρ₀` grid-dependent. The half-cell hydrostatic extrapolation gives a true surface density, so that concern does not apply.

## 5. Two surface-state quantities were wrong, independently of the datum

Found while reviewing the above, and fixed here because both are surface-state bugs that the same
consumers hit.

**The stability correction compared a temperature against a potential temperature.**
`surface_virtual_potential_temperature` returned `T₀ (1 + δᵛᵈ qᵛ⁺)` — a virtual *temperature* — and
`bulk_richardson_number` differenced it against `θᵥ` aloft, a virtual *potential* temperature, then
normalised by their mean. The bias is ~1 K at 1 atm, the same order as the air–surface differences
Riᴮ exists to resolve. It now divides by the dry Exner factor `Π₀ᵈ = (p₀/pˢᵗ)^{Rᵈ/cᵖᵈ}`, so both
sides of the difference are potential temperatures. This shifts Cᴰ/Cᵀ/Cᵛ in every stability-corrected
run, flat grids included.

`bulk_richardson_number` also took gravity as a `g = 9.81` default that every caller left unset, so a
run configured with a different gravitational acceleration silently used Earth's. It now comes from
the thermodynamic constants.

**The static-energy surface state omitted the geopotential.** `e = cᵖᵐ T + g z − ℒˡ qˡ − ℒⁱ qⁱ`, but
the surface value it is differenced against was `cᵖᵐ T₀`, with no `g z₀`. On a raised or terrain
domain that mis-states the sensible-heat flux by `g z₀ / cᵖ`, about 20 K of apparent Δθ per 2 km of
elevation. It is now `cᵖᵐ T₀ + g z₀` at the local bottom face.

**The reference's ground temperature is now stored rather than inverted.**
`default_drag_surface_temperature` recovered it as `pˢ / (Rᵈ ρˢ)`, exact only because the
`ReferenceState` constructor builds its density with the dry gas law; `compute_reference_state!`
rebuilds that density with the mixture constant, so after a reset the same expression returns the
*virtual* temperature, high by ~3.6 K at qᵛ = 0.02. The reference now carries its bottom-face
temperature, set consistently by both paths, so no gas constant has to close an inversion.

## Resolved during review: `ContinuousForcing` with `field_dependencies` on GPU

Recorded because the constraint is easy to undo by accident.

Admitting `p` and `ρ` to `fields(model)` made that NamedTuple heterogeneous: under
`AnelasticDynamics` both are reduced `Field{Nothing, Nothing, Center}` objects, where every other
entry adapts to a bare `OffsetArray`. `ContinuousForcing` resolves `field_dependencies` to
positional indices and dereferences them through `field_arguments`, which indexes the tuple at a
runtime position, so the lookup inferred as a `Union` and GPUCompiler rejected every kernel
performing one — `ijl_get_nth_field_checked`. BOMEX, whose momentum drag depends on `(:ρu, :ρv)`,
died that way in the docs build.

Fixed by moving the pair into `dynamics_thermodynamic_fields`, passed as a third element of
`boundary_condition_args`. `fields(model)` stays homogeneous for the positional lookup; boundary
conditions read `p` and `ρ` by name, which is type-stable however heterogeneous the merged tuple
is. `test/forcing_and_boundary_conditions.jl` locks the invariant on CPU CI by adapting the tuple
and inferring the lookup, and drives a real `field_dependencies` flux BC end to end.

Both GPU runners pass.

## Results

Both reproductions from the issue:

```
Flat grid raised to z = 2000 m
  reference pr[1]        = 78818.4 Pa   ← was 100727.2
  continuous from z = 0  = 78818.4 Pa

Terrain, HydrostaticallyBalancedDensity cold start over a 1.9 km hill
  spread of state p[k=1]     = 20083.6 Pa   ← was 275.6
  spread of reference p[k=1] = 20083.6 Pa
  max |state - reference|    = 0.066 Pa     ← was 20371.4
```

## Compatibility

**No behavior change to the reference states where the domain bottom is at z = 0.** The datum reduction returns the datum exactly. Verified by running the same probe on `main` and on this branch and diffing at 17 significant figures: dry, moist, 3D-θ and isothermal references plus a full model cold start are bit-identical. The adaptive quadrature was later factored into a shared refinement driver, and its abscissae are written to stay bit-for-bit what they were, so profiles built from a `z`-dependent θ remain reproducible to the last ulp — re-verified against `origin/main` after that refactor.

**Surface fluxes change everywhere, by a small amount, on flat grids too.** They previously used the datum and now use the diagnosed pressure at the ground, which differs by the half-cell hydrostatic extrapolation and by however far the model state has drifted from the reference. On a flat domain at z = 0 that is a fraction of a percent at t = 0; over terrain it is the full O(ρgh) correction. Several tests were updated to compare against `wall_air_pressure` rather than the constant.

**The rename is breaking, deliberately.** Nothing keeps accepting `surface_pressure` for a datum — that is the ambiguity being removed. But the five affected constructors do now *reject* it with an error naming `base_pressure` and explaining that the old spelling means something else today, rather than a bare `MethodError`, because a script carried over from before the rename would otherwise build a different atmosphere for no visible reason. This follows how `CompressibleDynamics` already reports the relocated `temperature_tolerance` keyword. `PrescribedDynamics` keeps its `surface_pressure` keyword, which is the bottom-face quantity and legitimate. NumericalEarth/NumericalEarth.jl#488 passes a terrain-level value into the datum slot and now has two named slots to choose between.

**Snow's superscript changes, and it is a public prognostic field.** Introducing `pˢ` collided with
snow's `qˢ`, so per review discussion `ˢ` is reserved for *surface* in thermodynamic and
surface-layer contexts and snow takes `ˢⁿ`: `set!(model; ρqˢ=…)` is now `ρqˢⁿ`, and
`AM.microphysical_fields.ρqˢ` is `ρqˢⁿ`. Also renamed: `qˢ`, `𝕎ˢ`, `wˢ` (snow terminal velocity —
the subsidence `wˢ` is untouched), `ρnˢ`, `δˢ`, and the conversion source terms `Sᵃᶜᶜˡˢ`, `Sᵃᶜᶜⁱˢ`,
`Sᵃᶜⁿᵛⁱˢ`, `Sʳˢ`, `Sˢʳ`. This resolves a latent collision too: `wˢ` previously meant both snow
terminal velocity and subsidence velocity, in different files.

**The stability correction and the static-energy surface flux change on every grid**, including flat ones at z = 0, for the reasons in §5. These are corrections to quantities that were wrong independently of the datum, so unlike the reference states they are not bit-identical to `main` anywhere. BOMEX and RICO want revalidating against the Riᴮ change before this merges.

## Tests

- `test/reference_states.jl` — a raised domain pins both `ExnerReferenceState` constructors to the z = 0 datum and asserts they agree with each other (fails on `main`); a `set_to_mean!` reset on that domain must rebuild the bottom-face anchor and the bottom boundary value of `pressure`.
- `test/terrain_following_reference_state.jl` — the terrain cold start must reproduce its reference's horizontal pressure range rather than flatten it; `compute_hydrostatic_pressure!` must carry the same gradient; bulk fluxes over a 1921 m plateau must see a surface pressure matching the reference's terrain-surface value and must track the state when it changes; and a terrain cold start with `reference_state = nothing` must still develop the full surface pressure gradient.
- `test/forcing_and_boundary_conditions.jl`, `test/polynomial_bulk_coefficients.jl` — updated to the diagnosed pressure; the static-energy surface flux must vanish for an isothermal column on a raised domain, which it cannot without the `g z₀` term; `surface_virtual_potential_temperature` is pinned against the explicit Exner reduction.
- `test/reference_states.jl` — a moist reset must report the true ground temperature and not the virtual one, and the old `surface_pressure` keyword must be rejected with a message naming its replacement.
- `test/diagnostics.jl` — the "Hydrostatic pressure computation" expectation was still anchored at the datum and so was left failing by the change in §3. It now checks the diagnosed anchor against the datum at the accuracy the half-cell extrapolation has, and the column integration against that anchor at full precision, rather than loosening one tolerance over both.

The cleanup commit adds no tests: it removes dead parameters and consolidates duplicated helpers, so it is covered by the existing suites. One behavioural edge does tighten, though. `columnwise_surface_pressure_field` now returns the datum directly when the bottom face sits at z = 0, which is what the reduction means there, so `surface_pressure == base_pressure` holds exactly on the 3D path as it already did on the column path; previously that case round-tripped through `pˢᵗ ((p₀/pˢᵗ)^κ)^(1/κ)` and came back a few ulp off.

## Verification

CPU, against the branch head: `reference_states` 310/310, `terrain_following_reference_state`
540/540, `forcing_and_boundary_conditions` 84/84, `polynomial_bulk_coefficients` 195/195,
`diagnostics` 79/79, `parcel_dynamics` 51/51, `atmosphere_model_construction` 380/380,
`set_atmosphere_model` 59/59, `adiabatic_balance_set` 57/57, `kinematic_driver` 21/21,
`moist_air_buoyancy` 1/1, `quality_assurance` (Aqua 31/31, ExplicitImports 5/5, No Core.Box),
doctests.

GPU: a terrain-following compressible model with all three bulk fluxes and a
`HydrostaticallyBalancedDensity` cold start; a split-explicit compressible model with a
stability-corrected `PolynomialCoefficient` drag, which exercises `_explicit_horizontal_step!` — the
substepper kernel that receives `dynamics` and reads the reference state; and an anelastic model on a
raised domain whose reference reset leaves the stored ground pressure and the bottom-face boundary
value agreeing to 1e-10.

After the cleanup commit, re-run on GPU: `reference_states`, `polynomial_bulk_coefficients`,
`boundary_conditions`, `kinematic_driver`, `diagnostics` (614); `terrain_following_reference_state`,
`unit_tests`, `atmosphere_model_construction`, `set_atmosphere_model`, `quality_assurance` (1184);
`latitude_longitude_grid`, `terrain_following_latlon`, `parcel_dynamics`, `moist_air_buoyancy`,
`dynamics` (3875); doctests. All pass, `forcing_and_boundary_conditions` included once the
`ContinuousForcing` regression above was fixed.

After merging `main`, the full CI matrix is green: both `aws-linux-nvidia-gpu-l4-4xl` runners on
Julia 1.11.9 and 1.12.6, ubuntu/macOS/windows on 1.12.6 and Julia min, six GPU benchmark jobs,
`Validation`, and the docs build.


docs/src/appendix/notation.md
```diff
@@ -106,7 +106,8 @@ The following table also uses a few conventions that suffuse the source code and
 | ``\mathcal{L}^l(T)``                | `ℒˡ`   | `liquid_latent_heat(T, constants)`  | Temperature-dependent latent heat of condensation                              |
 | ``\mathcal{L}^i(T)``                | `ℒⁱ`   | `ice_latent_heat(T, constants)`     | Temperature-dependent latent heat of deposition                                |
 | ``θ₀``                              | `θ₀`   | `RS.potential_temperature`          | (Constant) reference potential temperature for the anelastic formulation       |
-| ``p₀``                              | `p₀`   | `RS.surface_pressure`               | Surface reference pressure                                              |
+| ``p₀``                              | `p₀`   | `RS.base_pressure`                  | Reference pressure at ``z = 0``: the datum the reference profiles are anchored to |
+| ``pˢ``                              | `pˢ`   | `RS.surface_pressure`               | Reference pressure at a column's bottom face (the terrain surface over terrain), ``p₀`` reduced to that height |
```
glwagner, line 111: "^s" means "snow" ... although it may not be confused, something to think about.

kaiyuan-cheng: how about "p_s"?

glwagner: We could use ^s for "surface" and change snow to "^sn" -- what do you think?

I don't think we should use sub vs super script to distinguish meaning. that might be confusing

kaiyuan-cheng: I agree the potential confusion. How about p^{sfc}?

kaiyuan-cheng: On a second thought, I will proceed with your proposal to change the symbol for snow.

glwagner: ok!

navidcy:
what's referred to as "consumer"?

glwagner:
> what's referred to as "consumer"?

I think generally "consumer" refers to the code that calls a function. It is a bit more general than "user" (because a "consumer" may be an internal function / routine).

If I am reading the description correctly, three consumers of `surface_pressure` are `HydrostaticallyBalancedDensity`, `set_to_mean!` and `compute_hydrostatic_pressure!`.

navidcy:
oh gotcha!

glwagner:
@kaiyuan-cheng should we merge this? lots of breaking changes though.

kaiyuan-cheng:
> @kaiyuan-cheng should we merge this? lots of breaking changes though.

@glwagner I think we should merge this. I will resolve conflict and then wait for approval.

ewquon:
@kaiyuan-cheng I have some additional fixes I need to push.

ewquon:
Note: After updating from main, the behavior of `wall_distance(i, j, k, grid, ::Bottom)` (recently introduced in #967) is changed and now correctly measures the distance from the domain bottom (for any terrain elevation, not necessarily `z == 0`) to the first near-wall cell center. 

I.e.,

```diff
diff --git a/src/BoundaryConditions/wall_faces.jl b/src/BoundaryConditions/wall_faces.jl
index a3e4a23d..d39be037 100644
--- a/src/BoundaryConditions/wall_faces.jl
+++ b/src/BoundaryConditions/wall_faces.jl
@@ -50,10 +50,10 @@ wall_normal_direction(::HorizontalWall) = ZDirection()
 @inline wall_value(ℓ, m, grid, ::XNormalWall,    f::Function, clock) = f(node(1, ℓ, m, grid, nothing, Center(), Center())..., clock.time)
 @inline wall_value(ℓ, m, grid, ::YNormalWall,    f::Function, clock) = f(node(ℓ, 1, m, grid, Center(), nothing, Center())..., clock.time)

-# Wall-normal distance from the wall to the near-wall cell centre: the height of the
-# first cell centre for the bottom wall, and half the cell width otherwise
-@inline wall_distance(i, j, k, grid, ::Bottom) = znode(i, j, k, grid, Center(), Center(), Center())
-@inline wall_distance(i, j, k, grid, ::Top) = Δzᶜᶜᶜ(i, j, k, grid) / 2
+# Wall-normal distance from the wall to the near-wall cell centre: half the cell width. Half the
+# first-cell thickness rather than `znode` on the bottom, so the measurement height is taken above
+# the local surface — the two agree only when the bottom face sits at z = 0.
+@inline wall_distance(i, j, k, grid, ::HorizontalWall) = Δzᶜᶜᶜ(i, j, k, grid) / 2
 @inline wall_distance(i, j, k, grid, ::XNormalWall) = Δxᶜᶜᶜ(i, j, k, grid) / 2
 @inline wall_distance(i, j, k, grid, ::YNormalWall) = Δyᶜᶜᶜ(i, j, k, grid) / 2
```

ewquon (approved):

PR #895, opened 2026-08-04
Reactant compat to `0.2.268 - 0.2.274, 0.2.278`

dkytezab:
For NumericalEarth compatibility. First trying pin to see if anything breaks on the downgraded version.


Project.toml
```diff
@@ -43,7 +43,7 @@ KernelAbstractions = "0.9"
 Oceananigans = "0.110.14"
 Printf = "1"
 RRTMGP = "0.21"
-Reactant = "0.2.278"
+Reactant = "=0.2.275"
```
giordano: The main thing is that we want to avoid 0.2.277, right?
```suggestion
Reactant = "0.2.275 - 0.2.276, 0.2.278"
```

giordano: Or
```suggestion
Reactant = "0.2.268 - 0.2.276, 0.2.278"
```
to just extend previous version. Or anyway something like that

giordano: Actually, based on https://github.com/NumericalEarth/Breeze.jl/pull/882#issue-5038255833 v0.2.275 was problematic for us, so my somewhat final suggestion is
```suggestion
Reactant = "0.2.268 - 0.2.274, 0.2.278"
```

PR #901, opened 2026-08-12
Transfer field-valued surface emissivity and albedo to RRTMGP

kaiyuan-cheng:
## What

Surface emissivity and albedo could be given as 2D `Field`s, but a field-valued property never reached the solver.

RRTMGP's surface boundary arrays `sfc_emis`, `sfc_alb_direct` and `sfc_alb_diffuse` are allocated with `undef` and stored *by reference* in `LwBCs`/`SwBCs` — RRTMGP neither copies nor validates them. Breeze filled them only on the `isa Number` branch, so passing a `Field` left the solver reading whatever the allocation happened to contain: a silently wrong albedo at best, NaN fluxes at worst. In the gray model, `update_rrtmgp_surface_properties!` was a `return nothing` stub with a TODO.

## How

`update_rrtmgp_surface_boundary_conditions!` copies all three properties from their 2D fields into RRTMGP's `(nband, ncolumn)` arrays. Scalars become `ConstantField`s so every property takes one path. It runs at construction and again before every solve, so a property that evolves (a coupled model's albedo, a snow-dependent albedo) is picked up rather than frozen.

The band loops are separate on purpose: `nband_lw = 16` but `nband_sw = 14`, so a single loop would over- or under-fill.

Also here: a scalar outside `[0, 1]` is now rejected at construction, naming the offending keyword — an albedo given in percent would otherwise reflect more shortwave than the surface receives.

## Verification

Checked against the imposed boundary conditions, not just the array contents, using a 4×4 grid with albedo varying in `x` (0.05 → 0.80) and emissivity in `y` (0.70 → 1.00):

- **Shortwave**, clear-sky and all-sky: `F↑(sfc) / F↓(sfc)` equals the specified albedo to machine precision in every column (max error 1.1e-16).
- **Longwave**, all three optics: the inferred surface source `(F↑ − (1−ε)F↓) / ε` is identical across all 16 columns — exactly `σT₀⁴ = 459.3003` for gray, and `459.2313` for the full-spectrum models, the 0.015% gap being RRTMGP's LW band coverage of the Planck integral. That is Kirchhoff's law holding per column with the right ε.

Gray shortwave is unaffected by albedo because `NoScatSWRTE` does not reflect at the surface (`F↑_sw ≡ 0` for every α). That is pre-existing and now noted in the tests, so the gray albedo assertions aren't over-read.

`test/radiative_surface_properties.jl` is new. Against the unfixed code it fails 18 of 30 assertions; with the fix the suite is 35/35.

## Tests

All run locally on CPU, Float64:

| Suite | Result |
|---|---|
| `radiative_surface_properties` (new) | 35/35 |
| `unit_tests` | pass |
| `gray_radiative_transfer` | 77/77 |
| `clear_sky_radiative_transfer` | 19/19 |
| `all_sky_radiative_transfer` | 21/21 |
| `radiation_scheduling` | 15/15 |
| `quality_assurance` (Aqua, ExplicitImports, Core.Box) | 37/37 |
| `doctests` | pass |

## Notes for review

- The second commit is a cleanup pass: it removes a duplicated range check, collapses nine near-identical scalar-wrapping blocks into `constant_field_property`, and reaches the full-spectrum solver through RRTMGP's published accessors (`RRTMGP.surface_emissivity` and friends) rather than hardcoding `solver.lws.bcs.sfc_emis`.
- All three properties are treated as spectrally grey: every band receives the same value. This is self-consistent — the specified α *is* the broadband albedo, exactly — and is stated in the function's docstring. Per-band properties would be a new method, not a rewrite.
- Not addressed here: `surface_temperature` still fills `t_sfc` from `update_rrtmgp_gas_state!` rather than through this path. That placement looks right (RRTMGP puts `t_sfc` on `AtmosphericState`, not on `bcs`), but flagging it as the one surface property that does not follow the new uniform rule.


giordano:
Can I get a review of #853 before doing further development of the extension? It's been ready for a few weeks 🥲

ext/BreezeRRTMGPExt/gray_radiative_transfer_model.jl
```diff
@@ -158,7 +157,7 @@ function AtmosphereModels.RadiativeTransferModel(grid::AbstractGrid,
         rrtmgp_T₀ .= surface_temperature.constant
     end
 
-    grid_parameters = RRTMGPGridParams(FT; context, nlay=Nz, ncol=Nc)
+    grid_parameters = RRTMGPGridParams(FT; context, domain_nlay=Nz, ncol=Nc)
```
glwagner, line 160: 🤦 

damm you RRTMGP

glwagner (approved):

PR #902, opened 2026-08-12
Fix the AIVA split time-step pairing under AcousticRungeKutta3

### LINKED ISSUE #897: Adaptive implicit vertical advection NaNs immediately on a terrain-following nested domain
Enabling adaptively implicit vertical advection (`WENO(order=5, time_discretization=AdaptiveVerticallyImplicitDiscretization(cfl=0.5))` on momentum and all scalars) in a compressible, terrain-following, ERA5-nested LAM produces `NaN` within the first ~100 outer iterations, while the identical configuration with explicit advection integrates stably for hours at the same Δt.

## Isolation matrix (12 km ERA5→Breeze nest, 540×315×50, Float32, A100)

| Advection | Boundary frame | Δt at failure | Result |
|---|---|---|---|
| explicit WENO(5) | 5-cell relax / 60 km blend | 2.5–10 s | NaN at sim-hour 2–3 (separate boundary issue) |
| explicit WENO(5) | 8-cell relax / 240 km blend | 2.5–10 s | stable 6 h |
| AIVA | 5-cell / 60 km | 11 s | NaN < 100 iterations |
| AIVA | 8-cell / 240 km | 11 s | NaN < 100 iterations |
| AIVA | 8-cell / 240 km | **capped at 10 s** | NaN < 100 iterations |

The last row is the discriminating one: AIVA fails at the same Δt the explicit scheme holds indefinitely, before any rain forms (`qʳ = 0` at the last healthy print), and independent of the frame settings. The wizard integration works as designed (`cell_advection_timescale` drops the vertical term; Δt goes straight to the horizontal limit).

## Configuration

- `nested_atmosphere_model` (NumericalEarth `glw/cleanup`) over ERA5, `TerrainFollowingVerticalDiscretization` (ETOPO2022, 60 m surface spacing, ~20 km top), open lateral boundaries + Davies relaxation, split-explicit compressible dynamics, 1-moment mixed-phase microphysics, Breeze `main` @ 0.8.0.
- Failure signature: `InexactError: Int64(NaN32)` reaching the wizard after the state NaNs; the first coupled step itself completes and prints a healthy state.
- Full driver: https://github.com/glwagner/AtmosphericRivers.jl (`downscale.jl`, `AR_AIVA=1`); happy to bisect further or produce a bare-nest MWE.

Suspected interaction: the implicit tridiagonal solve × terrain-following metrics × open-boundary halos (the AIVA tests exercise flat, closed domains). Float32 is also untested territory per the test matrix.

glwagner:
Fixes the time-step pairing of `AdaptiveImplicitVerticalAdvection` (AIVA) under `AcousticRungeKutta3`, replacing this PR's previous approach, which reviews showed introduced a conservation error.

## The bug

AIVA splits every vertical advective flux into an explicit fraction `s = min(1, cfl Δz / (|w| Δt))` and a first-order-upwind implicit remainder. The two halves only sum to one transport if both are sized with the same `Δt`. Under `AcousticRungeKutta3` they were not: Oceananigans' generic hook leaves `td.Δt[] = clock.last_Δt` — the whole outer step — at every stage, and this PR's earlier `set_advection_timestep!` made it worse by adding a second writer to the same (shared) Ref, writing mid-stage after tracer tendencies were frozen against the first value. Worked example at vertical Courant 1.2, `cfl = 0.5`, stage 1: the explicit half is scaled by `s(last_Δt) = 0.42` while the implicit remainder is sized with `s(βΔt) = 1` (no remainder at all) — 58% of the vertical moisture flux silently dropped for that stage.

## The fix

One writer, once per stage, through Oceananigans' own extension point: `update_advection_timestep!` specialized on `AcousticRungeKutta3`, mirroring the upstream `RungeKutta3TimeStepper` / `SplitRungeKuttaTimeStepper` methods. Plus:

- **`maybe_prepare_first_time_step!`** now seeds `clock.last_stage_Δt`, so stage 1 of a cold start gets the stage interval `β₁Δt` rather than the full `Δt` (a 3× overestimate of the split interval, traced on a real run).
- **A `FluxFormAdvection` disambiguation method** delegating to the z-scheme, structurally copying upstream. This fixes two CI failures at once: `Aqua.test_ambiguities` (1 ambiguity; the first `set!` on a `FluxFormAdvection`+AIVA model threw) and the would-be stale `FluxFormAdvection` import.
- Stage 1 deliberately uses the *previous* outer step's `Δt` — rebuilding the split at stage entry would desynchronize it from the tracer tendencies frozen one stage earlier, recreating the bug. Documented in the docstring; the cost is confined to CFL targeting on one stage when `Δt` changes.

Tests go from 48 to 83 assertions, including a changing-`Δt` pairing test instrumented from a real run, a cold-start test, and a `FluxFormAdvection`+AIVA smoke test. On the pre-fix code the QA suite is red and the new tests fail; on this branch: quality assurance clean, `implicit_vertical_advection.jl` 83/83, `terrain_following_split_explicit.jl` 509/509, `open_boundary_momentum.jl` 35/35.

## Second commit: freeze the stage-entry advecting state

Review of the first commit surfaced the velocity/density analog of the same pairing principle: the slow tendencies split fluxes of the *stage-entry* state, so the post-loop implicit solve must size the withheld remainder from that same state (invariant: wᴸ = wᵉ + wⁱ, identical interpolation and stage Δt). Measured on the first commit, the velocity half held only by staleness (nothing refreshes w mid-stage — today), while the density half was a real error: `_recover_full_state!` rewrites `dynamics.dry_density` before `implicit_substep!` reads it. The substepper now caches the full stage-entry wᴸ (not the clipped wᵉ, which loses the remainder in saturated cells) and its carrier density, filled right after `prepare_acoustic_cache!`; the solve reads the frozen cache. Closure-only configurations are bit-identical (no cache allocated), and closure diffusion always acts on the full field. Tests: 83 → 91 assertions, including cache immunity to mid-stage live-field corruption and the split identity at forced saturation with a no-saturation guard. Re-measured after the fix: the stratified mountain-wave gate is unchanged to four digits — this is a correctness invariant, not a performance change, and it eliminates "broken partition" as an alternative explanation for the accuracy-limited negative result recorded in #897.

## What this PR deliberately does not do

It makes AIVA's split *correct*, not *beneficial*. Extensive stability and accuracy testing against converged references (issue #897 has the full record) found no configuration in which AIVA extends the usable time step of the compressible core: where the split engages, first-order upwind on the remainder destroys the accuracy of exactly the strong-|w| features that triggered it, and it does not extend the stable range. The earlier reproducer under `validation/` is dropped — its uniform-θ setup could not exercise the split, and its top rung was horizontally CFL-limited.

Three pre-existing AIVA bugs found during review will be filed separately with reproducers: bounds-preserving WENO + AIVA double-counts the vertical flux; precipitating species lose sedimentation flux from the implicit half; and a `field_index` off-by-one for scalars on the acoustic path.


glwagner:
~~⚠️ Do not merge yet~~ — **RETRACTED / SUPERSEDED (2026-08-19).** This warning applied to the PR's original `set_advection_timestep!` approach, whose conservation defect it describes. The PR has since been rewritten from scratch (see the current description): the mid-stage second writer is gone, the split time step is set through Oceananigans' own `update_advection_timestep!` hook, and the desynchronization this comment warned about is fixed by construction and covered by tests that fail on the pre-fix code. Kept for the historical record; the analysis below describes code that no longer exists on this branch.

test/runtests.jl
```diff
@@ -25,9 +25,10 @@ if filter_tests!(testsuite, args)
     end
 end
 
-if Sys.isapple() && get(ENV, "GITHUB_ACTIONS", "false") == "true"
+if (Sys.isapple() || Sys.islinux()) && get(ENV, "GITHUB_ACTIONS", "false") == "true"
     GC.gc(true); GC.gc(false); GC.gc(true)
-    # macOS runners have little memory compared to the other runners, let's set a more
+    # macOS and ubuntu runners have little memory (earlyoom kills workers on the ~13 GiB
+    # ubuntu machines — Malt.TerminatedWorkerException); set a more
```
giordano: This is wrong

1. the cap on macOS is sensibly lower because the memory available on the macOS runners is half of that on the other runners
2. memory usage on macOS tends also to be less than on other systems, so that thresholds have to be different
3. the computation below has a division by 2 because there are workers (forced from the outside), so you're ending up with a ***larger*** memory threshold now: [5.2 GiB](https://github.com/NumericalEarth/Breeze.jl/actions/runs/32312772331/job/96258999985?pr=902#step:13:407), compared to the default of 3.8 GiB

giordano: Instead of doing this, I was hoping to use the new (as of today) ability to [automatically retry failed tests](https://github.com/JuliaTesting/ParallelTestRunner.jl/pull/157), but I want first to be able to [select which tests are known to be flaky](https://github.com/JuliaTesting/ParallelTestRunner.jl/pull/157#pullrequestreview-4900982346) (which is not implemented yet)

giordano: On GPU runners the [memory threshold is now 23 GiB](https://github.com/NumericalEarth/Breeze.jl/actions/runs/32312772331/job/96259000136?pr=902#step:14:413), which [leads to an OOM](https://github.com/NumericalEarth/Breeze.jl/actions/runs/32312772331/job/96259000136?pr=902#step:14:480)

giordano: Only alternative I can suggest right now is to reduce the number of workers for the ubuntu job with Julia min (we don't see this issue with Julia v1.12) to 3, instead of the default 4: #919

glwagner: ok :-/

glwagner: You're right on all three counts — apologies for the noise. That commit (afa82210) produced a *larger* threshold (5.2 GiB) exactly as you describe, and it flipped the previously-green GPU-L4 jobs to `Malt.TerminatedWorkerException`. I reached the same conclusion from the CI evidence and superseded it in 997a6d0a before seeing this comment:

- Runners with ≥ 20 GiB (including the GPU-L4 machines) are back to the defaults, untouched.
- Only the small (~13 GiB) ubuntu runners get a **fixed 2500 MB** cap — deliberately *below* the 3.8 GiB default — because `earlyoom` was SIGTERMing workers there even with the defaults (both `Julia min` and `1.12.6` ubuntu failed that way in [run 32300369770](https://github.com/NumericalEarth/Breeze.jl/actions/runs/32300369770) before any runtests change of mine existed; this branch adds ~35 assertions to `implicit_vertical_advection.jl`, which plausibly nudged the peak).

The full matrix is green with that in place. That said, this is your machinery — if you'd rather drop the Linux clause entirely and handle the small-runner pressure differently (or you think the default should hold and the two failures were coincidence), I'll happily revert to macOS-only; the substantive content of this PR doesn't depend on it.

src/TimeSteppers/acoustic_substep_helpers.jl
```diff
@@ -218,6 +218,39 @@ end
...
+
+# The advective coefficient state for the post-loop implicit solve: the frozen stage-entry
+# cache when advection needs it, the live fields otherwise (closure-only solves keep their
+# current behavior exactly).
+advecting_state(model) =
+    advecting_state(model.timestepper.substepper.advecting_vertical_velocity_cache,
+                    model.timestepper.substepper.advecting_density_cache, model)
+
+advecting_state(::Nothing, ::Nothing, model) =
+    (advecting_vertical_velocity(model.dynamics, model.velocities), dynamics_density(model.dynamics))
```
glwagner:
```suggestion
advecting_state(::Nothing, ::Nothing, model) = (advecting_vertical_velocity(model.dynamics, model.velocities), dynamics_density(model.dynamics))
```

test/runtests.jl
```diff
@@ -33,6 +33,11 @@ if Sys.isapple() && get(ENV, "GITHUB_ACTIONS", "false") == "true"
     # currently available memory (with a ~20% margin), with a lower bound of 1700 MiB.
     max_rss_memory = max(1_700, round(Int, available_memory() / 2 ^ 20 / 2  * 0.8))
     ENV["JULIA_TEST_MAXRSS_MB"] = string(max_rss_memory)
+elseif Sys.islinux() && get(ENV, "GITHUB_ACTIONS", "false") == "true" && available_memory() < 20 * 2^30
+    # The small (~13 GiB) ubuntu runners run earlyoom, which SIGTERMs workers once several
+    # approach 4 GiB at once (Malt.TerminatedWorkerException). A LOW fixed cap makes workers
+    # recycle gracefully well before that; the large-memory GPU runners are excluded.
+    ENV["JULIA_TEST_MAXRSS_MB"] = "2500"
```
glwagner, line 39: @giordano is this ok?

giordano: This is definitely better than the previous attempt, not sure this will always avoid the issue (the turbulence closures test uses lots of memory on its own on v1.11), I guess we'll see

src/TimeSteppers/acoustic_runge_kutta_3.jl
```diff
@@ -108,6 +111,49 @@ function AcousticRungeKutta3(grid, prognostic_fields;
...
+stage, so the explicit velocity fraction frozen into `Gⁿ` pairs with the implicit fraction
+the next stage applies. One writer, once per stage, read by every field — mirrors
+Oceananigans' `RungeKutta3TimeStepper` and `SplitRungeKuttaTimeStepper` specializations.
+
+Stage 1 of step ``n`` is written before ``Δtₙ`` is known and deliberately uses ``β₁ Δtₙ₋₁``:
+rewriting the split at stage entry would desynchronize it from tendencies frozen against the
+earlier value. The cost is confined to CFL targeting on one stage when ``Δt`` changes.
+"""
+@inline function Oceananigans.Advection.update_advection_timestep!(a::AdaptiveImplicitVerticalAdvection,
+                                                                   timestepper::AcousticRungeKutta3, clock)
```
glwagner:
```suggestion
@inline function Oceananigans.Advection.update_advection_timestep!(a::AdaptiveImplicitVerticalAdvection, timestepper::AcousticRungeKutta3, clock)
```

src/CompressibleEquations/acoustic_substepping.jl
```diff
@@ -125,6 +128,10 @@ struct AcousticSubstepper{N, FT, D, AD, US, CF, MP, TAV, GT, TS, WC, DC}
 
     advecting_vertical_velocity_cache :: WC  # stage-entry w for the implicit advective remainder
     advecting_density_cache :: DC            # stage-entry ρᵈ; `nothing` without adaptive-implicit advection
+
+    # The time-averaged transport velocity that the moisture and tracer tendencies in `Gⁿ` were
+    # built with, frozen before the next acoustic loop overwrites `time_averaged_velocities`.
+    transport_vertical_velocity_cache :: TWC
```
glwagner: what's the difference between this and the `advecting_vertical_velocity_cache` ?

kaiyuan-cheng: They represent different velocities at different time levels:

- `advecting_vertical_velocity_cache`: the stage-entry predictor velocity. It is cached before `compute_slow_*_tendencies!` and pairs with the implicit remainder of prognostic vertical advection for momentum and thermodynamics.

- `transport_vertical_velocity_cache`: the acoustic-substep-averaged transport velocity. It is cached after scalar tendencies are built and pairs with the implicit remainder of moisture/tracer vertical advection.

glwagner: Let's improve the terminology (one is time-averaged, the other is not, so the name should be something like `vertical_velocity_cache` and `time_averaged_vertical_velocity_cache`). I wonder if there is a better way to write this algorithm that would avoid the need for this split (I think WRF does this slightly differently)

PR #907, opened 2026-08-14
[CI] Run with Julia v1.13

giordano:
Not to be merged, I want to see where we are with Julia v1.13 compatibility.


giordano:
Exciting news: all tests are passing, including the Reactant ones.  And tests are slightly faster than in v1.12 (precompilation also appears to be quite a bit faster).  CC @Pangoraw @wsmoses

giordano:
This PR served its testing purpose, so it can be closed.

PR #909, opened 2026-08-17
Check in successful manifest

giordano:
Use same manifest as #878, let's see if this is successful.

PR #908, opened 2026-08-15
Revert "Bump CloudMicrophysics in /"

giordano:
The full docs build in #878 worked, but now on main we're getting NaNs.  Let's see if reverting that PR resolves the issue (and if not, we'll need to find the culprit)

PR #910, opened 2026-08-17
Temporarily pin LLVM to v9.11

giordano:
While I investigate the [issue with NaNs when building the docs](https://github.com/NumericalEarth/Breeze.jl/pull/908#issue-5159079297), I suggest pinning LLVM to the [last known working version](https://github.com/NumericalEarth/Breeze.jl/pull/909#issuecomment-5317390014).


giordano:
This isn't the issue either, sigh.  The bot suggests that the boundary conditions https://github.com/NumericalEarth/Breeze.jl/blob/c713975970323d88dad39c6714d175b9ae0dfc43/examples/tropical_cyclone_with_rainband.jl#L325-L327 are bogus and `bottom`/`top` are set to uninitialised data, which may have spurious (and non-reproducible) NaNs.

PR #912, opened 2026-08-18
[test] Temporarily pin Reactant to v0.2.278

giordano:
As mentioned in https://github.com/NumericalEarth/Breeze.jl/pull/908#issuecomment-5302142248, Reactant v0.2.279 causes a stack overflow error in `lat_lon_compilation.jl` on Julia v1.11.  CC @Pangoraw.


Pangoraw (approved):
I will investigate;

PR #917, opened 2026-08-18
Fix aliased base-state reads in the acoustic substepper

### LINKED ISSUE #906: Aliased model.momentum in _recover_full_state! makes terrain-following runs non-reproducible on GPU and threaded CPU
`_recover_full_state!` is launched with `model.momentum` as both output and input
(`src/CompressibleEquations/acoustic_substepping.jl:1530-1541`). The comment at `:1526-1528` states
this is safe because all reads are grid-point-local. **That is false for `TerrainCompressibleDynamics`**
(`src/CompressibleEquations/terrain_compressible_physics.jl:469-482`), which reads the aliased arrays
at neighboring points.

The consequence is that terrain-following compressible integrations are **not bitwise reproducible** —
on GPU, and on CPU with more than one thread. Confirmed causally on both backends: removing only the
aliasing makes previously non-deterministic configurations bitwise identical, while unpatched controls
in the same job on the same hardware still diverge at step 1. Measured cost of the fix is under 1%.

## The aliasing and the claimed invariant

`src/CompressibleEquations/acoustic_substepping.jl:1530-1541` passes `model.momentum` as output `m`
and `model.momentum.ρu/ρv/ρw` as inputs `ρuᴸ/ρvᴸ/ρwᴸ` — the same arrays. The kernel writes
`m.ρu[i,j,k]` at `:1250-1252`. The justification at `:1526-1528`:

> the recovery kernel reads them as Uᴸ AND writes the full state back to the same fields — per-thread
> read-before-write makes this aliasing safe because all reads are local to the same grid point.

True for the height-coordinate method (`:1256-1257`, purely `[i,j,k]`). False for the terrain method
(`src/CompressibleEquations/terrain_compressible_physics.jl:469-482`), which forms
`ℑzᵃᵃᶠ(ℑxᶜᵃᵃ, total_momentum, ρuᴸ, ρu′)` — reading `ρuᴸ` at `(i-1,j,k)`, `(i,j,k-1)`, `(i-1,j,k-1)`
while another thread writes `m.ρu[i-1,j,k]`. Both interpolants are multiplied by `slope_x`/`slope_y`,
so the effect vanishes at zero slope.

## Measured: non-reproducibility

NVIDIA T4, Breeze 0.8.0 + Oceananigans 0.110.14 + CUDA 6.2.1, Julia 1.12.6. 96x96x48, 200 steps,
fixed `Δt = 2 s`, identical seeded initial condition, separate processes on the same GPU.

| case | runs | result |
|---|---|---|
| **terrain (slopes ≈ 0.2)** | 4 | **not bitwise**; diverge at **step 1**. Final max\|Δ\|: ρu ≈2.0e-12, ρw ≈2.3e-12. ~442 k of 442368 points differ |
| flat | 3 | bitwise identical, all 201 steps |
| TFVD with `h ≡ 0` | 3 | bitwise identical, all 201 steps |

The `h ≡ 0` control is the discriminator: the terrain-following coordinate machinery is fully active
and the run reproduces exactly, so the non-determinism tracks the non-zero slope factor.

**A small test reports a false pass.** The same terrain case at 32x32x16 is bitwise identical over 300
steps — at 16 k points the kernel occupies too few blocks for scheduling to resolve differently. A
regression test needs to be large enough to expose this.

CPU is reproducible single-threaded and **not** reproducible at 8 threads (first divergence at step 1),
with or without real cores per thread.

**An adaptive `Δt` propagates it into the model clock.** With
`conjure_time_step_wizard!(…; cfl = 0.5)`, the 96x96x48 terrain case diverges in the `Δt` sequence at
step 17 — the first CFL-limited step — `9.97144500561715752` vs `9.97144500561693903` s. 184 of 201
steps then carry a different `Δt`, and final model time differs by 2.7e-11 s. At 32x32x16, where no
seed exists, the wizard reproduces exactly, so it is an amplifier and not a source.

## Why the per-call magnitude is small

A CPU kernel-level comparison (shipped launch vs the same launch given deep copies) differs by exactly
**one ulp**, because the `ρuᴸ/ρvᴸ` dependence cancels: `:475` contributes
`+slope_x · ℑzᵃᵃᶠ∘ℑxᶜᵃᵃ[ρuᴸ + ρu′]` while `:477-478` → `:459` contributes
`−slope_x · ℑzᵃᵃᶠ∘ℑxᶜᵃᵃ[ρuᴸ]` — same stencil, opposite sign. Corrupting `ρuᴸ` by 30x the field's own
amplitude moved `ρw` by only 2.16e-15.

That cancellation is exact algebraically but not in floating point, so what survives is an
order-dependent rounding difference. It reseeds on every run and compounds through the integration —
hence a one-ulp kernel effect and a macroscopically divergent trajectory.

A positive control confirms the clobbering is real: the bare `ℑzᵃᵃᶠ∘ℑxᶜᵃᵃ∘total_momentum` stencil,
aliased vs deep-copied with no compensating term, differs by 4.5e-01 to 6.8e-01 over half the cells.

## Confirmed: removing the aliasing removes the non-determinism

A `Pkg.develop`ed copy of Breeze 0.8.0 with **one file changed** — persistent scratch fields filled via
`parent(dst) .= parent(src)` for `dry_density`, `ρᵡ`, `ρu`, `ρv`, `ρw` immediately before the launch,
passed as the `ᴸ` arguments. Nothing else touched. Same 96x96x48 terrain case, same job, same GPU:

| comparison | bitwise | first divergence |
|---|---|---|
| **patched a vs b** | **YES**, max\|Δ\| = 0 | none, all 201 rows identical |
| **patched a vs c** | **YES**, max\|Δ\| = 0 | none, all 201 rows identical |
| unpatched a vs b (control, same job) | NO | step 1 |
| patched a vs unpatched a | NO | step 1 (builds genuinely differ) |

The same patch also fixes the threaded CPU backend. 32x32x16 terrain, 300 steps,
`JULIA_NUM_THREADS=8` with 8 real cores, patched and unpatched interleaved in one job:

| comparison | bitwise | first divergence |
|---|---|---|
| **patched a vs b** (8 threads) | **YES**, max\|Δ\| = 0 | none, all 301 rows identical |
| unpatched a vs b (control, same cores) | NO | step 1; final max\|Δ\| 1.6e-12 (ρu, ρw) |

So the multi-threaded CPU spread and the GPU non-determinism are the same defect, not two.

## Cost

GPU, 200 steps at 96x96x48 on an uncontended T4, same job: patched 51.360 / 51.213 / 53.415 s (mean
**51.996 s**), unpatched 51.078 / 52.301 s (mean **51.690 s**) — **+0.6%, inside the run-to-run
spread**, since the unpatched pair alone spans 2.4%. Consistent with the arithmetic: ~156 MB/step of
extra device traffic against ~300 GB/s is ~0.2% of a 257 ms step.

CPU, 300 steps at 32x32x16, 8 threads: patched 51.478 / 51.334 s (mean **51.406 s**), unpatched
51.223 / 51.488 s (mean **51.355 s**) — **+0.1%**, also inside noise.

Under ~1% on both backends. A production version owning the scratch on the substepper would avoid the
first-call allocation.

## Suggested fix

Un-alias the inputs as above. If the aliasing is worth keeping for memory reasons, then at minimum
correct the comment to state the actual invariant — that the `Uᴸ` dependence of
`acoustic_recovered_vertical_momentum` must cancel exactly against `transport_ρw` — and add a
regression test at a size that exposes the effect, so the next change to the terrain recovery fails
loudly rather than silently.

## Caveats

One GPU architecture (T4), single GPU, and one CPU thread count (8) throughout. The kernel-level
aliased-vs-copied comparison isolates the launch rather than going through `time_step!`. Configuration
is a standalone compressible dycore — Periodic/Periodic/Bounded, `LinearDecay` TFVD,
`SplitExplicitTimeDiscretization`, `Centered(order=2)`, corrugated bell terrain — not a nested or
multi-GPU setup.

## The experimental patch

Experiment-grade — it parks the scratch in a module-level `Ref` so the measurement isolates the copies
rather than repeated allocation. A production fix should own these fields on the substepper. Included
so the measurements above are reproducible.

<details>
<summary>diff against Breeze 0.8.0</summary>

```diff
--- /shared/home/eliot/.julia_t4/packages/Breeze/rqttH/src/CompressibleEquations/acoustic_substepping.jl	2026-08-13 17:43:19.000000000 +0000
+++ /shared/home/eliot/AeFS/Breeze-0.8.0-aliasfix/src/CompressibleEquations/acoustic_substepping.jl	2026-08-13 19:27:59.000000000 +0000
@@ -1232,6 +1232,27 @@
 # kernel writes ρ can read a GPU-scheduling-dependent mix of old and new
 # neighbor values. The driver calls AtmosphereModels.compute_velocities! after
 # recovery and halo fill.
+# ALIASING FIX (experiment): persistent scratch holding a snapshot of the stage-entry Uᴸ base
+# (ρᴸ, ρθᴸ, ρuᴸ, ρvᴸ, ρwᴸ). Allocated on first use and reused thereafter, so the measured cost is
+# the copies only, not repeated allocation. A production fix would own these on the substepper.
+const RECOVERY_BASE_SCRATCH = Ref{Any}(nothing)
+
+function unaliased_recovery_base(ρ, ρθ, m)
+    s = RECOVERY_BASE_SCRATCH[]
+    if s === nothing || size(parent(s.ρ)) != size(parent(ρ)) ||
+       typeof(parent(s.ρ)) != typeof(parent(ρ)) || size(parent(s.ρw)) != size(parent(m.ρw))
+        s = (ρ  = similar(ρ),  ρθ = similar(ρθ),
+             ρu = similar(m.ρu), ρv = similar(m.ρv), ρw = similar(m.ρw))
+        RECOVERY_BASE_SCRATCH[] = s
+    end
+    parent(s.ρ)  .= parent(ρ)
+    parent(s.ρθ) .= parent(ρθ)
+    parent(s.ρu) .= parent(m.ρu)
+    parent(s.ρv) .= parent(m.ρv)
+    parent(s.ρw) .= parent(m.ρw)
+    return s
+end
+
 @kernel function _recover_full_state!(ρ, ρθ, m,
                                       ρ′, ρθ′, ρu′, ρv′, ρw′,
                                       ρᴸ, ρuᴸ, ρvᴸ, ρwᴸ, ρθᴸ,
@@ -1527,6 +1548,10 @@
     # same fields — per-thread read-before-write makes this aliasing safe
     # because all reads are local to the same grid point.
     ρᵡ = thermodynamic_density(model.formulation)
+    # ALIASING FIX (experiment): snapshot the Uᴸ base into scratch so the kernel cannot read
+    # what it writes. On terrain grids `acoustic_recovered_vertical_momentum` interpolates
+    # ρuᴸ/ρvᴸ at NEIGHBORING points, so the in-place aliasing is not read-before-write-local.
+    ᴸ = unaliased_recovery_base(model.dynamics.dry_density, ρᵡ, model.momentum)
     launch!(arch, grid, :xyz, _recover_full_state!,
             model.dynamics.dry_density, ρᵡ,
             model.momentum,
@@ -1535,9 +1560,9 @@
             substepper.momentum_perturbation.u,
             substepper.momentum_perturbation.v,
             substepper.momentum_perturbation.w,
-            model.dynamics.dry_density,
-            model.momentum.ρu, model.momentum.ρv, model.momentum.ρw,
-            ρᵡ,
+            ᴸ.ρ,
+            ᴸ.ρu, ᴸ.ρv, ᴸ.ρw,
+            ᴸ.ρθ,
             grid, model.dynamics)
 
     # Thread clock + model fields so time-dependent Open BCs on the recovered
```

</details>

kaiyuan-cheng:
Fixes #906.

Two substep kernels are launched with a field as both input and output while reading that field across neighbouring grid points, so a thread can read a value another thread has already overwritten. #906 documents the first as a bitwise-reproducibility failure on GPU and on multi-threaded CPU; this PR fixes it, and fixes a second instance of the same class that the issue does not cover.

## 1. Terrain recovery no longer reads the aliased base momenta

`_recover_full_state!` is launched with `model.momentum` as output and `model.momentum.ρu/ρv/ρw` as the stage-base inputs. The height-coordinate recovery reads only `[i, j, k]`, so the read-before-write justification holds there. The terrain-following method did not: it interpolated `ρuᴸ`/`ρvᴸ` across neighbours while other threads wrote them.

As #906 works out, that dependence cancels. `ℑ` is linear, so converting the substepped contravariant momentum back to Cartesian ρw gives

```
ρw = (ρw̃ᴸ + ρw̃′) + slopeₓ ℑ(ρuᴸ + ρu′) + slopeᵧ ℑ(ρvᴸ + ρv′)
   = (ρwᴸ − slopeₓ ℑρuᴸ − slopeᵧ ℑρvᴸ) + ρw̃′ + slopeₓ ℑ(ρuᴸ + ρu′) + slopeᵧ ℑ(ρvᴸ + ρv′)
   = ρwᴸ + ρw̃′ + slopeₓ ℑρu′ + slopeᵧ ℑρv′
```

with `ρw̃ᴸ` = `terrain_vertical_transport_momentum` of the same stage-entry momenta that `_initialize_terrain_vertical_momentum_perturbation!` used as the base for `ρw̃′`. So `ρuᴸ` and `ρvᴸ` are no longer arguments at all, and the base ρw arrives as a value rather than a field:

```julia
acoustic_recovered_vertical_momentum(i, j, k, grid, dynamics, ρwᴸ_ccf, ρu′, ρv′, ρw̃′)
```

No aliased field is in reach of a stencil in either method, so this cannot be reintroduced inside the function; a vertical coordinate that genuinely needs `ρwᴸ` neighbours has to change the signature first.

The rewritten grouping also drops a catastrophic cancellation the old one carried for large base momenta, since it no longer adds `+slopeₓ ℑ(ρuᴸ + ρu′)` and subtracts `−slopeₓ ℑρuᴸ`.

## 2. The vertical solve no longer writes over its own `(ρw)′` input

Not covered by #906. `_build_vertical_rhs!` was launched with `substepper.momentum_perturbation.w` as both the destination and the `(ρw)′` source, while the implicit vertical-damping term reads

```julia
∂z²_ρw′ˢ⁻ = ∂zᶜᶜᶠ(i, j, k, grid, ∂zᶜᶜᶜ, ρw′)   # faces k-1, k, k+1
```

The right-hand side now lands in a dedicated `vertical_solver_source_term` z-face field, which the tridiagonal solve reads while still writing its solution into `momentum_perturbation.w`. I checked that no z-face scratch was free at that point in the loop (`time_averaged_velocities.w` accumulates across substeps, `slow_vertical_momentum_tendency` is read inside this very kernel, the solver's `scratch` is `Nz`-sized Thomas storage), and that all three `get_coefficient` methods read only `γRᵐᴸ`, `Πᴸ`, `θᴸ` and grid metrics, never the destination, so moving off the in-place `solve!` changes nothing but the source of the right-hand side. Cost is one z-face `FT` array, measured below.

Inert under the default `damp_vertical = false`, where `dˢ⁻ = 0` zeroes the term.

## Verification

To be clear about what was and was not measured: the GPU and multi-threaded non-determinism itself is measured in #906 and was not re-reproduced here. What is verified below is the algebra, the CPU-level behaviour, and that both suites pass on CPU and on an H100.

- **On single-threaded CPU the fix is bit-identical to unpatched `main`.** Ten split-explicit steps over 300 m Gaussian terrain, plus a flat-terrain-following versus height-coordinate comparison, give exactly the same numbers with and without the patch (`max|ρw|` = 1.131864e+00; relative deviations 1.568e-16 / 3.988e-12 / 9.797e-17 for ρu / ρw / ρ). That is the expected signature of a pure race fix, and it is why the recovery race cannot be demonstrated here: within one thread both `ℑ(ρuᴸ + ρu′)` and `ℑ(ρuᴸ)` read the same values at the same instant, so the cancellation stays exact even when those values are already corrupted. It only breaks when another thread writes between the two reads, matching #906 finding CPU reproducible single-threaded and not reproducible at 8 threads.
- **The rewritten recovery is the old formula.** Max `|old − new|` = **2.2e-16** over every z-face point on an 8×8×6 grid with `h = 200 sin(2πx/Lx) sin(2πy/Ly)` and a smooth O(10) base state. Consistent with the one-ulp per-call magnitude reported in #906.
- **The source-term alias was changing results.** A/B on the real kernel from one frozen input state, one run writing to a separate destination and one writing in place: max `|Δ|` = 5.69e-07 against scale 8.47e-05, i.e. **6.7e-3 relative**, with `dˢ⁻` = 6.57e+03 under `damp_vertical = true`. Visible single-threaded because the corrupted neighbour enters through a single stencil with no cancelling partner.

| suite | CPU | H100 80GB HBM3 |
|---|---|---|
| `test/acoustic_substepping.jl` | 199/199, exit 0 | 199/199, exit 0 |
| `test/terrain_following_split_explicit.jl` | 512/512, exit 0 | 512/512, exit 0 |
| `test/quality_assurance.jl` | 37/37, exit 0 (Aqua 31, ExplicitImports 5) | — |

Single-threaded CPU. The `damp_vertical = true` terrain testsets pass on the corrected numerics; their assertions are `isfinite` and `≈ 0` with tolerance rather than pinned values.

Unrelated observation while baselining: on `main` the flat-terrain-following versus height-coordinate ρw agreement is 3.99e-12 rather than exact. Present with and without this patch, so out of scope here, but possibly worth its own look.

## Cost of the extra field

Measured on an H100 80GB HBM3, `Float64`, `damp_vertical = true`. `_build_vertical_rhs!` and `solve!` are timed as a pair with CUDA events, 50 trials, the same input restored (untimed) before each trial. The only difference between variants is whether the source-term store and the solver's source-term load hit the same array as the solve target.

| grid | pair, separate | pair, in place | delta | full `time_step!` | delta/step | extra memory |
|---|---|---|---|---|---|---|
| 128×128×64 | 0.182 ms | 0.155 ms | +26.6 µs (+17.1%) | 9.10 ms | +3.51% | 10.9 MiB |
| 256×256×64 | 0.365 ms | 0.350 ms | +14.8 µs (+4.2%) | 22.48 ms | +0.79% | 40.5 MiB |
| 512×512×64 | 1.563 ms | 1.416 ms | +147 µs (+10.4%) | 76.69 ms | +2.31% | 155.9 MiB |

The non-monotonicity is L2: at 256² one z-face `Float64` array is 33.5 MB and fits in the H100's 50 MB L2, so the second stream is nearly free; at 512² it is 134 MB and does not, and the 147 µs matches roughly 268 MB of extra traffic at ~3 TB/s. The 128² figure is launch overhead on a 0.155 ms base rather than bandwidth. So the production cost is about 2% of a time step and 156 MB at 512×512×64. Single-threaded CPU is 0.7% to 1.1% of a step at 64³ and 128×128×64.

Open question for review. That cost is paid unconditionally, while it only buys something when the implicit vertical damping is active: `∂z²(ρw)′` is the only stencil read of `(ρw)′` in the kernel (`sponge_rhs` and the `ρw′[i, j, k]` term are point-local) and it is scaled by `dˢ⁻`, which is exactly zero under the default `damp_vertical = false`. Allocating the field only when the damping folds into the tridiagonal system, and aliasing `momentum_perturbation.w` otherwise, would be type-stable and would make the default path free. This PR does not do that, because it makes safety contingent on a configuration flag: a stencil added to this kernel later would silently reintroduce the race.

## Tests

- `test/terrain_following_split_explicit.jl`: pins the cancellation against an analytic value, `ρwᴸ + ρw̃′ + slopeₓ ℑρu′ + slopeᵧ ℑρv′`, at an interior point with both slopes asserted non-zero. The old formula and a slope-free formula both fail it.
- `test/acoustic_substepping.jl`: asserts `vertical_solver_source_term` and `momentum_perturbation.w` are distinct arrays, that `dˢ⁻ != 0` so the stencil is actually exercised, and behaviourally that the production build leaves `(ρw)′` bit-identical.

## Review follow-ups

- `vertical_momentum_rhs` renamed to `vertical_solver_source_term`. The old name read as the slow tendency `Gˢρw` rather than the right-hand side of the tridiagonal system; `source_term` follows `FourierTridiagonalPoissonSolver`, whose `source_term` field the anelastic pressure solve already uses.
- `promote_to_grid_precision` removed. It preserved the `FT` accumulation the old `ρuᴸ + ρu′` stencil got for free from the base momentum, but that is about one ulp of a value already stored at `substep_floattype`, and a no-op at the default `substep_floattype = eltype(grid)`. With a reduced `substep_floattype` the perturbation average now rounds at storage precision.
- Aliasing comments and the issue references they carried are gone from the struct definition and from `_build_vertical_rhs!`.


src/CompressibleEquations/acoustic_substepping.jl
```diff
@@ -211,6 +218,7 @@ function AcousticSubstepper(grid, split_explicit::SplitExplicitTimeDiscretizatio
                                 w = ZFaceField(grid))
 
     slow_vertical_momentum_tendency = ZFaceField(grid)
+    vertical_momentum_rhs = ZFaceField(grid) # RHS stays FT, like the (ρw)′ solve target
```
glwagner: what is the impact on performance?

src/CompressibleEquations/acoustic_substepping.jl
```diff
@@ -884,6 +893,10 @@ end
 # derivation (15). `dˢ⁻ = (1−ω) α Δz²` is the explicit half of the implicit vertical
 # damping (0 when damping off). Boundary rows: f[1] = 0 (matches b[1] = 1 ⇒ (ρw)′[1] = 0);
 # top face Nz+1 lives outside the solver (impenetrability w(top) = 0).
+#
+# `ρw′_rhs` MUST NOT alias `ρw′`: the damping term reads (ρw)′ at faces k−1, k, k+1, so writing
+# the RHS over (ρw)′ would let a thread read a neighbour another thread has already overwritten
+# (issue #906). The driver passes `substepper.vertical_momentum_rhs` for exactly that reason.
```
glwagner: I don't think we should reference specific issues in comments

src/CompressibleEquations/acoustic_substepping.jl
```diff
@@ -118,6 +121,9 @@ struct AcousticSubstepper{N, FT, D, AD, US, CF, MP, TAV, GT, TS}
     time_averaged_velocities :: TAV
 
     slow_vertical_momentum_tendency :: GT
+    # Explicit RHS of the (ρw)′ tridiag system. Kept out of `momentum_perturbation.w` so the
+    # implicit-damping stencil `∂z²(ρw)′` reads unwritten neighbours (issue #906).
+    vertical_momentum_rhs :: GT
```
glwagner: Let's brainstorm a good name for this. At first I thought it referred to the tendency `Gw`. But it is actually the rhs of the tridiagonal system.

Also, we need to remove comments from the struct definition

src/CompressibleEquations/acoustic_substepping.jl
glwagner: the whole comment may not be necessary either. There are plenty of places where aliasing is not allowed -- this is only one of them (the comment merely reflects a historical bug, but is probalby not helpful to future developers)

src/CompressibleEquations/terrain_compressible_physics.jl
```diff
@@ -466,23 +466,34 @@ end
...
-    ρvᶜᶜᶠ = ℑzᵃᵃᶠ(i, j, k, grid, ℑyᵃᶜᵃ, total_momentum, ρvᴸ, ρv′)
-    ρw̃_stage = transport_ρw(i, j, k, grid, dynamics,
-                                                           ρuᴸ, ρvᴸ, ρwᴸ)
-    @inbounds ρw̃ᵐ⁺ = ρw̃_stage + ρw̃′[i, j, k]
+    ρu′ᶜᶜᶠ = ℑzᵃᵃᶠ(i, j, k, grid, ℑxᶜᵃᵃ, promote_to_grid_precision, ρu′)
+    ρv′ᶜᶜᶠ = ℑzᵃᵃᶠ(i, j, k, grid, ℑyᵃᶜᵃ, promote_to_grid_precision, ρv′)
 
-    return ρw̃ᵐ⁺ + slope_x * ρuᶜᶜᶠ + slope_y * ρvᶜᶜᶠ
+    @inbounds return (ρwᴸ_ccf + ρw̃′[i, j, k] +
+                      slope_x * ρu′ᶜᶜᶠ + slope_y * ρv′ᶜᶜᶠ)
```
glwagner:
```suggestion
    @inbounds return (ρwᴸ_ccf + ρw̃′[i, j, k] + slope_x * ρu′ᶜᶜᶠ + slope_y * ρv′ᶜᶜᶠ)
```

src/CompressibleEquations/terrain_compressible_physics.jl
```diff
@@ -466,23 +466,34 @@ end
                                                ρu_stage, ρv_stage, ρw_stage)
 end
 
+# Promote a `substep_floattype` perturbation to the grid float type before it enters an
+# interpolation stencil, so the recovered primary prognostic is accumulated in `FT`.
+@inline promote_to_grid_precision(i, j, k, grid, ϕ) = @inbounds convert(eltype(grid), ϕ[i, j, k])
```
glwagner: why is this needed? Is there a bug somewhere?

src/CompressibleEquations/acoustic_substepping.jl
```diff
@@ -79,6 +79,9 @@ Fields:
   chemistry/TKE); the slow ρθ tendency uses the current RK predictor velocity instead, not this cache.
 - `slow_vertical_momentum_tendency` (Gˢρw, z-faces): advection+Coriolis+closure+forcing (PGF/buoyancy
   excluded — those are in the fast operator).
+- `vertical_momentum_rhs` (z-faces): explicit RHS of the (ρw)′ tridiagonal system. A separate field
+  rather than in-place in `momentum_perturbation.w`, because the implicit-vertical-damping term reads
+  (ρw)′ across neighbouring faces (issue \\#906).
```
```suggestion
- `vertical_momentum_rhs` (z-faces): explicit RHS of the (ρw)′ tridiagonal system.
```

src/CompressibleEquations/terrain_compressible_physics.jl
kaiyuan-cheng: No bug. It was preserving a precision property the old form got. It can be removed.

src/CompressibleEquations/acoustic_substepping.jl
kaiyuan-cheng: Renamed to vertical_solver_source_term.

src/CompressibleEquations/acoustic_substepping.jl
kaiyuan-cheng: ~2% more cost

src/CompressibleEquations/terrain_compressible_physics.jl
glwagner: ok yeah, was curious why `ϕ` would not have the correct precision

src/CompressibleEquations/acoustic_substepping.jl
glwagner: 😭

glwagner: it's ok, we'll recover it elsewhere!

glwagner (approved):
seems we got a little overeager with the performance optimizations.

PR #919, opened 2026-08-19
[CI] Reduce number of parallel jobs for Ubuntu jobs with Julia v1.11

giordano:
Ref: https://github.com/NumericalEarth/Breeze.jl/pull/902#discussion_r3817600009.


glwagner (approved):
sad but necessary it would seem

PR #925, opened 2026-08-21
Mass-weight the vertically implicit diffusion solve, fill closure-field halos, route the implicit solve by name

### LINKED ISSUE #870: Closure-field halos are never filled, halving the vertical viscous stress in the first column

### LINKED ISSUE #871: Vertically-implicit diffusion is not mass-flux weighted, giving a spurious `c ∝ 1/ρ` steady state

### LINKED ISSUE #915: scalar_substep! passes an off-by-one closure field index: every scalar gets its predecessor's diffusivity
Found during review of #902's AIVA work; confirmed with the reproducer below on current main. Related record: #897. Sibling of #913/#914 (same code path).

## The bug

In `scalar_substep!` (src/TimeSteppers/acoustic_substep_helpers.jl), the closure field index passed to `implicit_step!` is `Val(i - n_acoustic)` with `n_acoustic = 5`, indexing prognostics `(ρᵈ, ρu, ρv, ρw, ρθ, ρqᵉ, …)`. But the closure's scalar names are `(ρθ, ρqᵉ, …)` — ρθ is scalar 1 — so every scalar on the acoustic path asks the closure for the *previous* scalar's diffusivity: the first moisture variable gets ρθ's κ, and the last scalar's κ is never used.

## Measured signature

With per-field diffusivities `κ = (ρθ=1, ρqᵉ=2, c1=3, c2=4)` on a model with two extra tracers, the reproducer instruments which κ each implicit solve receives:

- current mapping: **wrong for 3/3 scalars** (each receives its predecessor's κ)
- `length(closure.κ) = 4`, largest index reached by current code = 3 — the mismatch is silent, no bounds error
- proposed fix (`Val(i - n_acoustic + 1)`): wrong for 0/3; `implicit_substep!`'s own `Val(1)` for ρθ is consistent with it

## Why it matters

Silent wrong vertical implicit diffusion for every moisture variable and tracer whenever a closure with per-field diffusivities is used with `VerticallyImplicitTimeDiscretization` — no crash, just the wrong κ, which is the worst kind of wrong.

## Reproducer

<details><summary>78-line script (CPU, seconds)</summary>

```julia
# Adversarial verification of the "Val(i - n_acoustic) is off by one" claim.
#
# Strategy: do NOT trust an end-to-end signal alone. Ask the live objects directly:
#   for each prognostic index i on the acoustic scalar path, what diffusivity does
#   `Oceananigans.TurbulenceClosures.diffusivity(closure, closure_fields, Val(i - 5))`
#   return, and is it the diffusivity the user asked for for `names[i]`?
# Then do an independent end-to-end confirmation on a DIFFERENT signal than the reviewer's
# (κ on the tracer only, expecting the tracer to be the one that moves if the map is right).

using Breeze
using Oceananigans
using Oceananigans.Units: kilometers
using Oceananigans.TurbulenceClosures: VerticallyImplicitTimeDiscretization, diffusivity
using Breeze.Microphysics: SaturationAdjustment
using Breeze.CompressibleEquations: SplitExplicitTimeDiscretization
using Printf

FT = Float64
Nx, Nz = 16, 32
grid = RectilinearGrid(CPU(); size=(Nx, Nz), halo=(5, 5), x=(0, 3.2kilometers), z=(0, 3.2kilometers),
                       topology=(Periodic, Flat, Bounded))

# Distinct, identifiable diffusivities so the returned number names its own field.
κ = (ρθ = 1.0, ρqᵉ = 2.0, c1 = 3.0, c2 = 4.0)
closure = VerticalScalarDiffusivity(VerticallyImplicitTimeDiscretization(); κ, ν = 0.0)

dynamics = CompressibleDynamics(SplitExplicitTimeDiscretization())
model = AtmosphereModel(grid; dynamics, microphysics=SaturationAdjustment(),
                        tracers=(:c1, :c2), closure,
                        advection=WENO(FT; weight_computation=Oceananigans.Utils.NormalDivision, order=5))

prognostic = Oceananigans.prognostic_fields(model)
names = keys(prognostic)
println("timestepper        = ", typeof(model.timestepper).name.name)
println("prognostic_fields  = ", names)
println("closure.κ (ordered by the closure's scalar names) = ", model.closure.κ)
println("requested κ        = ", κ)
println()

n_acoustic = 5
println("what `scalar_substep!` line 187 asks the closure for:")
bad = 0
for (i, name) in enumerate(names)
    i <= n_acoustic && continue
    id = i - n_acoustic
    got = diffusivity(model.closure, model.closure_fields, Val(id))
    want = κ[name]
    ok = got == want
    ok || (global bad += 1)
    @printf("  i=%d  name=%-4s  Val(i-5)=Val(%d)  -> κ = %.1f   (user asked κ[%s] = %.1f)  %s\n",
            i, string(name), id, got, string(name), want, ok ? "OK" : "<-- WRONG FIELD")
end
println()

println("what the proposed fix `Val(i - 4)` would ask for:")
bad_fix = 0
for (i, name) in enumerate(names)
    i <= n_acoustic && continue
    id = i - 4
    got = diffusivity(model.closure, model.closure_fields, Val(id))
    want = κ[name]
    ok = got == want
    ok || (global bad_fix += 1)
    @printf("  i=%d  name=%-4s  Val(i-4)=Val(%d)  -> κ = %.1f   (user asked κ[%s] = %.1f)  %s\n",
            i, string(name), id, got, string(name), want, ok ? "OK" : "<-- WRONG FIELD")
end
println()
@printf("current mapping wrong for %d/%d scalars; proposed fix wrong for %d/%d\n",
        bad, length(names) - n_acoustic, bad_fix, length(names) - n_acoustic)

# Also: what does implicit_substep! hard-code for the thermodynamic variable?
θ_got = diffusivity(model.closure, model.closure_fields, Val(1))
@printf("implicit_substep! uses Val(1) for ρθ -> κ = %.1f (user asked κ[ρθ] = %.1f)  %s\n",
        θ_got, κ.ρθ, θ_got == κ.ρθ ? "OK" : "WRONG")

# Bonus: is the last scalar's id even in range? An out-of-range Val would index past κ.
@printf("length(closure.κ) = %d ; largest id used by current code = %d ; by fix = %d\n",
        length(model.closure.κ), length(names) - n_acoustic, length(names) - 4)
```

</details>

### LINKED ISSUE #930: Explicitly-discretized closure + AIVA applies vertical diffusion twice
Found during review of #925; confirmed on current `main`. Sibling of #913/#914/#915 (same code path).
Distinct from #913: that one double-counts the *advective* flux on the bounds-preserving WENO path;
this one double-counts the *diffusive* flux, for any advection scheme.

## The bug

When the closure carries `ExplicitTimeDiscretization` **and** any advection scheme uses
`AdaptiveVerticallyImplicitDiscretization`, the closure's vertical diffusion is applied twice per
step — once through the explicit flux divergence, once through the vertically-implicit solve.

The trigger is the *advection* scheme, so this reaches models where nothing asked for an implicit
closure at all: AIVA on momentum alone doubles the diffusion of a tracer advected with plain `WENO`.

The chain:

1. An explicitly-discretized closure applies its **full** vertical diffusion through the explicit
   flux divergence (`Jᶜz`, `𝒯_uz` in `src/TurbulenceClosures/TurbulenceClosures.jl`, which dispatch
   on `time_discretization(closure)`). Correct, and unconditional.

2. AIVA anywhere in the model makes `AtmosphereModel` build an implicit solver
   (`advection_needs_solver` in `src/AtmosphereModels/atmosphere_model.jl`), even though no closure
   asked for one.

3. Because a solver exists, the guard runs `implicit_step!` for every prognostic. It builds each
   tridiagonal row by calling two coefficient extractors — `κzᶜᶜᶠ` for tracers, `νzᶠᶜᶠ` for
   momentum.

4. **Those must return zero here**, or step 1's diffusion is applied a second time. Their catch-all
   fallbacks do return `zero(grid)`, so a closure defining no specialized method is safe whatever
   the guard does. But `abstract_scalar_diffusivity_closure.jl` defines both for
   `closure::AbstractScalarDiffusivity` *without restricting the time discretization*, so every
   closure Breeze ships hands back its full coefficient instead.

A vertically-implicit closure is fine: step 1 does not happen for it, because the
`VerticallyImplicitTimeDiscretization` dispatch zeroes its explicit vertical flux. Only the
explicit-closure-plus-AIVA combination applies both halves.

## Measured signature

`VerticalScalarDiffusivity(ExplicitTimeDiscretization(); κ = 10)` acting on `cos(πz/Lz)`, the
gravest zero-flux eigenmode, so the amplitude decays as `exp(-κk²t)` and inverting that gives the
rate the run actually applied:

| advection | implicit solver | inferred κ |
|---|---|---|
| plain WENO everywhere | none | 9.99 ✓ |
| AIVA on momentum only, tracer on plain WENO | present | **19.98** ✗ |
| AIVA everywhere | present | **19.98** ✗ |

`SmagorinskyLilly`, `scalar_advection = nothing` so the whole change is diffusive, one step:

| momentum advection | max abs Δρc |
|---|---|
| plain WENO | 6.0883e-6 |
| AIVA | **1.2213e-5** (ratio 2.006) |

And step 4 directly: what the two extractors return for an explicitly-discretized closure, swept
over the interior of an 8³ shear-plus-wave flow. **Every entry should be 0.0** — step 1 has already
applied the whole diffusion, so there is nothing left for the implicit row to add. Each nonzero
number is diffusivity that gets applied twice:

| closure (explicit discretization) | max κzᶜᶜᶠ | max νzᶠᶜᶠ |
|---|---|---|
| `SmagorinskyLilly` | 0.299 | 0.299 |
| `AnisotropicMinimumDissipation` | 0.0361 | 1.811 |
| `VerticalScalarDiffusivity(ν=1, κ=1)` | 1.0 | 1.0 |

None of the closures Breeze ships returns zero, so none is safe.

Note these values are a property of the *closure* alone — the run above uses plain `WENO`, no AIVA.
AIVA does not change what the extractors return; it is what causes them to be consulted at all, by
making the solver exist (step 2). That is why the defect needs both ingredients.

(Sweeping rather than sampling matters: AMD clips its diffusivity at zero, so a single point can
land on a clipped cell and read 0.)

## Why it matters

Every closure Breeze ships, at its default `ExplicitTimeDiscretization`, wherever some advection
scheme is adaptive-implicit. Both `ν` and `κ` are affected, so tracers, moisture, the thermodynamic
variable and horizontal momentum all take the doubled coefficient — silently, with no crash.

## Reproducers

<details><summary>1. <code>VerticalScalarDiffusivity</code> inferred rate (CPU, ~1 min)</summary>

```julia
using Breeze, Oceananigans
using Oceananigans.TurbulenceClosures: ExplicitTimeDiscretization
using Oceananigans.Advection: AdaptiveVerticallyImplicitDiscretization

FT = Float64
Oceananigans.defaults.FloatType = FT
Nz, Lz = 32, 100.0
grid = RectilinearGrid(CPU(); size=(4, 4, Nz), x=(0, 100), y=(0, 100), z=(0, Lz))

κ, Δt, Nt = 10.0, 0.05, 20      # Δt keeps the *doubled* rate explicitly stable: 2κΔt/Δz² = 0.10
k = π / Lz                       # cos(kz) is the gravest zero-flux eigenmode, so it decays as exp(-κk²t)

closure = VerticalScalarDiffusivity(ExplicitTimeDiscretization(); κ)
aiva = WENO(FT; time_discretization = AdaptiveVerticallyImplicitDiscretization(FT; cfl=0.5))
plain = WENO(FT)

for (label, madv, sadv) in (("plain WENO everywhere", plain, plain),
                            ("AIVA momentum only   ", aiva,  plain),
                            ("AIVA everywhere      ", aiva,  aiva))
    model = AtmosphereModel(grid; closure, momentum_advection=madv, scalar_advection=sadv, tracers=:ρc)
    set!(model; ρc = (x, y, z) -> cos(k * z))
    c₀ = sum(interior(model.tracers.ρc) .^ 2)
    for _ in 1:Nt
        time_step!(model, Δt)
    end
    c₁ = sum(interior(model.tracers.ρc) .^ 2)
    decay = sqrt(c₁ / c₀)
    κ_inferred = -log(decay) / (k^2 * Δt * Nt)   # invert exp(-κk²t) for the rate the run actually applied
    println(label, "  solver = ", isnothing(model.timestepper.implicit_solver) ? "nothing" : "present",
            "  decay = ", round(decay, digits=5), "  inferred κ = ", round(κ_inferred, digits=2))
end

println("requested κ = ", κ)
```

```
plain WENO everywhere  solver = nothing  decay = 0.99019  inferred κ = 9.99
AIVA momentum only     solver = present  decay = 0.98047  inferred κ = 19.98
AIVA everywhere        solver = present  decay = 0.98047  inferred κ = 19.98
requested κ = 10.0
```

</details>

<details><summary>2. <code>SmagorinskyLilly</code>, one step, purely diffusive (CPU, ~1 min)</summary>

```julia
using Breeze, Oceananigans
using Oceananigans.Advection: AdaptiveVerticallyImplicitDiscretization

FT = Float64
Oceananigans.defaults.FloatType = FT
grid = RectilinearGrid(CPU(); size=(8, 8, 8), x=(0, 800), y=(0, 800), z=(0, 800),
                       topology=(Periodic, Periodic, Bounded))

aiva  = WENO(FT; time_discretization = AdaptiveVerticallyImplicitDiscretization(FT; cfl=0.5))
plain = WENO(FT)

ρuᵢ(x, y, z) = z / 800 + 0.1 * sinpi(2x / 800) * cospi(2y / 800)
ρcᵢ(x, y, z) = cospi(z / 800)

for (label, madv) in (("momentum plain WENO", plain), ("momentum AIVA      ", aiva))
    model = AtmosphereModel(grid; closure=SmagorinskyLilly(), momentum_advection=madv,
                            scalar_advection=nothing, tracers=:ρc)
    set!(model; θ=300, ρu=ρuᵢ, ρc=ρcᵢ)
    Oceananigans.TimeSteppers.update_state!(model)
    ρc₀ = Array(interior(model.tracers.ρc))
    time_step!(model, 1.0)
    ρc₁ = Array(interior(model.tracers.ρc))
    println(label, "  solver = ", isnothing(model.timestepper.implicit_solver) ? "nothing" : "present",
            "  max|Δρc| = ", round(maximum(abs, ρc₁ .- ρc₀), sigdigits=5))
end
```

</details>

<details><summary>3. Coefficients the solve receives, swept over the interior (CPU, ~1 min)</summary>

```julia
using Breeze, Oceananigans
using Oceananigans.TurbulenceClosures: κzᶜᶜᶠ, νzᶠᶜᶠ, SmagorinskyLilly, AnisotropicMinimumDissipation,
                                       VerticalScalarDiffusivity, ExplicitTimeDiscretization
FT = Float64
Oceananigans.defaults.FloatType = FT
grid = RectilinearGrid(CPU(); size=(8,8,8), x=(0,800), y=(0,800), z=(0,800),
                       topology=(Periodic,Periodic,Bounded))
etd = ExplicitTimeDiscretization()

for (lbl, clo) in (("SmagorinskyLilly  ", SmagorinskyLilly()),
                   ("AMD               ", AnisotropicMinimumDissipation()),
                   ("VerticalScalarDiff", VerticalScalarDiffusivity(etd; ν=1.0, κ=1.0)))
    m = AtmosphereModel(grid; closure=clo, advection=WENO(FT), tracers=:ρc)
    set!(m; θ=300, ρu=(x,y,z)->z/800 + 0.1*sinpi(2x/800)*cospi(2y/800),
            ρv=(x,y,z)->0.05*cospi(2x/800)*sinpi(2y/800),
            ρc=(x,y,z)->cospi(z/800))
    Oceananigans.TimeSteppers.update_state!(m)
    K = m.closure_fields; f = Oceananigans.fields(m)
    # sweep the whole interior instead of sampling one point
    κs = [κzᶜᶜᶠ(i,j,k,grid,m.closure,K,Val(3),m.clock,f) for i=1:8, j=1:8, k=1:8]
    νs = [νzᶠᶜᶠ(i,j,k,grid,m.closure,K,nothing,m.clock,f) for i=1:8, j=1:8, k=1:8]
    fieldinfo = isnothing(K) ? "none" :
        join([string(n, " max=", round(maximum(abs, interior(K[n])), sigdigits=4))
              for n in keys(K) if K[n] isa Field], ", ")
    println(lbl, "  max κzᶜᶜᶠ = ", round(maximum(abs, κs), sigdigits=6),
                 "   max νzᶠᶜᶠ = ", round(maximum(abs, νs), sigdigits=6))
    println("                      closure_fields: ", fieldinfo)
end
```

```
SmagorinskyLilly    max κzᶜᶜᶠ = 0.299064   max νzᶠᶜᶠ = 0.298915
                      closure_fields: νₑ max=0.3015
AMD                 max κzᶜᶜᶠ = 0.0360538   max νzᶠᶜᶠ = 1.81101
                      closure_fields: νₑ max=1.818
VerticalScalarDiff  max κzᶜᶜᶠ = 1.0   max νzᶠᶜᶠ = 1.0
                      closure_fields: none
```

</details>

### Make the implicit row ignore explicitly-discretized closures (recommended)

`_ivd_upper_diagonal` / `_ivd_lower_diagonal` / `_implicit_linear_coefficient` in
`src/TurbulenceClosures/vertically_implicit_diffusion_solver.jl` are reached only from the implicit
solve, and already carry exactly this shape for `closure === nothing`:

```julia
 # When closure is `nothing` (e.g. AIVA without turbulence closure), diffusion contributions are zero
 @inline _implicit_linear_coefficient(i, j, k, grid, ::Nothing, args...) = zero(grid)
 @inline _ivd_upper_diagonal(i, j, k, grid, ::Nothing, args...) = zero(grid)
 @inline _ivd_lower_diagonal(i, j, k, grid, ::Nothing, args...) = zero(grid)

+# An explicitly-discretized closure contributes its whole vertical flux through the explicit
+# tendency, so it must contribute nothing to the implicit row.
+const ExplicitlyDiscretizedClosure = AbstractTurbulenceClosure{<:ExplicitTimeDiscretization}
+
+@inline _implicit_linear_coefficient(i, j, k, grid, ::ExplicitlyDiscretizedClosure, args...) = zero(grid)
+@inline _ivd_upper_diagonal(i, j, k, grid, ::ExplicitlyDiscretizedClosure, args...) = zero(grid)
+@inline _ivd_lower_diagonal(i, j, k, grid, ::ExplicitlyDiscretizedClosure, args...) = zero(grid)
```

Verified: with those three methods injected into `Oceananigans.TurbulenceClosures`, reproducer 1
infers `κ = 9.99` on all three rows.

This belongs upstream — Oceananigans' own models call `implicit_step!` unconditionally, so they have
the same hole.

### Not: restricting the `κz*`/`νz*` extractors

Restricting them to `AbstractScalarDiffusivity{<:VerticallyImplicitTimeDiscretization}` looks like
the obvious fix and is wrong. `κzᶜᶜᶠ` also serves the *explicit* flux — `diffusive_flux_z` for
`AIDorAVD` in `abstract_scalar_diffusivity_closure.jl` — so restricting it would delete the explicit
vertical diffusion of every vertical and isotropic diffusivity closure.

### A guard on the Breeze side does not reach the whole bug

Guarding the implicit step on something vertically implicit actually being present:

```julia
if !isnothing(model.timestepper.implicit_solver) &&
   (vertically_implicit_closure(model.closure) || needs_implicit_solver(advection))
```

fixes only the cases where the affected field's *own* advection is not adaptive-implicit. Measured
with that guard in `ssp_rk3_substep!`:

| advection | inferred κ |
|---|---|
| plain WENO everywhere | 9.99 ✓ |
| AIVA on momentum only | 9.99 ✓ (was 19.98) |
| AIVA everywhere | **19.98** ✗ unchanged |

When the field's own scheme is adaptive-implicit the guard has to let the solve run, because the
advection coefficients are needed; diffusion and advection share one `get_coefficient` row, so the
guard cannot admit one and refuse the other. That is the production case — #913 notes that AR
configurations put AIVA on every moisture tracer.

A Breeze-only fix would therefore have to zero the diffusion half inside Breeze's own
`get_coefficient` methods, in both the z-Center and z-Face wrappers, duplicating what the upstream
change does once.

### Regression test

Reproducer 1, asserting the inferred rate is κ rather than 2κ, with AIVA on the tracer itself — the
case a guard alone does not reach.

## Not a regression, and not fixed by #925

#925 neither introduces the doubling nor removes it: the measurements above are the same on `main`
and on that branch. It happens to rewrite the guard in step 3, which is why this turned up while
reviewing it, but the behavior is unchanged either way — so this needs a fix of its own. (#925's
mass weighting does shift reproducer 2's doubled value slightly, to 1.2176e-5; reproducer 1 infers
the same rates on both.)

Separately: #925 *does* fix #915. Its name-keyed `closure_scalar_index` gives the right κ for 3/3
scalars in #915's own reproducer, against 0/3 on `main`.

Environment: Oceananigans v0.110.19, CPU, Julia 1.12.

glwagner:
Two fixes that @ewquon found and wrote while building the TKE closure in #875, split out of that PR so they can merge on their own, plus the name-based routing of the implicit solve they need and the `DrySurface` trait the #875 example introduced. The closure itself stays in #875.

## Fixes pulled out of #875

**Closes #871** — the vertically implicit solve acted on the density-weighted prognostic `ρc` with Oceananigans' unweighted coefficients, i.e. on `∂t(ρc) = ∂z(κ ∂z(ρc))`, while the explicit path forms `∂z(ρ κ ∂z c)`. That relaxes a conserved tracer toward `c ∝ 1/ρ` — about +20 K/km over a deep column, hidden by the 100 m test column over which `ρ` varies by under 1 %. `MassWeightedImplicitDiffusion` (`src/AtmosphereModels/mass_weighted_implicit_diffusion.jl`) routes every z-Center prognostic to mass-flux-weighted coefficients forming `∂t(ρc) = ∂z(ρ κ ∂z c)`, with the diagonal written out explicitly so mass is conserved; `ρw` keeps the z-Face path (see the file header for why). The implicit step now runs whenever an implicit solver exists rather than only under adaptive implicit advection, so the weighting reaches the single-column configuration too. Tests on a 4 km column: implicit ≈ explicit, steady state well mixed in the *specific* variable, exact mass conservation on a stretched grid, viscosity parity.

**Closes #870** — `fill_halo_regions!` was never called on `model.closure_fields`, so a (Center, Center, Face) diffusivity read zero in the horizontal halos and the vertical viscous stress in the first column was halved (Smagorinsky/AMD included). One line in `compute_auxiliary_variables!`, the same call every Oceananigans model makes. The call fills the *vertical* halos too — a diffusivity's top halo goes from zero to the mirrored interior value — but that part is inert, because the boundary faces are masked and the surface flux comes from the boundary condition rather than the interior stencil. Verified on an 8³ Smagorinsky case: toggling the fill moves the momentum tendency by ≤ 2e-14 relative (roundoff) on the columns that read no horizontal halo, while the edge columns move by O(1) — the latter being the fix. (Distributed rank boundaries are *not* covered: `only_local_halos=true`, and Breeze does not yet compute closure fields one cell into the halo as Oceananigans does.)

## Other Fixes

**Closes #915** — name-based routing of the implicit solve. `ssp_rk3_substep!` indexed the closure's scalar diffusivity by `Val(i - 3)` and `scalar_substep!` by `Val(i - 5)`, which are wrong as soon as a prognostic tuple starts with the compressible dry density `ρᵈ` (every scalar shifted by one; harmless only for closures with a single `κ`). Both steppers now use `closure_scalar_index(model, name)` / `skip_vertical_diffusion(model, name)` (in `AtmosphereModels`, next to `prognostic_field_names`; the same function builds the constructor's `scalar_names`) and the acoustic path `acoustic_prognostic_names(model)`, so the routing is keyed on what a prognostic *is* rather than where it sits. Tested for anelastic and compressible dynamics with user tracers, including a step through both steppers; on #915's own reproducer the routing now returns the requested `κ` for 3/3 scalars, against 0/3 before.

**Closes #930** (result of the routing change this PR already needed, no code was written for this) — an explicitly-discretized closure combined with adaptive implicit vertical advection applied the closure's vertical diffusion twice: once through the explicit flux divergence, once through the implicit solve. The trigger is the *advection* scheme, so AIVA on momentum alone doubles the diffusion of a tracer carried by plain `WENO`. Oceananigans has two `implicit_step!` methods, and only the one taking `advection` filters an explicitly-discretized closure to `nothing` before the coefficient extractors see it (CliMA/Oceananigans.jl#5631, in v0.110.20); `main` picks between them on `needs_implicit_solver(advection)`, so a prognostic whose own scheme is not adaptive-implicit still takes the unfiltered path. Running the implicit step whenever a solver exists — the same routing change the mass weighting above needs — sends every prognostic through the filtered one. `VerticalScalarDiffusivity(ExplicitTimeDiscretization(); κ = 10)` relaxing the gravest zero-flux eigenmode, inferred κ for plain WENO / AIVA on momentum only / AIVA everywhere: 9.99 / **19.98** / 9.99 on `main`, 9.99 / 9.99 / 9.99 here. `SmagorinskyLilly` with `scalar_advection = nothing`, one step, max abs Δρc: 6.0883e-6 (plain) vs **1.2213e-5** (AIVA) on `main`, 6.0883e-6 both ways here.

Also, cb8fdd89 addresses #625 item 39.

## New feature for #875

**`DrySurface`** — a surface with no saturation humidity, for bulk surface-layer schemes over land: with `PlanarLiquidSurface` a `PolynomialCoefficient` forms the surface virtual potential temperature from the saturation humidity at the surface temperature, which over land is ≈ 0.33 K of spurious virtual warming at 265 K. Promoted from the #875 example into `Thermodynamics` (it was extending a `src` function from an example), with a docstring and test.

---

Tests run locally on CPU: `vertical_diffusion`, `polynomial_bulk_coefficients`, `acoustic_substepping`, `implicit_vertical_advection`, `turbulence_closures`, `unit_tests`, `quality_assurance`.


ewquon:
I pushed the merge of `main` into this branch — two merge commits (#902/#922/#927, then #926) plus a regression test for #870, which had none. The PR is `MERGEABLE` again.

Resolving it turned up an interaction between this PR and #902 that no conflict marker showed, and I made a judgment call there.

## The interaction

AIVA partitions the advecting velocity; the explicit half advects with `s·w`, the implicit with `(1-s)·w`; those sum to `w`. The density isn't partitioned and enters identically in both halves as the face mass-flux weight `ρᶠ`. What matters is that the two halves read the same `w` and `ρ`: the explicit flux is `ρᶠ·(s·w)·ĉ_WENO` and the implicit one `ρᶠ·((1−s)·w)·ĉ_upwind`; the mass flux with blended reconstruction is `ρᶠ·w·[s·ĉ_WENO + (1−s)·ĉ_upwind]`. However, `w` and `ρᵈ` are overwritten in the acoustic loop. #902 freezes `w` and `ρᵈ` together at stage entry so that the effective transport remains `ρw`.

This PR gives `ρᵈ` a second job in the same tridiagonal row: weighting the diffusion coefficients. Merged naively, the diffusion half inherits the frozen density too. But nothing about diffusion is split. The role of density is to weight the diffusive flux and convert the prognostic `ρθ` into the `θ` the operator acts on, thus the density should be a live value matching the `ρθ` that has been advanced by the acoustic loop. 

## Evidence

`test/implicit_vertical_advection.jl` (from #902, run as a standalone script on all four trees), "Below threshold, AIVA does not disturb vertically-implicit diffusion": 4 km column, 8 steps of Δt = 1 s, reporting `max |θ(adaptive) − θ(explicit)|` against the test's `rtol = sqrt(eps(Float64))`, i.e. ≈ 4.5e-6 K at θ ≈ 300 K.

| tree | max abs Δθ | |
|---|---|---|
| #902 only (`main`, 63c90c20) | 0.0 K | pass |
| #925 only (cb8fdd89, pre-merge) | 0.0 K | pass |
| #902 + #925, naive merge | **6.57e-3 K** | fail — ~1500× tolerance |
| #902 + #925, with the change below | 0.0 K | pass |

Neither PR is wrong alone; only the combination is. All three passing rows agree *exactly*, not merely within tolerance.

## What I did

`MassWeightedImplicitDiffusion` now carries an optional density that overrides the one the solve is called with, for the diffusion half only:

```julia
struct MassWeightedImplicitDiffusion{A, D}
    scheme :: A
    diffusion_density :: D
end

MassWeightedImplicitDiffusion(scheme) = MassWeightedImplicitDiffusion(scheme, nothing)

# Dispatch rather than a branch: `nothing` defers to the density the solve was called with.
@inline mass_weighting_density(::Nothing, ρ) = ρ
@inline mass_weighting_density(ρᵈ, ρ) = ρᵈ
```

`implicit_substep!` sets it to the live coupling density for momentum and `ρθ`, keeping your frozen `(w, ρᵈ)` for the advection coefficients. `ssp_rk3_substep!` and `scalar_substep!` pass `nothing` and keep single-density behavior, since only the acoustic substepper freezes anything. 

Reasoning: The backward-Euler solve represents the state at the end of the stage, so the density consistent with the `ρθ` it is solving for is the live one. The frozen density exists for a different job — pairing the explicit and implicit fractions of one advective flux. The diffusive flux has no explicit/implicit splitting.

Separate commit: `implicit_step_advection` is renamed to `implicit_step_scheme`. It stopped returning an advection scheme in this PR — routing every z-Center prognostic through `MassWeightedImplicitDiffusion` made it return an object that configures *both* halves of the row — and passing a diffusion density to something named for advection made that hard to ignore. The value still travels in `implicit_step!`'s `advection` slot.

Passing locally on CPU after the merge and rename: `implicit_vertical_advection`, `vertical_diffusion`, `turbulence_closures`, `unit_tests`, `quality_assurance`, `doctests`, `polynomial_bulk_coefficients`, `acoustic_substepping_{components,stability,open_boundaries}`, `substepper_{rest_state,structural}`, `terrain_following_split_explicit`, `open_boundary_momentum`.

test/vertical_diffusion.jl
```diff
@@ -137,3 +141,259 @@ using Test
...
+    ρ = Breeze.AtmosphereModels.total_density(model.dynamics)
+    scheme = Breeze.AtmosphereModels.implicit_step_scheme(nothing, :ρc)
+    w = model.velocities.w
+    clock = model.clock
+    mf = Oceananigans.fields(model)
+    c, f = Center(), Face()
+    id = Breeze.AtmosphereModels.closure_scalar_index(model, :ρc)
+
+    # The trailing arguments `implicit_step!` forwards. Boundary conditions are appended from
+    # Oceananigans v0.110.20 onward; `weighted_args` mirrors that call exactly.
```
giordano:
```suggestion
```

test/vertical_diffusion.jl
```diff
@@ -137,3 +141,259 @@ using Test
...
+##### and immersed boundary conditions after `(advection, w, density)` — a method that fixes the
+##### trailing argument count stops matching, and the solve silently falls through to Oceananigans'
+##### *unweighted* diffusion-only fallback. Nothing errors; the weighting just disappears.
+#####
+##### The physics tests above do catch that today, but only because this particular fallthrough
+##### happens to change the steady state. This asserts the dispatch itself: the weighted method must
+##### win for the argument list `implicit_step!` actually passes, and must return something the
+##### unweighted fallback does not.
+#####
+
```
giordano:
```suggestion
```

ewquon (approved):

ewquon:
Now that Oceananigans v0.111.0 is out:

**Merge this after #936.** That PR bumps compat to `0.111` and deletes the z-Face advection block,
`VerticalMomentumImplicitAdvection` and its `get_coefficient` seam. We rewrite the same three files,
so #936 first leaves this one a mechanical rebase.

**Reverted the `implicit_step_advection` → `implicit_step_scheme` rename**, since #936 removes the
function it renamed.

**One line will be needed once the compat moves to 0.111** (commit to follow). CliMA/Oceananigans.jl#5903 gave
`implicit_step!` an early return keyed on `needs_implicit_solver(advection)`.
`MassWeightedImplicitDiffusion` wraps the scheme before it reaches that slot and does not extend the
trait, so it answers `false`: with an explicitly-discretized closure the solve returns before running
and the weighting silently disappears. Forwarding to `a.scheme` fixes it, and is inert on 0.110.20.

bischtob:
NUM-227

PR #927, opened 2026-08-22
Split acoustic substepping tests into three files for load balancing

giordano:
This file grew way too much, and it causes particularly bad load balancing when running the tests.


glwagner (approved):

giordano:
With this reorganisation, macOS jobs finish in 30 minutes instead of 50+ 🙂 Also, macOS with Julia v1.11 was the job to finish, which hasn't been the case for a very long time.

PR #929, opened 2026-08-26
Accept any AbstractDateTime in the solar geometry

Pangoraw:
`day_of_year`, `hour_angle` and both `cos_solar_zenith_angle` methods annotated the concrete `Dates.DateTime`, so any other `AbstractDateTime` hit a `MethodError`:

```
MethodError: no method matching cos_solar_zenith_angle(::ReactantDateTime{TracedRNumber{Int64}}, ::Float64, ::Float64)

Closest candidates are:
  cos_solar_zenith_angle(!Matched::DateTime, ::Any, ::Any)
   @ Breeze src/CelestialMechanics/solar_zenith_angle.jl:131
```

The bodies never needed the concrete type. They call `Dates.dayofyear`, `hour`, `minute` and `second` plus arithmetic — all part of the `AbstractDateTime` interface — so widening the annotations is the entire change. `DateTime <: AbstractDateTime`, so every existing call resolves exactly as before; this only *adds* dispatch.

## Why

This matters for datetime types whose fields aren't plain `Int64`. Under Reactant a compiled step's clock is traced, and a wall-clock datetime derived from it is a `ReactantDateTime` whose milliseconds are a `TracedRNumber`. That is an `AbstractDateTime` and all its accessors are defined, so with this change the full solar geometry — Spencer declination, equation of time, hour angle — computes inside a compiled executable.

Concretely, it is the difference between radiation having a real diurnal cycle and being pinned to a constant. Sampled from a traced clock over 24 h at 125°W, 47°N, every value matching a host computation to 1e-10:

```
 0h: -0.622   4h: 0.007   8h: 0.349  12h: 0.061  16h: -0.567  20h: -0.910
```

`FixedCosineZenith` — the only solar position that traces today — is a flat line through that.

## Scope

This is one of several independent things needed for clock-driven radiation under Reactant; it is the Breeze-side piece and it stands on its own. The corresponding `Dates`/`Adapt` gaps live in Reactant and are proposed separately in EnzymeAD/Reactant.jl#3224. Notably, with that PR the `compute_datetime` conversion in `BreezeRRTMGPExt` needs **no change** — hence nothing here touches it.

## Tests

New `test/solar_zenith_angle.jl`:

- known values — solstice day-of-year and declination, overhead sun at 23.5°N, night on the far side, hour angle ≈ 0 at solar noon
- polar day and polar night at 89°N
- the widened dispatch, using a locally defined `AbstractDateTime` so it needs **no extra dependency** and runs everywhere, asserting bit-identical results against the equivalent `DateTime`
- the single-column-grid method

Locally: **45 passed, 0 failed**, on CPU.

https://claude.ai/code/session_01LShNSUDBfvosAH2onkcq3q


src/CelestialMechanics/solar_zenith_angle.jl
```diff
@@ -11,8 +11,11 @@
 $(TYPEDSIGNATURES)
 
 Return the day of year (1-365/366) for a given DateTime.
+
+Accepts any `AbstractDateTime`, not just `Dates.DateTime`, so that datetime types whose fields are
+traced numbers (for example Reactant's `ReactantDateTime`) can flow through the solar geometry.
```
giordano: Typical LLM talk arguing about history of code in comments, or worse, public-facing docstrings
```suggestion
```

CC @ewquon in relation to https://github.com/NumericalEarth/NumericalEarth.jl/pull/576#discussion_r3864287518.

src/CelestialMechanics/solar_zenith_angle.jl
```diff
@@ -11,8 +11,11 @@
 $(TYPEDSIGNATURES)
 
 Return the day of year (1-365/366) for a given DateTime.
```
giordano:
```suggestion
Return the day of year (1-365/366) for a given `AbstractDateTime`.
```

src/CelestialMechanics/solar_zenith_angle.jl
```diff
@@ -11,8 +11,11 @@
...
+
+Accepts any `AbstractDateTime`, not just `Dates.DateTime`, so that datetime types whose fields are
+traced numbers (for example Reactant's `ReactantDateTime`) can flow through the solar geometry.
 """
-day_of_year(dt::DateTime) = Dates.dayofyear(dt)
+day_of_year(dt::AbstractDateTime) = Dates.dayofyear(dt)
```
giordano, line 15: Unrelated to this PR, but it's unclear to me why we have this function in the first place, instead of using `Dates.dayofyear` directly.

src/CelestialMechanics/solar_zenith_angle.jl
Pangoraw: excellent point, thanks

giordano (approved):

PR #933, opened 2026-08-27
Fix compatibility with Oceananigans v0.110.20

giordano:
https://github.com/CliMA/Oceananigans.jl/pull/5631 moved `needs_implicit_solver` from `Oceananigans.Advection` to `Oceananigans.BoundaryConditions`

PR #936, opened 2026-08-28
(0.10.0) Upstream `Solvers.get_coefficient` in Oceananigans

simone-silvestri:
<s>To capture #933 and forward Oceananigans' 0.110.20 compatibility to NumericalEarth</s>

The objective of this PR is to upstream the bespoke `Solvers.get_coefficient` in Oceananigans. These implementations are not Breeze-specific, they are just an extension of the implicit solver for `ZFaceField`s which is useful also for Oceananigans' native NonhydrostaticModel

Depends on https://github.com/CliMA/Oceananigans.jl/pull/5903


Project.toml
```diff
@@ -1,6 +1,6 @@
 name = "Breeze"
 uuid = "660aa2fb-d4c8-4359-a52c-9c057bc511da"
-version = "0.9.0"
+version = "0.9.1"
```
giordano:
```suggestion
version = "0.10.0"
```
because #926 is a breaking change

simone-silvestri:
Maybe we wait for https://github.com/CliMA/Oceananigans.jl/pull/5903 so we can adapt here?

giordano (approved):
Assuming CI is happy (only docs build failed before, but that should be fixed by #945), this is good to go for me.  Always happy to see lines of code shrinking 😁

PR #900, opened 2026-08-12
Implement PredictedParticleProperties (P3) microphysics

kaiyuan-cheng:
Supersedes #395 — same work, rebased onto current `main` as one commit instead
of 401.

Single-category ice microphysics tracking Fortran
[P3-microphysics v5.5.0](https://github.com/P3-microphysics/P3-microphysics)
(Morrison and Milbrandt 2015a; Milbrandt et al. 2025 for predicted liquid
fraction). Instead of sorting frozen hydrometeors into fixed categories, P3
carries one ice category whose rime mass, rime volume, and liquid coating are
prognostic, so density and fall speed follow from the state rather than a
category label.

Eight prognostic densities by default: cloud liquid mass, rain mass and number,
dry ice mass and number, rime mass and rime volume, and liquid coating on ice.
Cloud droplet number, aerosol number, and saturation deficit are gated on a
type. Warm rain uses Khairoutdinov and Kogan (2000).

39 files under `src/Microphysics/PredictedParticleProperties/`: 5D/6D ice
integrals parsed from the Fortran ASCII tables, rain tables generated at
startup by Chebyshev–Gauss quadrature, and the full process-rate set (riming,
aggregation, deposition, melting, wet growth, nucleation, freezing,
evaporation, coupled saturation adjustment). GPU-compatible throughout.
Documented across eight pages, including a scheme-scoped notation page.

## Two deliberate divergences

**Rime-volume threshold.** `consistent_rime_state` gates on
`minimum_mass_mixing_ratio / maximum_rime_density` (≈1.1e-17 m³/kg), not the
`1e-15` literal `calc_bulkRhoRime` ships. That quotient is Fortran's own
`bsmall`, sitting commented out beneath the live test and already used at the
corresponding tests elsewhere; a third site is marked
`!*** NOTE: change to 'bsmall'`. States with `bᶠ ∈ [1.1e-17, 1e-15)` and
significant `qᶠ` are now repaired rather than discarded. Both Fortran-parity
suites pass, so no traced regime reaches that band.

**Density-correction floor.** Three rain/ice terminal-velocity paths now route
through `ice_air_density_correction`, picking up its 0.01 kg/m³ air-density
floor. Engages only above ~30 km.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #939, opened 2026-08-29
[test] Fix P3 integration tests on GPU

giordano:
Ref: https://github.com/NumericalEarth/Breeze.jl/pull/900#issuecomment-5461129519.  Close #938.

### DIFF
--- a/test/predicted_particle_properties_integration.jl
+++ b/test/predicted_particle_properties_integration.jl
@@ -11,6 +11,7 @@ using Breeze.ParcelModels: step_parcel_state!
 
 using Oceananigans: Bounded, CPU, Center, CenterField, Face, Field, Flat, GridFittedBottom,
                      ImmersedBoundaryGrid, RectilinearGrid, compute!, set!, time_step!
+using Oceananigans.Architectures: on_architecture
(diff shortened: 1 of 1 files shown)


test/predicted_particle_properties_integration.jl
```diff
@@ -21,7 +22,7 @@ using Oceananigans.TimeSteppers: update_state!
         underlying_grid = RectilinearGrid(default_arch, FT;
                                           size = (2, 1, 4),
                                           x = (0, 2), y = (0, 1), z = (0, 4))
-        bottom_height = reshape(FT[0, 2], 2, 1)
+        bottom_height = on_architecture(default_arch, reshape(FT[0, 2], 2, 1))
```
giordano, line 25: This is the same pattern as upstream in Oceananigans: https://github.com/CliMA/Oceananigans.jl/blob/83d509b0b5db0c81ea73cb48e70ed84b6779bd54/test/test_split_explicit_vertical_integrals.jl#L166-L168.  Perhaps this can be automated upstream, in the meantime we follow the same pattern.

kaiyuan-cheng:
@giordano Thanks a lot!

PR #952, opened 2026-09-01
Density-weight the vertically implicit diffusion solve for `ρw`

### LINKED ISSUE #944: Vertical momentum `ρw` is not mass-flux weighted in the vertically-implicit solve
Follows #925. Blocked on Oceananigans releases: the weighting itself needs one carrying
CliMA/Oceananigans.jl#5904 (merged 2026-08-28; the newest tag, v0.110.20, predates it), and the
deduplication in step 2 below needs 0.111.0 carrying CliMA/Oceananigans.jl#5903.

## What is missing

#925 mass-flux weights the vertically-implicit solve for every z-Center prognostic — `ρu`, `ρv`, `ρθ`, `ρqᵗ`, tracers — so the implicit operator matches the `∂z(ρ κ ∂z c)` the explicit path forms. **`ρw` is excluded.** It still solves `∂t(ρw) = ∂z(ν ∂z(ρw))` while the explicit path forms `∂z(ρ ν ∂z w)`, so vertical momentum is the one prognostic whose implicit viscosity disagrees with its explicit counterpart wherever `ρ` varies.

## Why it is blocked

Upstream's z-Face `ivd_lower_diagonal` returned the wrong coefficient: it evaluated `ν` one cell center too high, and `Δzᶜ` two centers too high (fixed by CliMA/Oceananigans.jl#5904). Breeze's weighting multiplies upstream's coefficients by a density ratio, so a wrong coefficient stays wrong no matter how `ρ` is placed — there is no correct z-Face implementation to write in Breeze on top of it.

## Why it matters

Operator error against a hand-derived reference, `Lz = 4 km`, `ρ = 1.2 exp(-z/7830 m)` (ρ varies ≈ 40% over the column), interior faces:

| viscosity | unweighted (today) | weighted, with #5904 |
|---|---|---|
| constant `ν` | 15.1% (L2) | machine precision |
| `ν` decreasing with height | 22.3% (L2) | machine precision |

## What the change involves

1. **z-Face diffusion coefficients.** Upstream already has `ivd_upper_diagonal`/`ivd_lower_diagonal` at `ℓz::Face` — #5904 corrected the lower one's `ν` and `Δzᶜ` indices, and #5903 does not touch them (what it adds at z-Face is the *advection* trio in step 2). What is missing is the weighting, so the `mass_weighted_ivd_*` helpers become `ℓz::Center`-specific with `::Face` mirrors. The rule is the same for both: each coefficient is upstream's, times `ρ` where that coefficient's flux lives, divided by `ρ` where the value it multiplies is reconstructed as a specific variable. Only the locations exchange, because `ρw` sits at faces with its stress at centers — the opposite of a tracer:

   | prognostic | specific variable | flux lives at | reconstructed at |
   |---|---|---|---|
   | `ρθ`, `ρqᵛ`, `ρu`, `ρv` (z-Center) | `c = ρc/ρᶜ` | face | center |
   | `ρw` (z-Face) | `w = ρw/ρᶠ` | center | face |

   so where the z-Center rows carry `ρᶠ/ρᶜ`, the z-Face rows carry `ρᶜ/ρᶠ`:

   ```
   du(k) = upstream_du(k) · ρᶜₖ / ρᶠₖ₊₁
   dl(m) = upstream_dl(m) · ρᶜₘ / ρᶠₘ
   d(k)  = 1 − Δt·λₖ − [upstream_du(k)·ρᶜₖ + upstream_dl(k−1)·ρᶜₖ₋₁] / ρᶠₖ
   ```

   `λ` is the closure's `_implicit_linear_coefficient`, passed through unweighted because it damps `ρw` and `w` at the same rate. The diagonal is written out rather than formed as `1 − du − dl` because both off-diagonal fluxes act on `ρwₖ`, so both divide by this row's `ρᶠₖ` rather than by the neighbors' — the same argument as at z-Centers, with `ρᶠ` in place of `ρᶜ`.

2. **Delete Breeze's z-Face *advection* coefficients and delegate to upstream.** #5903 adds
   `implicit_advection_{upper,lower}_diagonal`/`diagonal` for `ℓz::Face` carrying the same `ρᶜ/ρᶠ`
   weighting Breeze derived, plus `explicit_velocity_scaleᶜᶜᶜ` and `implicit_vertical_velocityᶜᶜᶜ`.
   Once Breeze is on 0.111.0, `implicit_w_advection_*`, both `ᶜᶜᶜ` velocity helpers, the
   `VerticalMomentumImplicitAdvection` wrapper and its `get_coefficient` seam all go, and `:ρw` routes
   through `MassWeightedImplicitDiffusion` like every other prognostic — its methods add the weighted
   diffusion and pass `ℓz` through to upstream's advection coefficients. One wrapper instead of two.

   What stays Breeze-specific is the *explicit* momentum-flux CFL scaling: Breeze advects momentum
   with the mass flux `(ρu, ρv, ρw)`, so upstream's scaling would split on `|ρw|` rather than `|w|`,
   inconsistently with the implicit solve.

3. **Seam.** On 0.111.0 there is a single fixed-arity `get_coefficient` per diagonal, so the `args...`
   form from #925/#933 becomes an exact signature. Anything placed in the `advection` slot must also
   extend `needs_implicit_solver`, or #5903's new guard skips the solve outright.

Net a deletion: roughly −130 in `implicit_vertical_advection.jl` (the z-Face coefficient block, the `ᶜᶜᶜ` velocity helpers, the wrapper and its seam, and header item 3 with its TODO) against roughly +45 in `mass_weighted_implicit_diffusion.jl` for the `::Face` mirrors, plus tests.

## Already verified on a prototype

Measured against the current `main` + a locally applied #5904:

- **Composition** — exact to machine precision on every `ν` profile tested, and on stretched grids, with no change to the weighting code.
- **Diagonal dominance** — the weighting reduces row-sum slack at the surface row but cannot trigger the tridiagonal solver's silent-skip branch: `min|β| / (10 eps)` ≥ 6.4e14 across Δt up to 1e12, Float32 included.
- **GPU** — `Adapt` covers the new field, coefficients are type-stable and allocation-free, no kernel-rule violations.
- **Performance** — within noise on CPU (≤1%), allocations byte-identical.

## Tests to add

A `ρw` analogue of the deep-column tests in `vertical_diffusion.jl` (implicit ≈ explicit on a 4 km column; exact mass conservation on a stretched grid), plus a dispatch test of the kind added in #925 so a future `get_coefficient` seam change cannot silently drop the weighting.

ewquon:
Closes #944.

`ρw` was the one prognostic whose vertically-implicit viscosity was not mass-flux
weighted — it solved `∂t(ρw) = ∂z(ν ∂z(ρw))` while the explicit path forms
`∂z(ρ ν ∂z w)`. Its rows carry `ρᶜ/ρᶠ` where the z-Center rows carry `ρᶠ/ρᶜ`;
the derivation is in the commit message.

Unblocked by CliMA/Oceananigans.jl#5904, released in v0.111.0 and picked up in #936.

Verified on CPU and a T4 in both precisions: the coefficients match a hand-derived
reference to 64 eps on uniform and stretched grids under constant and height-varying
ν, `implicit_step!` inverts that reference row by row, and the Δzᶠ-weighted column
sums telescope exactly. Unweighted, the operator misses that reference by 5.4–6.4%.

---

Also, for clarity, renamed `MassWeightedImplicitDiffusion` → `DensityWeightedImplicitOperator`
and `implicit_step_advection` → `implicit_step_scheme`; see b020c76 and f0445cf for the rationale.


src/AtmosphereModels/implicit_vertical_advection.jl
```diff
@@ -44,15 +44,15 @@ const AIVA = AdaptiveImplicitVerticalAdvection
...
+# Explicit schemes are wrapped too, since the diffusion half needs the weighting either way; the
+# advection half passes `ℓz` through to upstream's coefficients unchanged.
 #
 # `diffusion_density` weights the diffusion half when it must differ from the density the solve is
-# called with, which happens only under the acoustic substepper (see `MassWeightedImplicitDiffusion`).
-implicit_step_advection(advection, name::Symbol, diffusion_density=nothing) =
-    name === :ρw ? advection : MassWeightedImplicitDiffusion(advection, diffusion_density)
+# called with, which happens only under the acoustic substepper (see `MassWeightedImplicitOperator`).
+implicit_step_scheme(advection, diffusion_density=nothing) =
+    MassWeightedImplicitOperator(advection, diffusion_density)
```
glwagner:
```suggestion
implicit_step_scheme(advection, diffusion_density=nothing) = MassWeightedImplicitOperator(advection, diffusion_density)
```

src/AtmosphereModels/mass_weighted_implicit_diffusion.jl
```diff
@@ -58,18 +63,18 @@ using Oceananigans.TurbulenceClosures:
 # diffusion half has no such pairing to keep: it reconstructs the specific variable as `q / ρᵈ`
 # from the field it is solving for, which the acoustic loop has already advanced, so it needs the
 # live `ρᵈ`. `nothing` means "use the density the solve was called with".
-struct MassWeightedImplicitDiffusion{A, D}
+struct MassWeightedImplicitOperator{A, D}
     scheme :: A
     diffusion_density :: D
```
glwagner: should the abstraction have consistent naming like "DensityWeighted", and then a `density` property? I also think "diffusion" can be inferrred from context so the property can just be called `density`

src/AtmosphereModels/density_weighted_implicit_diffusion.jl
```diff
@@ -93,22 +98,22 @@ BoundaryConditions.needs_implicit_solver(a::MassWeightedImplicitDiffusion) =
 # The batched solver evaluates the off-diagonals only for k = 1 … Nz-1, so the divisions below
 # never reach a halo value of `ρ`; the diagonal's `ρᶠ⁺` at k = Nz+1 does, but it only multiplies
 # a `du` that the peripheral-node mask has already zeroed.
-@inline function mass_weighted_ivd_upper_diagonal(i, j, k, grid, clo, K, id, ℓx, ℓy, ℓz, Δt, clk, fields, ρ)
+@inline function mass_weighted_ivd_upper_diagonal(i, j, k, grid, clo, K, id, ℓx, ℓy, ℓz::Center, Δt, clk, fields, ρ)
     du = _ivd_upper_diagonal(i, j, k, grid, clo, K, id, ℓx, ℓy, ℓz, Δt, clk, fields)
     ρᶠ = densityᶜᶜᶠ(i, j, k+1, grid, ρ)   # where κₖ₊₁ weights the flux
     ρᶜ = densityᶜᶜᶜ(i, j, k+1, grid, ρ)   # where qₖ₊₁ is reconstructed as cₖ₊₁ = qₖ₊₁/ρᶜₖ₊₁
```
glwagner, line 104: are these `density***` operators different from ordinary interpolation?

glwagner (approved):

src/AtmosphereModels/density_weighted_implicit_diffusion.jl
ewquon: From Oceananigans:
```julia
densityᶜᶜᶜ(i, j, k, grid, ρ) = @inbounds ρ[i, j, k]
densityᶜᶜᶠ(i, j, k, grid, ρ) = ℑzᵃᵃᶠ(i, j, k, grid, ρ)
```

PR #953, opened 2026-09-01
Thread boundary-condition args through the auxiliary halo fills

ewquon:
Three `fill_halo_regions!` calls inside `update_state!` refill halos of fields that may carry user boundary conditions without passing the boundary-condition args, so `getbc` is reached with no clock:

| function | file | fields refilled |
|---|---|---|
| `compute_auxiliary_dynamics_variables!` | `src/CompressibleEquations/compressible_time_stepping.jl:378-379` | dry density, formulation prognostics |
| `compute_terrain_temperature_and_pressure!` | `src/CompressibleEquations/terrain_compressible_physics.jl:703-704` | dry density, formulation prognostics |
| `compute_auxiliary_thermodynamic_variables!` | `src/AtmosphereModels/update_atmosphere_model_state.jl:243` | microphysical prognostics |

For a time-dependent boundary condition this is a `MethodError` — `ContinuousBoundaryFunction` is not callable without the clock. A compressible model carrying a continuous `Value` condition on the dry density, on the thermodynamic prognostic, or on a prognostic microphysical moment therefore cannot be stepped at all. A condition type that tolerates a missing clock is worse off: it is evaluated at the wrong time, without complaint.

The ordering is what makes the first two reachable. Within a single `update_state!`:

```
fill_halo_regions!(prognostic_fields(model), boundary_condition_args(model)..., async = true)
compute_auxiliary_variables!(model)
  ├─ compute_velocities!                        fills density + momentum halos WITH args,
  │                                             then sets u = ρu / ℑxᶠᵃᵃ(ρᵈ) over 1:Nx+1
  ├─ compute_auxiliary_thermodynamic_variables! ← site 3 (microphysical prognostics)
  └─ compute_auxiliary_dynamics_variables!      ← site 1, or site 2 for a TerrainCompressibleModel
                                                refills density + formulation prognostics
                                                WITHOUT args, then computes T and p
```

The velocity has therefore already been diagnosed from a correctly-timed density halo by the time that halo is overwritten with one evaluated at the wrong time, so `ρu = ℑxᶠᵃᵃ(ρᵈ) u` no longer holds at the boundary face — an identity `_compute_velocities!` establishes by construction over `1:Nx+1`. Sites 1 and 2 are alternatives by dispatch, never both.

Site 3 is the same defect on the prognostic microphysical moments. It does not touch the density and is independent of that identity.

The refill exists because the fill at the top of `update_state!` is `async = true`, so the equation-of-state kernel cannot assume it has landed. For the *formulation prognostics* that is real work — nothing else fills `ρθ` synchronously. For the *density* it is redundant with the synchronous fill `compute_velocities!` performs immediately above, and dropping it there may be preferable to threading the args; see the discussion below.

#742 fixed this class at three sites, `compute_velocities!` among them. These run afterward and undo it.

`boundary_condition_args` moves to a module-level import in `CompressibleEquations`: `compressible_time_stepping.jl` is included before `acoustic_substepping.jl`, which held the only existing import, so relying on it would depend on include order.

### Tests

Three regression tests in `test/forcing_and_boundary_conditions.jl`, each verified to fail without the corresponding fix **and to land on its own call site**:

| test | stashed failure lands on |
|---|---|
| `Time-dependent Value BC on density and ρθ` | `compressible_time_stepping.jl:378` |
| `Time-dependent Value BC on density and ρθ over terrain` | `terrain_compressible_physics.jl:703` |
| `Time-dependent Value BC on a microphysical prognostic` | `update_atmosphere_model_state.jl:243` |

Each uses a continuous `Value` condition linear in time, so the expected halo value at a given clock time is analytic and no reference run is needed. The first also asserts the `ρu = ℑxᶠᵃᵃ(ρᵈ) u` identity at the boundary face. The microphysics test uses `DCMIP2016KesslerMicrophysics`, the only scheme whose `prognostic_field_names` is non-empty *and* whose field materialization threads user boundary conditions.

`test/forcing_and_boundary_conditions.jl` goes from 83 to 87 passing; `quality_assurance.jl` (Aqua, ExplicitImports, No Core.Box) passes unchanged.

### Not fixed here

A related group of unguarded fills sits on the `set!` path — `reconcile_densities!` (`compressible_time_stepping.jl:165, 186`), `establish_relative_humidity_densities!` (`:218`), and `set_potential_temperature_hydrostatically_balanced_density!` (`set_to_mean.jl:352`) — and `nudge_initial_fields!` (`adiabatic_balance.jl:83`) fills the prognostics the same way. `set!` with a continuous condition on the dry density therefore still fails.

Those are not reachable from `update_state!`, and threading them piecemeal cannot be demonstrated by a passing test until all are done, so they are left for a follow-up.

### DIFF
--- a/src/AtmosphereModels/update_atmosphere_model_state.jl
+++ b/src/AtmosphereModels/update_atmosphere_model_state.jl
@@ -240,7 +240,7 @@ function compute_auxiliary_thermodynamic_variables!(model::AtmosphereModel)
             model.moisture_density)
 
     fill_halo_regions!(model.temperature)
-    fill_halo_regions!(model.microphysical_fields)
+    fill_halo_regions!(model.microphysical_fields, boundary_condition_args(model)...)
     fill_halo_regions!(model.formulation)
 
     return nothing
--- a/src/CompressibleEquations/CompressibleEquations.jl
+++ b/src/CompressibleEquations/CompressibleEquations.jl
@@ -60,6 +60,7 @@ using Oceananigans.Grids: rnode, znode
 using Oceananigans.Operators: ℑxᶜᵃᵃ, ℑxᶠᵃᵃ, ℑyᵃᶜᵃ, ℑyᵃᶠᵃ, ℑzᵃᵃᶜ, ℑzᵃᵃᶠ,
                                 ∂zᶜᶜᶜ, ∂zᶜᶜᶠ
 using Oceananigans.BoundaryConditions: fill_halo_regions!
+using Oceananigans.Models: boundary_condition_args
 using Oceananigans.Operators: divᶜᶜᶜ
 using Oceananigans.Utils: prettysummary, launch!, KernelParameters
 
--- a/src/CompressibleEquations/compressible_time_stepping.jl
+++ b/src/CompressibleEquations/compressible_time_stepping.jl
@@ -374,9 +374,11 @@ function AtmosphereModels.compute_auxiliary_dynamics_variables!(model::Compressi
     dynamics = model.dynamics
 
     # Ensure halos are filled (may have been async from update_state!)
-    # These fields are needed for pressure computation via equation of state
-    fill_halo_regions!(dynamics.dry_density)
-    fill_halo_regions!(prognostic_fields(model.formulation))
+    # These fields are needed for pressure computation via equation of state.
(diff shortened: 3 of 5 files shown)


glwagner:
I don't think any auxiliary fields can have custom boundary conditions that would require fields, clock. Only the prognostic variables can handle those?

PR #954, opened 2026-09-03
Expose the hard-coded P3 empirical parameters

glwagner:
Promotes the sixteen empirical scalars buried in P3's cloud-width, rain-ventilation and
rain-fall-speed formulas into three concrete, immutable, `isbits` parameter containers that
are visible from the public constructors and reach every table, kernel and rate that uses
them. Defaults and physics are unchanged.

```julia
cloud = CloudDroplets(Float64;
    shape = CloudShape(Float64;
        relative_dispersion_number_coefficient = 5.714e-10,
        relative_dispersion_intercept = 0.2714,
        minimum_shape_parameter = 2,
        maximum_shape_parameter = 15))

rain = RainDrops(Float64;
    fall_speed = RainFallSpeed(Float64;
        branch_velocity_scales = (4579.5, 49.62, 17.32),
        branch_mass_exponents = (2/3, 1/3, 1/6),
        transition_diameters = (134.43e-6, 1511.64e-6, 3477.84e-6),
        plateau_velocity = 9.17),
    ventilation = RainVentilation(Float64;
        constant_coefficient = 0.78,
        reynolds_coefficient = 0.32))

p3 = P3Microphysics(Float64; cloud, rain)
```

## The sixteen promoted scalars

| Container | Field | Default | Meaning |
|---|---|---:|---|
| `CloudShape` | `relative_dispersion_number_coefficient` | `5.714e-10` m³ | coefficient on Nᶜˡ in the Liu–Daum relative-dispersion relation, in SI |
| `CloudShape` | `relative_dispersion_intercept` | `0.2714` | intercept of that relation |
| `CloudShape` | `minimum_shape_parameter` | `2` | lower bound on the diagnosed μᶜˡ |
| `CloudShape` | `maximum_shape_parameter` | `15` | upper bound on the diagnosed μᶜˡ |
| `RainVentilation` | `constant_coefficient` | `0.78` | f₁ᵣ, the still-air ventilation term |
| `RainVentilation` | `reynolds_coefficient` | `0.32` | f₂ᵣ, multiplying Sc^(1/3) Re^(1/2) |
| `RainFallSpeed` | `branch_velocity_scales` | `(4579.5, 49.62, 17.32)` m/s | three power-law velocity scales |
| `RainFallSpeed` | `branch_mass_exponents` | `(2/3, 1/3, 1/6)` | three mass exponents |
| `RainFallSpeed` | `transition_diameters` | `(134.43, 1511.64, 3477.84)` μm | ordered branch boundaries |
| `RainFallSpeed` | `plateau_velocity` | `9.17` m/s | large-drop terminal-speed plateau |

`4 + 2 + 3 + 3 + 3 + 1 = 16`. In the manuscript's notation these are `C_{μcl}`, `C_vent`
and `C_Vr`; the runtime process-level count rises from 63 to 79. Table-generator parameters
remain separate.

The velocity scales are the published cm s⁻¹ coefficients converted to SI. The mass
argument stays the dimensionless ratio `drop_mass / one_gram`, numerically identical to the
old "mass in grams".

## What now reads the configured values

**Cloud width.** `liu_daum_shape_parameter(Nᶜˡ, shape)` is the only implementation.
All three call sites pass `p3.cloud.shape`:

1. the construction-time diagnosis in `CloudDroplets`;
2. the prognostic `diagnose_cloud_dsd`; and
3. the cloud immersion-freezing PSD correction in `immersion_freezing_cloud_rate`.

A one-argument `liu_daum_shape_parameter(Nᶜˡ)` remains as an interactive convenience that
builds the default container. No model path calls it.

**Rain ventilation.** `RAIN_VENTILATION_CONSTANT` and `RAIN_VENTILATION_REYNOLDS` are gone.
`rain_ventilation_integral(table, ventilation, …)` takes the pair explicitly, and
both consumers — direct rain evaporation and the coupled saturation-adjustment relaxation
coefficient — pass `p3.rain.ventilation`. The coefficients deliberately stay out of the
tabulated `I_VD` and are applied at runtime, so they survive materialization.

**Rain fall speed.** `rain_fall_speed(D, ρ_correction, fall_speed)` carries the whole
piecewise law. All three startup quadratures (`RainMassWeightedVelocity`,
`RainNumberWeightedVelocity`, `RainVelocityDiameterIntegral`) hold the concrete parameter
object and pass it, so one configured law feeds the mass-weighted velocity, the
number-weighted velocity, and the evaporation velocity-diameter table.

**Construction path.** `PredictedParticlePropertiesMicrophysics` and `read_lookup_tables`
gained a `rain` keyword. `read_lookup_tables` no longer discards it, and
`tabulate_rain_from_quadrature` replaces only the three lookup placeholders, preserving every
supplied physics parameter.

## API migration

- `RainDrops.fall_speed_coefficient` and `fall_speed_exponent` are **removed**. They
  described a single power law `V = aᵥ D^bᵥ` that the active piecewise Gunn–Kinzer/Beard
  calculation never read: they were only stored, displayed, copied at materialization, and
  asserted in tests. They are replaced by `RainDrops.fall_speed`, a
  `RainFallSpeed` that does control the law. `Base.show` now reports the active
  law instead of the obsolete pair.
- `CloudDroplets` gains a `shape` field (and keyword). The existing
  `shape_parameter` keyword still overrides the construction-time value only; the docstring
  now says so explicitly, because that override never controlled the two local-number paths.
### Naming

Types and the objects instantiated from them no longer carry a `Parameters` or `Properties`
suffix, and struct fields are documented with inline comments rather than per-field
docstrings.

| Before | After |
|---|---|
| `CloudDropletProperties` | `CloudDroplets` |
| `RainProperties` | `RainDrops` |
| `IceProperties` | `IceParticles` |
| `IceBulkProperties` | `IceBulk` |
| `ProcessRateParameters` | `ProcessRate` |
| `CloudShapeParameters` | `CloudShape` |
| `RainFallSpeedParameters` | `RainFallSpeed` |
| `RainVentilationParameters` | `RainVentilation` |
| `RainMassWeightedVelocityEvaluator` | `RainMassWeightedVelocity` |
| `RainNumberWeightedVelocityEvaluator` | `RainNumberWeightedVelocity` |
| `RainEvaporationVentilationEvaluator` | `RainVelocityDiameterIntegral` |

and for fields, keywords and local bindings: `shape_parameters` → `shape`,
`ventilation_parameters` → `ventilation`, `fall_speed_parameters` → `fall_speed`,
`bulk_properties` → `bulk`, `n_points` → `points`.

Three of these are more than suffix-stripping:

- **`CloudDroplets`, `RainDrops`, `IceParticles`.** These containers describe particle
  *populations* — number concentration and PSD shape for cloud, fall speed and ventilation
  for rain, rime and collection properties for ice — and the docstrings already used that
  vocabulary: *"Ice particles in P3 span a continuum from small pristine crystals to large
  heavily-rimed graupel"*, *"Cloud droplets in P3 are treated simply"*. Naming the particle
  rather than the substance also avoids the mass-noun problem that bare `Rain`/`Ice` have,
  since neither pluralises. `drops` rather than `droplets` for rain, because *droplet* is
  reserved for cloud-sized (≲ 100 μm). The field names stay `cloud`, `rain` and `ice`:
  they name the category, not the type, and they are what appears at the use sites.
- **`RainVelocityDiameterIntegral`.** The old name promised the ventilation integral, but
  the type computes only ``I_{VD} = ∫ D \sqrt{V D} e^{-λ^r D} dD``; the ventilation
  coefficients are deliberately applied at runtime, not baked into the table.
- **`points`, not `n_points`.** Matches `TabulatedFunction(...; points = 100)` in
  Oceananigans, which is what these objects are fed into.

`PredictedParticleProperties` and `PredictedParticlePropertiesMicrophysics` are deliberately
**not** renamed: "Predicted Particle Properties" is the published name of the scheme, not a
container suffix. The same convention applied outside the microphysics module is split into
a separate PR (#957) to keep this one focused.

Two follow-ups worth a maintainer's opinion, not done here:
- `ProcessRate` (the parameter container) now sits close to `P3ProcessRates` (the computed
  per-cell rates). The field name `p3.process_rates` disambiguates, but the pair may deserve
  a better split.
- The file names still read `rain_properties.jl`, `ice_properties.jl`,
  `cloud_droplet_properties.jl`, `process_rate_parameters.jl`. Renaming them was left out to
  keep the diff reviewable.

### Formatting

The renames shortened many identifiers by ~10 characters, which orphaned the continuation
lines aligned under them. Those are re-aligned (verified by diffing an alignment report
against `origin/main`, so only regressions this PR introduced were touched), and argument
lists are joined onto one line where they fit.

- `RainDrops.maximum_mean_diameter` is **kept but documented as inactive**. It is not
  one of the sixteen empirical literals, and no active source reads it — the rain spectrum is
  bounded by `ProcessRate.minimum_rain_slope` / `maximum_rain_slope` through
  `rain_slope_parameter`. Removing it is an API change the brief leaves to maintainer
  discretion, so it is flagged here rather than taken unilaterally; the docstring no longer
  claims it bounds the rain spectrum. **Follow-up issue needed** to remove it.

## Deliberately excluded

Not promoted to the calibration API, and documented as such in `p3_theory.md`:

- `GUNN_KINZER_WATER_DENSITY = 997` kg m⁻³ — the reference basis of the published fit;
  substituting the configurable ρʷ would rescale the published coefficients.
- `π/6`, one gram, and cm↔m conversions — definitions and unit conversions.
- Quadrature nodes, quadrature point count, lookup-axis ranges, interpolation resolution —
  numerical controls.
- Numerical floors and sink-limiter settings.
- Coefficients baked into the external ice lookup-table artifact — those belong to the
  separate, versioned table-generator effort.

## Numerics

Validation is host-side, in the constructors: nonnegative number coefficient, positive
dispersion intercept, `minimum_shape_parameter ≤ maximum_shape_parameter`, nonnegative
velocity scales / plateau / mass exponents / ventilation coefficients, and strictly
increasing positive transition diameters. Kernels stay branch-free and allocation-free.

`rain_fall_speed` keeps nested `ifelse` for the four regimes, but selects the branch scale
and exponent *before* the power, so the parameterized law costs one generic `^` where the
pinned code used up to three specialized `cbrt`/`sqrt` calls. Measured on this machine
(`Float64`, host): 6.43 ms vs 6.22 ms per 200k evaluations — 3% slower per call, and
allocation-free. The whole startup tabulation is 3 tables × 200 points × 128 nodes =
76 800 evaluations, so ≈ 2.5 ms against ≈ 2.4 ms before. `rain_fall_speed` runs only in that
startup tabulation, never in a device kernel; the runtime ventilation path uses only scalar
coefficients, `cbrt` and `sqrt`.

Default output is **not** bitwise identical, for two documented reasons:

1. The cloud coefficient is now stated in SI (`5.714e-10` m³) rather than as `0.0005714`
   applied to `Nᶜˡ × 1e-6`. Algebraically identical; the roundings of χ differ in the last
   bits, and `μ = 1/χ² - 1` amplifies a relative perturbation of χ by ≈ 2(μ+1)/μ.
2. Configurable exponents force `m^b` in place of `cbrt(m)^2`, `cbrt(m)` and `sqrt(cbrt(m))`.
   `b` is not exactly representable, giving a relative error of order `|b ln m| × ulp(b)/b`:
   ≈ 3e-15 in `Float64` and ≈ 2e-6 in `Float32` over the tabulated size range.

Parity is asserted at `rtol = 1e-13` (`Float64`) and `2e-5` (`Float32`) pointwise, and
`1e-12` / `5e-5` for the integrals — tight enough to catch a changed unit, a flipped branch
inequality, or a dropped parameter, all of which move results by percent or more.

## Tests

New `test/predicted_particle_properties_empirical_parameters.jl`, both `Float32` and
`Float64` throughout:

- **Default parity.** `liu_daum_shape_parameter` and `rain_fall_speed` are checked against
  verbatim copies of the pinned formulas at marine, continental and both bound-active
  concentrations, and below / at / immediately above every transition diameter. The
  quadrature results are checked against numbers recorded from the pre-refactor
  implementation across the complete tabulated slope range, plus the full ventilation
  integral at six (qʳ, nʳ) states.
- **Plumbing and sensitivity.** Custom cloud coefficients move the construction-time
  `shape_parameter`, the prognostic `diagnose_cloud_dsd`, and immersion freezing; doubling
  every fall-speed scale doubles both velocity tables and scales `I_VD` by √2; zeroing each
  ventilation coefficient removes exactly its term from the integral, from evaporation, and
  from the coupled relaxation coefficient; `P3Microphysics(FT; rain = custom)` retains the
  custom values after `read_lookup_tables`.
- **Full coverage.** Each of the sixteen scalars is perturbed individually and shown to
  change a physically relevant result in a regime where it is active — including the upper
  shape bound, which never binds at the default intercept (`1/0.2714² - 1 ≈ 12.58 < 15`) and
  is therefore exercised by lowering it into range.
- **Type stability and architecture.** `isbits`, concrete field types, `@inferred` on the
  three parameterized helpers, allocation-free kernel helpers, no `Float64` promotion in a
  `Float32` scheme, and a stepped model on `default_arch` after `on_architecture` adaptation.

Existing P3 tests were updated for the new signatures and the removed fields.

## Documentation

Constructor docstrings for the three new containers, `CloudDroplets`,
`RainDrops` and `PredictedParticlePropertiesMicrophysics`, all with `jldoctest` blocks,
`$(TYPEDSIGNATURES)` and `$(TYPEDFIELDS)`; a new "Empirical Warm-Phase Coefficients" section
and a "Rain Terminal Velocity" subsection in `p3_theory.md` with the corrected field mapping;
a worked custom-parameter example in `p3_usage.md`; and the three new types exported.

---

https://claude.ai/code/session_01TaBjuhm4zDo2fK4ZN7eMpk


src/Microphysics/PredictedParticleProperties/cloud_droplet_properties.jl
```diff
@@ -21,48 +127,63 @@ end
...
 $(TYPEDSIGNATURES)
 
+Convenience wrapper evaluating [`liu_daum_shape_parameter`](@ref) with the default
+[`CloudShapeParameters`](@ref).
+
+Provided for interactive use only. No prognostic or immersion-freezing path may call it:
+those must read `p3.cloud.shape_parameters` so that a configured fit is actually used.
+"""
+@inline liu_daum_shape_parameter(Nᶜˡ) =
+    liu_daum_shape_parameter(Nᶜˡ, CloudShapeParameters(typeof(float(Nᶜˡ))))
```
glwagner:
```suggestion
@inline liu_daum_shape_parameter(Nᶜˡ) = liu_daum_shape_parameter(Nᶜˡ, CloudShapeParameters(typeof(float(Nᶜˡ))))
```

src/Microphysics/PredictedParticleProperties/coupled_saturation_adjustment.jl
```diff
@@ -110,7 +110,8 @@ end
     nʳ_eff = max(nʳ, FT(p3.minimum_number_mixing_ratio))
     active = qʳ_eff >= p3.minimum_mass_mixing_ratio
 
-    ventilation = rain_ventilation_integral(p3.rain.evaporation, qʳ_eff, nʳ_eff,
+    ventilation = rain_ventilation_integral(p3.rain.evaporation, p3.rain.ventilation,
+                                           qʳ_eff, nʳ_eff,
```
glwagner:
```suggestion
    ventilation = rain_ventilation_integral(p3.rain.evaporation, p3.rain.ventilation, qʳ_eff, nʳ_eff,
```

src/Microphysics/PredictedParticleProperties/lookup_table_reader.jl
```diff
@@ -206,9 +215,17 @@ function tabulate_rain_from_quadrature(rain::RainProperties, arch=CPU(),
...
+    # law reaches the mass-weighted velocity, the number-weighted velocity, and the
+    # evaporation velocity-diameter table alike.
+    fall_speed = convert(RainFallSpeedParameters{FT}, rain.fall_speed)
+
+    vel_mass_eval = RainMassWeightedVelocityEvaluator(FT; n_points=quadrature_points,
+                                                      floors, fall_speed)
+    vel_num_eval = RainNumberWeightedVelocityEvaluator(FT; n_points=quadrature_points,
+                                                       floors, fall_speed)
+    evap_eval = RainEvaporationVentilationEvaluator(FT; n_points=quadrature_points,
+                                                    fall_speed)
```
glwagner:
```suggestion
    vel_mass_eval = RainMassWeightedVelocityEvaluator(FT; n_points=quadrature_points, floors, fall_speed)
    vel_num_eval = RainNumberWeightedVelocityEvaluator(FT; n_points=quadrature_points, floors, fall_speed)
    evap_eval = RainEvaporationVentilationEvaluator(FT; n_points=quadrature_points, fall_speed)
```

src/Microphysics/PredictedParticleProperties/rain_process_rates.jl
```diff
@@ -257,7 +257,8 @@ The evaporation rate is enhanced by ventilation (air flow around falling drops).
 
     # Internal helpers return negative (S - 1 < 0 when subsaturated).
     # Negate to get positive magnitude.
-    evap_rate = -rain_evaporation_rate(p3.rain.evaporation, qʳ_eff, nʳ_eff, S,
+    evap_rate = -rain_evaporation_rate(p3.rain.evaporation, p3.rain.ventilation,
+                                        qʳ_eff, nʳ_eff, S,
```
glwagner:
```suggestion
    evap_rate = -rain_evaporation_rate(p3.rain.evaporation, p3.rain.ventilation, qʳ_eff, nʳ_eff, S,
```

src/Microphysics/PredictedParticleProperties/rain_process_rates.jl
```diff
@@ -280,14 +281,17 @@ I_{evap}(λ^r) = \\frac{f_{1r}}{(λ^r)^2}
...
 `(; λʳ, Nʳ₀, integral)`, since every caller needs the intercept
 ``N^r_0 = n^r λ^r`` alongside the integral.
 
 Consumed by [`rain_evaporation_rate`](@ref) and by the coupled
-saturation-adjustment relaxation coefficient.
+saturation-adjustment relaxation coefficient, both of which pass `p3.rain.ventilation`.
 """
-@inline function rain_ventilation_integral(table, qʳ, nʳ, ν, Dᵛ, parameters)
+@inline function rain_ventilation_integral(table, ventilation_parameters,
+                                           qʳ, nʳ, ν, Dᵛ, parameters)
```
glwagner:
```suggestion
@inline function rain_ventilation_integral(table, ventilation_parameters, qʳ, nʳ, ν, Dᵛ, parameters)
```

src/Microphysics/PredictedParticleProperties/cloud_droplet_properties.jl
```diff
@@ -4,12 +4,113 @@
...
+
+    relative_dispersion_number_coefficient ≥ 0 ||
+        throw(ArgumentError("relative_dispersion_number_coefficient must be nonnegative, " *
+                            "got $relative_dispersion_number_coefficient"))
+    relative_dispersion_intercept > 0 ||
+        throw(ArgumentError("relative_dispersion_intercept must be positive, " *
+                            "got $relative_dispersion_intercept"))
+    minimum_shape_parameter ≤ maximum_shape_parameter ||
+        throw(ArgumentError("minimum_shape_parameter must not exceed maximum_shape_parameter, " *
+                            "got $minimum_shape_parameter and $maximum_shape_parameter"))
```
glwagner: let's make this more compact

src/Microphysics/PredictedParticleProperties/cloud_droplet_properties.jl
glwagner: Applied — now a single line. Reads \`liu_daum_shape_parameter(Nᶜˡ) = liu_daum_shape_parameter(Nᶜˡ, CloudShape(typeof(float(Nᶜˡ))))\` after the `Parameters`-suffix rename.

src/Microphysics/PredictedParticleProperties/coupled_saturation_adjustment.jl
glwagner: Applied. The local holding the returned NamedTuple is now `integrals` rather than `ventilation`, so the argument could take the name `ventilation` without shadowing it.

src/Microphysics/PredictedParticleProperties/lookup_table_reader.jl
glwagner: Applied — all three evaluator constructions are one line each. They also lost the `Evaluator` suffix and `n_points` is now `points`, matching `TabulatedFunction`.

src/Microphysics/PredictedParticleProperties/rain_process_rates.jl
glwagner: Kept as you applied it; the second argument is now named `ventilation` rather than `ventilation_parameters`.

src/Microphysics/PredictedParticleProperties/rain_process_rates.jl
glwagner: Kept as you applied it, with `ventilation_parameters` renamed to `ventilation`.

src/Microphysics/PredictedParticleProperties/cloud_droplet_properties.jl
glwagner: Compacted: the three checks are one line each now, using short local aliases (`a`, `b`, `μ_min`, `μ_max` — the same symbols the docstring uses) so the messages fit without string concatenation. Did the same to the `RainFallSpeed` and `RainVentilation` constructors so the three read alike.

kaiyuan-cheng (approved):

PR #957, opened 2026-09-03
Drop Parameters/Properties suffixes outside Microphysics

glwagner:
Drops the `Parameters` / `Properties` suffix from types and from the objects instantiated
from them, everywhere outside Breeze's microphysics. Split out of #954 so that PR stays
focused on the P3 empirical-parameter work.

Pure renaming — no behaviour change.

## Types

| Before | After | Module |
|---|---|---|
| `StabilityFunctionParameters` | `StabilityFunction` | `BoundaryConditions` |
| `SurfaceRadiativeProperties` | `SurfaceRadiation` | `AtmosphereModels` (used by `BreezeRRTMGPExt`) |

## Fields, keywords and local bindings

| Before | After |
|---|---|
| `stability_function_parameters` | `stability_function` |
| `surface_properties` | `surface_radiation` |
| `air_properties` | `air` |
| `activation_parameters` | `activation` |
| `grid_parameters` | `rrtmgp_grid` |
| `disabled_parameters` | `disabled_microphysics` |

`air` also lines up with its sibling fields in `FourCategories` (`cloud_liquid`, `cloud_ice`,
`rain`, `snow`, ...).

## What was deliberately left alone

- **Upstream fields.** `Microphysics1MParams.air_properties` belongs to CloudMicrophysics, so
  the seven `parameters.air_properties` accesses are untouched. Only the field Breeze owns
  was renamed.
- **Function names** — `air_transport_properties`, `p3_process_properties`,
  `p3_fall_speed_properties`, `p3_core_ice_properties`, `filtered_kernel_parameters`.
- **File names** — `rain_properties.jl`, `ice_properties.jl`, `transport_properties.jl`, ...
- **External interface names** — `pdf_rain_parameters` (imported from CloudMicrophysics) and
  `default_included_properties` (an Oceananigans method Breeze extends).

Worth a second opinion: `air`, `activation` and `rrtmgp_grid` hold types owned by
CloudMicrophysics and RRTMGP (`AirProperties`, `AerosolActivationParameters`,
`RRTMGPGridParams`), so those local names no longer mirror the upstream type names.

## Formatting

Continuation lines whose alignment the renames disturbed are re-aligned, and the argument
lists they belong to are joined where they fit on one line.

## Tests

`polynomial_bulk_coefficients`, `gray_radiative_transfer`,
`cloud_microphysics_1M_warm_ice_deposition`, `cloud_microphysics_2M` and
`quality_assurance` all pass (409 tests). `quality_assurance` covers ExplicitImports and
Aqua, which is what would catch a missed export or a stale reference.

---

https://claude.ai/code/session_01TaBjuhm4zDo2fK4ZN7eMpk


kaiyuan-cheng (approved):
looks good

PR #614, opened 2026-04-07
sedimentation_velocity interface + effective liquid/ice velocities

kaiyuan-cheng:
## Summary

- Introduce `sedimentation_speed` interface (positive magnitude) replacing `microphysical_velocities` as the primary interface for microphysics schemes
- Add `water_phase` for liquid/ice phase classification of mass tracers
- Add `NegatedField` wrapper and generic `microphysical_velocities` that calls `sedimentation_speed`
- Precompute effective sedimentation velocities `𝕎ᴸ` (total liquid) and `𝕎ᴵ` (total ice) on `AtmosphereModel.effective_sedimentation_velocities`
- Change sign convention: terminal velocity fields (`wʳ`, `wᶜˡ`, etc.) now store positive magnitudes

Overhauls #458 — computes effective velocities for total liquid and total ice separately instead of total moisture.

## Test plan

- [x] 1M microphysics tests pass (90/90)
- [x] 2M microphysics tests pass (82/82)
- [ ] Full CI

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #959, opened 2026-09-03
Unify sedimentation transport and correct condensate thermal coupling

kaiyuan-cheng:
## Summary

Unify sedimentation across the one-moment, two-moment, and P3 schemes and couple the falling condensate mass to the thermodynamic variable. Each scheme supplies a signed fall velocity and a phase for each disjoint condensate mass. P3 liquid coating carries liquid enthalpy at the ice fall speed; rime and particle-number/volume moments do not count as additional condensate mass. The same constituent interface supplies the surface precipitation diagnostic.

The compressible sedimentation response now separates the local composition derivative from the enthalpy crossing a face. The derivative accounts for renormalizing mass fractions as total density changes at fixed dry density. Transport uses phase enthalpy, so equal-temperature cells with different compositions no longer acquire spurious sedimentation heating. The temperature conversion includes the pressure response appropriate to an isolated fixed-volume cell with fixed gas partial densities:

`beta_cv = [1 - (R_m / cp_m) * (1 - Lambda / (cp_m * T))] / (cv_m * Pi)`,

where `Lambda = q_l L_l + q_i L_i`, `cv_m = cp_m - R_m`, and `Pi` is the Exner function.

The anelastic core retains its fixed-density dry-air replacement convention, transported weight `h_x - h_d`, and prescribed-pressure coefficient `1 / (cp_m * Pi)`. Changing only its transported weight to phase enthalpy would introduce a different dry-air energy exchange. This PR does not change that convention or add a compressible static-energy diagnosis.

## Discretization and limits

The thermodynamic correction uses each constituent's actual mass-flux difference between resolved-plus-fall transport and resolved transport, with the corresponding upwind donor temperature and local composition response. Bounds-preserving WENO follows the pinned Oceananigans 0.111 reconstruction; the copied helper requires separate migration work for newer operators.

For adaptive implicit vertical advection, both time steppers apply the thermal correction using the mass-flux remainder from the implicit scalar solve, including bottom outflow. The potential-temperature correction remains an instantaneous derivative multiplied by a finite increment. It is not an exact finite-step energy reconstruction: cellwise temperature errors can remain even when a column budget cancels. The PR makes no whole-model total-energy conservation claim.

## Validation

- New isolated temperature/budget regressions: 144 assertions pass in each of CPU Float32 and Float64, covering both phases, warm/cold/isothermal donors, both cores, open/closed bottoms, and no-fall controls. The same Float64 test on original production source fails 24 assertions in the 12 falling compressible configurations.
- Independent Claude review and coordinator two-cell budgets agree on the compressible coefficient and preserved anelastic convention. Finite explicit and implicit errors are recorded separately from instantaneous-rate success.
- Actual candidate CUDA regressions pass 144/144 assertions in each of Float32 and Float64 on a T4, with scalar indexing disabled (job 1264, clean commit 31f207f).
- Completed one-moment coverage: 458 pass / 2 fail across 460 assertions and 62 precision testsets. Both failures are unchanged Float32 StaticEnergy distinction assertions that reproduce with identical values and tolerances on original production source. No assertions or tolerances were weakened.
- Core ambiguity, unbound-parameter, export, dependency/compatibility, and ExplicitImports checks passed. The initial Aqua persistent-task failure did not reproduce in an isolated check with unchanged defaults; both candidate and original source then passed. Full optional-extension QA remains incomplete: a 900-second timeout occurred during Reactant import, before its tests. No full CI/QA pass is claimed.
- The broader baseline campaign and bulk results are kept on [the separate validation branch](https://github.com/NumericalEarth/Breeze.jl/tree/9fe61081c6b1c98a0f572d0bebb66c2d4718b73c/validation/sedimentation_energy/results); its requested CUDA baseline matrix is complete (85 isolated cases and eight coupled cases per precision; Float32 isolated assembled from disjoint 61+24-case runs). Diagnostic callback substitutions and finite-state assertions there are not candidate validation or conservation proofs. Separate caloric-reference and ordinary heating findings are tracked in #1009 and #1010.

Close #591 #614 #364 #383


docs/src/appendix/notation.md
```diff
@@ -64,6 +64,9 @@ The following table also uses a few conventions that suffuse the source code and
 | ``qⁱ``                              | `qⁱ`   | `AM.microphysical_fields.qⁱ`        | Ice mass fraction                                                              |
 | ``qᶜˡ``                             | `qᶜˡ`  | `AM.microphysical_fields.qᶜˡ`       | Cloud liquid mass fraction                                                     |
 | ``qᶜⁱ``                             | `qᶜⁱ`  | `AM.microphysical_fields.qᶜⁱ`       | Cloud ice mass fraction                                                        |
+| ``q^f``                             | `qᶠ`   | `AM.microphysical_fields.qᶠ`        | Rime mass fraction (P3); a portion of ``qⁱ``, not additional mass              |
+| ``b^f``                             | `bᶠ`   | `AM.microphysical_fields.bᶠ`        | Rime volume per unit mass (P3), m³/kg                                          |
```
glwagner, line 68: "b" can be confused with "buoyancy" (used elsewhere I think?) but we can address this later. A script V or \upsilon could work better here for a volume variable

docs/src/appendix/notation.md
```diff
@@ -89,6 +92,14 @@ The following table also uses a few conventions that suffuse the source code and
...
+| ``w^r``                             | `wʳ`   |                                     | Rain sedimentation velocity field (vertical component; negative for downward motion) |
+| ``w^s``                             | `wˢ`   |                                     | Snow sedimentation velocity field (vertical component; negative for downward motion) |
+| ``w^i``                             | `wⁱ`   |                                     | Ice mass-weighted sedimentation velocity field (P3; negative for downward motion) |
+| ``w^{cl}_n``                        | `wᶜˡₙ` |                                     | Cloud liquid number-weighted sedimentation velocity field (negative for downward motion) |
+| ``w^r_n``                           | `wʳₙ`  |                                     | Rain number-weighted sedimentation velocity field (negative for downward motion) |
+| ``w^i_n``                           | `wⁱₙ`  |                                     | Ice number-weighted sedimentation velocity field (P3; negative for downward motion) |
```
glwagner: how about {ncl}, {nr}, and {ni} for these?

docs/src/developer/microphysics/overview.md
```diff
@@ -113,11 +113,89 @@ iterative adjustment to partition moisture between vapor and condensate.
 | Auxiliary/Diagnostic | `CenterField` | None needed | `qᵛ`, `qˡ`, `qᶜˡ`, `qʳ` |
 | Velocities | `ZFaceField` | `bottom=nothing` | `wʳ`, `wᶜˡ`, `wʳₙ` |
 
-### Velocity and Humidity Functions
+### Sedimentation
+
+| Function | Arguments | Description |
+|----------|-----------|-------------|
+| `sedimentation_velocity` | `(microphysics, microphysical_fields, name)` | **Primary interface**: return the vertical sedimentation velocity field for tracer `name`, or `nothing` |
+| `condensate_phase` | `(microphysics, name)` | Return `Val(:liquid)` or `Val(:ice)`, the thermodynamic phase of condensate mass `name`; required for every sedimenting mass |
```
glwagner: how does this work for condensates that represent a mixed phase?

docs/src/developer/microphysics/overview.md
```diff
@@ -113,11 +113,89 @@ iterative adjustment to partition moisture between vapor and condensate.
...
+| `microphysical_velocities` | `(microphysics, microphysical_fields, name)` | **Generic wrapper** (don't override): wraps the sedimentation velocity in a velocity tuple |
+
+**Design principle**: Schemes implement `sedimentation_velocity` (how fast a tracer falls) and
+`condensate_phase` (which latent heat its mass carries); the generic `microphysical_velocities`
```
glwagner, line 125: would it be better to use an interface `latent_height(microphysics, constants, name)` ? that way arbitrary mixed phase condensates may be defined.

docs/src/developer/microphysics/overview.md
```diff
@@ -113,11 +113,89 @@ iterative adjustment to partition moisture between vapor and condensate.
...
+implicit vertical advection the tendency carries the content of the explicit fraction of each
+mass flux only; after the tracers' implicit solves of every stage the time steppers call
+`implicit_sedimentation_step!`, which moves the content of the remainder from the first-order
+fluxes the solves actually applied, at the solved state, so the heat follows the mass at any
+fall Courant number. Rain-out thus leaves latent warming aloft and pre-cools the layer that later
+evaporates the arriving rain, the mechanism that builds cold pools.
+
+### Surface Precipitation Flux
+
+`surface_precipitation_flux(model)` returns the flux of precipitating moisture through the
```
glwagner: would "bottom_precipitation_flux" be better here? The word "surface" is a bit overloaded, since we also have the concept of "droplet surface"

glwagner: Overloading "surface" for both a phase interface (air-ice or air-water surface, where we compute saturation specific humidity) or for the bottom of the atmosphere (eg air-LAND or air-OCEAN interface) also causes some confusion in boundary condition abstractions

src/Microphysics/PredictedParticleProperties/p3_microphysical_state.jl
```diff
@@ -290,17 +289,16 @@ function AM.materialize_microphysical_fields(p3::P3, grid, bcs)
...
-    # `w[i, j, Nz+1] = 0` so no precipitation falls in through the model top.
-    face_bcs = FieldBoundaryConditions(grid, (Center(), Center(), Face()); bottom=nothing)
-    wᶜˡ = ZFaceField(grid; boundary_conditions=face_bcs) # Cloud mass-weighted terminal velocity
-    wᶜˡₙ = ZFaceField(grid; boundary_conditions=face_bcs) # Cloud number-weighted terminal velocity
-    wʳ  = ZFaceField(grid; boundary_conditions=face_bcs)  # Rain mass-weighted terminal velocity
-    wʳₙ = ZFaceField(grid; boundary_conditions=face_bcs) # Rain number-weighted terminal velocity
-    wⁱ  = ZFaceField(grid; boundary_conditions=face_bcs)  # Ice mass-weighted terminal velocity
-    wⁱₙ = ZFaceField(grid; boundary_conditions=face_bcs) # Ice number-weighted terminal velocity
+    # so they live at z-Faces with the `AM.sedimentation_velocity_field` boundary conditions
+    # (the surface value is applied in `write_p3_fall_speeds!` rather than by halo filling,
```
glwagner: we may not want to reference function names in comments (they can go stale?)

src/Microphysics/PredictedParticleProperties/p3_microphysical_state.jl
```diff
@@ -843,42 +818,45 @@ with vapor, liquid (cloud + rain + liquid on ice), and ice components.
...
-@inline AM.microphysical_velocities(::P3, μ, ::Val{:ρbᶠ}) = (; u = ZeroField(), v = ZeroField(), w = μ.wⁱ)
-
-# Liquid on ice: same as ice mass
-@inline AM.microphysical_velocities(::P3, μ, ::Val{:ρqʷⁱ}) = (; u = ZeroField(), v = ZeroField(), w = μ.wⁱ)
+# diagnostic fields, so they are current whenever `update_state!` has run. The generic
+# `microphysical_velocities` wrapper turns each of these into the `(u, v, w)` NamedTuple
+# that the advection operator adds to the transport velocity.
+
+# Cloud mass and number: mass- and number-weighted Stokes fall speeds
+@inline AM.sedimentation_velocity(::P3, μ, ::Val{:ρqᶜˡ}) = μ.wᶜˡ
```
glwagner, line 827: I agree this is an improvement over `microphysical_velocities`. It does have the effect of restricting the added velocities to the third component

src/PotentialTemperatureFormulations/potential_temperature_tendency.jl
```diff
@@ -83,28 +85,94 @@ end
...
+#
+# with D = T − Π θ the latent deficit, Δcˣ = cˣ − cʳ, and cʳ, Rʳ, ℒʳ = ℒˡᵣ rˡ + ℒⁱᵣ rⁱ the heat
+# capacity, gas constant and latent deficit of the replacement (cᵖᵈ, Rᵈ, 0 for dry air; cᵖᵐ, Rᵐ,
+# cᵖᵐ D for the mixture). The first term is the deficit the falling condensate carries,
+# −ℒˣᵣ / (cᵖᵐ Π) to leading order; the rest accounts for the heat capacity and gas constant of
+# the mixture changing as condensate gives way to its replacement (lnΠ = (Rᵐ / cᵖᵐ) ln(p / pˢᵗ)
+# is written through Π so that every state type that defines an Exner function serves). The
+# shared `condensate_sedimentation_divergence` evaluates the content in each flux's upwind cell
+# and owns the discretization.
+@inline function theta_condensate_content(i, j, k, grid, formulation, dynamics, constants,
```
glwagner: what does "theta" mean here? potential temperature"? I suggest using the English name in the function def

glwagner:
> condensate_sedimentation_divergence (in AtmosphereModels) owns the discretization; the two
formulations supply only a content per unit falling mass.

Why this change? Does this mean that we do not properly upwind the velocities anymore? In general, this will prevent us from designing positivity preserving advection schemes and degrade monotonicity. Maybe I am not understand this correctly. I don't understand why this would be necessary.

src/PotentialTemperatureFormulations/potential_temperature_tendency.jl
kaiyuan-cheng: Yes, potential temperautre. Renamed.

docs/src/developer/microphysics/overview.md
kaiyuan-cheng: Added the support for truly mixed compositions, not the one used by P3 (where two condensates adherent to each other and fall at the same speed). The interface now accepts a liquid fraction in place of a pure phase.

docs/src/developer/microphysics/overview.md
kaiyuan-cheng: A mixture of latent heat alone is not enough. We still need heat contents

kaiyuan-cheng:
> > condensate_sedimentation_divergence (in AtmosphereModels) owns the discretization; the two
> > formulations supply only a content per unit falling mass.
> 
> Why this change? Does this mean that we do not properly upwind the velocities anymore? In general, this will prevent us from designing positivity preserving advection schemes and degrade monotonicity. Maybe I am not understand this correctly. I don't understand why this would be necessary.

Nothing changes about how the mass is transported. `scalar_tendency` still adds the constituent's fall velocity to the transport velocity and hands the sum to `div_ρUc` with the tracer's own scheme. 

What `condensate_sedimentation_divergence` does is to calculate the transport of of the thermodynamics content due to sedimentation. I think the calculation is consistent with dynamics for the following reasons:
1. the mass fluxes come from the tracer's own scheme
2. each flux carries the content of its own from its upwind cell
3. under the adaptive implicit advection only the explicit fraction is carried in the tendency, and the solver's remainder is moved afterwards from the fluxes the solve actually applied

docs/src/developer/microphysics/overview.md
glwagner: would another option be to ask for the latent heat of the condensate? instead of condensate_phase we would ahave `condensate_latent_heat(microphysics, constants, name)` .

docs/src/developer/microphysics/overview.md
glwagner: what do you mean?

glwagner:
> What `condensate_sedimentation_divergence` does is to calculate the transport of of the thermodynamics content due to sedimentation. I think the calculation is consistent with dynamics for the following reasons:
> 
> 1. the mass fluxes come from the tracer's own scheme
> 2. each flux carries the content of its own from its upwind cell
> 3. under the adaptive implicit advection only the explicit fraction is carried in the tendency, and the solver's remainder is moved afterwards from the fluxes the solve actually applied

I guess I don't understand how it is mathematically possible this is consistent. Upwinding is done inside the advective flux functions -- a right biased or left biased stencil is used depending on the sign of the velocity. If advection is performed in separate steps, it cannot possibly be correct from a mathematical point of view.

I also don't understand the advantage.

src/Advection.jl
```diff
@@ -40,10 +87,189 @@ end
...
+# the part of the transport the tendency applies; the first-order remainder the implicit solve
+# applies is moved with its content after the solve, from the solved state
+# (`AtmosphereModels.implicit_sedimentation_mass_fluxes`).
+
+@inline function AtmosphereModels.sedimentation_mass_fluxes(i, j, k, grid, advection, wᵗ, wˢ, q)
+    w = SumOfArrays{2}(wᵗ, wˢ)
+    F⁻ = (sedimentation_mass_flux(i, j, k,   grid, advection, w,  q),
+          sedimentation_mass_flux(i, j, k,   grid, advection, wᵗ, q))
+    F⁺ = (sedimentation_mass_flux(i, j, k+1, grid, advection, w,  q),
+          sedimentation_mass_flux(i, j, k+1, grid, advection, wᵗ, q))
```
glwagner, line 119:
```suggestion
    F⁻ = (_advective_tracer_flux_z(i, j, k,   grid, advection, w,  q),
          _advective_tracer_flux_z(i, j, k,   grid, advection, wᵗ, q))
    F⁺ = (_advective_tracer_flux_z(i, j, k+1, grid, advection, w,  q),
          _advective_tracer_flux_z(i, j, k+1, grid, advection, wᵗ, q))
```

glwagner: then can delete the function below (just to eliminate a little bit of indirection)

glwagner:
Added a version-pinned, reproducible CPU/GPU campaign in `18f08809`, CPU/Metal evidence in `83753f93`, and CUDA evidence/limits in `a03c3b7e`. **Production code is unchanged.** [Methodology and commands](https://github.com/NumericalEarth/Breeze.jl/blob/a03c3b7e5bd48cd66efb216434109de459bb26f2/validation/sedimentation_energy/README.md); [results and raw data](https://github.com/NumericalEarth/Breeze.jl/tree/a03c3b7e5bd48cd66efb216434109de459bb26f2/validation/sedimentation_energy/results).

Production baseline is `dae9e46d`, Oceananigans **0.111.0**. CPU64, CPU32 and Apple M5 Max Metal32 each completed 85 isolated cases; CPU64/Metal32 also completed eight coupled cases. Passing harness assertions do **not** mean the physical invariants pass.

- The original compressible isothermal tendency is **0.00200308 K/s liquid / 0.00314619 K/s ice**, confirmed on GPU. Phase enthalpy in the donor difference removes this defect; equal-composition/no-fall controls pass.
- For ±5 K donors, phase enthalpy alone gives **0.996512 / 0.996030** of the conditional fixed-volume internal-energy reference. Adding the diagnostic `beta_cv` matches that reference to roundoff (CPU64) / Float32 noise. The bottom is closed: the reference needs **incoming mass**, not net accumulation when there is outflow. Neither substitution is installed as a production fix.
- A **4×4×32, limiter-active** bounded-WENO test records mass nonconservation in **0.111**: about **−1.8058e-5 kg/m²/s**, while local thermal/mass consistency remains at roundoff with the diagnostic callback. Ordinary WENO's mass residual is roundoff. This does not apply to 0.112's shared-face implementation; the PR's copied limiter helper is a **rebase hazard** and must migrate with upstream.
- Finite implicit theta errors and original-callback coupled temperature drift remain. A near-zero column thermal residual can coexist with wrong local temperatures, so a global balance alone is insufficient.

No full compressible energy theorem is claimed. Corrected-callback coupled runs, finite compressible energy reconstruction, P3 mixed/differential-speed transport, gravity-on budgets and newer bounded operators remain **NOT RUN**. Compressible StaticEnergy is unsupported in this baseline and was not tested. The anelastic reference is explicitly conditional on the dry-replacement convention.

**CUDA outcome:** NVIDIA Tesla T4, Float64, Julia 1.12.6, CUDA.jl 6.4.0/runtime 12.9: all **85 isolated cases / 201 assertions completed**, matching CPU64. Script/environment hashes match the committed campaign. The 20-minute allocation then timed out before producing any coupled case results; CUDA32 was **NOT RUN**. No extension/new allocation was made. The full GPU matrix is therefore not complete; Metal32 and CUDA64 isolated evidence are complete.

glwagner:
Kept this PR focused on the sedimentation change. Commit `c4f44628` removes the 6,294 lines of campaign infrastructure, pinned environment, and raw result files added during the audit. The complete reproducible evidence is preserved on [`validation/pr959-energy-campaign-20260917`](https://github.com/NumericalEarth/Breeze.jl/tree/validation/pr959-energy-campaign-20260917/validation/sedimentation_energy), at immutable commit `a03c3b7e5bd48cd66efb216434109de459bb26f2`. The current tree exactly matches the reviewed sedimentation revision `dae9e46d`; this cleanup changes no production behavior, existing tests, or documentation, and uses no history rewrite.

The implementation should remain small and readable, with focused regression tests and short explanations of the mass/enthalpy identities. Broader thermodynamic reformulation, acoustic changes, general AIVA cleanup, campaign infrastructure, and raw logs belong outside this PR.

Two independently reproduced follow-ups are now tracked separately:

- #1009: caloric phase-energy offsets versus saturation latent heats.
- #1010: compressible volumetric-heating conversion to dry-density-weighted potential temperature.

Existing divergence-damping work is tracked in #793; boundary-flux Exner consistency in #976. These are not new requirements for the sedimentation implementation.

The scope reduction does **not** remove the sedimentation correctness requirements identified by the audit: transported phase enthalpy must be distinguished from a local composition derivative; a nonisothermal compressible claim needs the compatible isolated thermal response; and a finite-step consistency claim needs agreement with the actual mass/energy flux update. Small tests should expose those requirements. The current source still has the reported thermal defects, so the cleanup is not a claim that they are fixed or that full energy conservation is established. The existing [evidence summary](https://github.com/NumericalEarth/Breeze.jl/pull/959#issuecomment-5725187970) and its explicit limits remain applicable.

glwagner:
The matched precipitating-supercell GPU comparison is complete. All three runs reached7200s from13 bitwise-identical initial physical fields, with source/environment pins and restart checks recorded. The test is a compressible OneMoment warm-rain/saturation-adjustment variant,168×168×40,1km/500m,WENO9,Float32.

| Arm | Source | Oceananigans | Minimum sampled near-surface θρ anomaly | In-domain water increase |
|---|---|---|---:|---:|
|Main|8cc115e9bb15|0.113.0|−1.756K|24.781%|
|Common ancestor|2dbf930a808a|0.111.0|−2.233K|24.578%|
|PRhead|31f207f9f14b|0.111.0|−27.073K|9.765%|

**These runs do not validate a physically admissible cold pool or energy conservation.** All arms develop negative vapor/rain and share two pre-existing OneMoment defects, independently reviewed by Claude and traced against source:

- #1012: rain conversion has no opposing equilibrium-water source. A separate scalar CPU probe, without transport/sedimentation/time integration, confirms identical nonzero summed water sources in all three revisions; clear/no-rain controls are zero.
- #1013: diagnosed cloud subtracts rain from adjusted liquid that already excludes rain, producing a substantial prognostic/diagnostic partition mismatch.

The fixed branch's stronger cooling is a numerical trajectory difference under these defects. Base/fixed compares the entire PR; main/fixed also includes other source/dependency changes. The last thermal correction alone and the originally reported missing-cold-pool configuration are not isolated by this campaign. Energy plots are reconstructed inventories, not closed residuals; rainfall export is sampled.

These issues remain separate so this PR stays focused on the minimal sedimentation correction. The earlier isolated instantaneous thermodynamic-response evidence is unchanged. No production source, dependency bound, or PR diff was changed during the campaign.

bischtob:
NUM-233

PR #963, opened 2026-09-04
Various minor improvements to the docs

giordano:
Original motivation was https://github.com/NumericalEarth/Breeze.jl/pull/900#issuecomment-5471362221, but then I also wanted to have an overview page for the examples, that we can link from one of the big buttons of the main page, so I put all together


docs/src/literated/index.md
```diff
@@ -0,0 +1,18 @@
+# Examples
+
+The examples are the best way to learn Breeze.
```
ewquon, line 3: Love this. Very clearly written.

ewquon (approved):

Issue #913, opened 2026-08-18
Bounds-preserving WENO + AIVA double-counts the vertical flux (transports 1 + (1−s) times)


glwagner:
Found during review of #902's AIVA work; confirmed with the reproducer below on current main. Related record: #897.

## The bug

`WENO(...; bounds=(0,1), time_discretization=AdaptiveVerticallyImplicitDiscretization(...))` transports the vertical flux **1 + (1−s)** times. The bounds-preserving (positivity-limited) WENO path computes its explicit vertical divergence without ever applying the AIVA scale factor `s`, but `needs_implicit_solver` is true for the scheme, so the `(1−s)` first-order-upwind remainder is applied *on top of* an already-full explicit flux.

Plain (unbounded) WENO + AIVA applies `s` correctly, so this is specific to the bounds-preserving flux path.

## Measured signature

One split step of uniform vertical transport at vertical Courant α = 3 (uniform ρ and w, so the exact answer is rigid translation and the tracer centroid displacement measures exactly how much transport was applied; reference = 60 explicit substeps):

| case | Δcentroid (m) | min q |
|---|---|---|
| reference | 150.0 | 2.6e-46 |
| plain WENO5 + AIVA, explicit+implicit | **150.0** ✓ | 2.0e-23 |
| bounded WENO5 (0,1) + AIVA, explicit half only | 150.0 (should be s·150 = 25) | −6.8e-4 |
| bounded WENO5 (0,1) + AIVA, + implicit half | **275.0** ✗ | −5.4e-4 |

(The bounds-preserving scheme also goes slightly negative here, which is a separate irony.)

## Why it matters

Production AR configurations (`AR_AIVA=1` in AtmosphericRivers.jl) construct **every moisture tracer** with `bounds=(0,1)` + AIVA, so any such run transports nearly double the vertical moisture flux wherever the split engages.

## Reproducer

<details><summary>119-line script (runs on CPU in ~1 min)</summary>

```julia
# Quantify the claimed double count: one split step of vertical transport at α = 3.
#
# Prognostic Q = ρ q. Explicit half:  Q★ = Q⁰ - Δt div_ρUc(scheme, ρ, U, Q⁰/ρ)
# Implicit half: implicit_step!(Q★, ..., scheme, (; w), ρ)
#
# With uniform ρ and uniform w the exact answer is a rigid translation by w Δt, and the
# first moment of a conservative flux-form scheme translates at w Δt regardless of the
[...]

PR #964, opened 2026-09-04
Fold the thermodynamic AIVA remainder into the acoustic substep loop

### LINKED ISSUE #897: Adaptive implicit vertical advection NaNs immediately on a terrain-following nested domain

glwagner:
Fixes #897 (the reopened engaged-ρθ instability).

## The bug
AIVA on the thermodynamic variable corrupted the state within ~5 steps whenever the split engaged, on any grid and precision: the acoustic loop's pressure coupling integrated against a ρθ missing its CFL-withheld transport for the whole substep sequence, and applying the remainder post-loop was too late (and, per the bolt-on experiments recorded on #897, applying it naively per-substep destabilizes the Crank–Nicolson bookkeeping instead — every bolt-on variant died by density collapse at residual CFL an order below the previously hypothesized threshold).

## The fix
Fold the withheld transport into the loop's own structure, in two CN-consistent pieces:

1. **Base state** (`add_residual_base_tendency!`): the first-order upwind flux divergence of the frozen stage-entry (ρθ, ρᵈ) carried by wⁱ = (1−s)w joins `Gˢρθ` once per stage, so the predictors apply it per substep with the loop's own CN factors — density-weighted identically to the implicit Center-field coefficients, so base + perturbation sum to the full-field operator.
2. **Perturbation** (`in_loop_implicit_advection`): an implicit residual solve on the **θ predictor, between Step B and Step C** — inside the predictor-substitution chain, so Step C's pressure gradient and Step D's recovery always act on a transport-consistent ρθ′★.
3. Post-loop, the thermodynamic variable keeps only density-weighted closure diffusion (`implicit_step_scheme(nothing, diffusion_density)`).

The neglected cross-term (residual correction of the acoustic coupling coefficient) is O(Δτ²·residual), the order the loop already drops.

## Proof
| case | post-loop (main) | bolt-on (#897 experiments) | **this PR** |
|---|---|---|---|
| flat 5 K bubble, engaged (α ≈ 0.25) | dies step 4 | dies step 25 | **survives 200** |
| 600 m hill, uniform θ, engaged | dies step 5 | dies step 28–38 | **survives 200, θ drift 4 mK / 1000 s** |
| prototype stress hill, Δt = 25 | dies iter 4 | dies iter 16 (CN oscillation) | **survives 100** |
| disengaged / momentum-only regressions | pass | pass | **pass, θ-dev ≡ 0** |
| full AIVA suite | — | — | **108/108 in Float64 and Float32** |

The new testset pins the engaged-ρθ cases and the uniform-θ invariant — the coverage gap that let the original bug ship behind a green suite (every prior θ test ran in the explicit limit).

The history intentionally keeps the falsified bolt-on commit (`a961b55`) before the fix (`2178d9a`) so the negative result stays on the record.

Production context: momentum-only AIVA already delivers 1.56× wall-clock on a 72 h, 3 km hindcast at a raised Δt cap; the run sits at cap only 6% of the time because explicit-ρθ vertical CFL binds. This PR is what lifts that constraint; an A/B on the same event will quantify it.


src/TimeSteppers/acoustic_runge_kutta_3.jl
```diff
@@ -192,8 +192,17 @@ function acoustic_rk3_substep!(model::AtmosphereModel, Δt, β)
...
+    # tendency (a no-op unless the thermodynamic scheme is adaptive-implicit); the
+    # perturbation part is solved per substep inside the loop.
+    add_residual_base_tendency!(model)
+
+    # Linearized acoustic substep loop: Nτ substeps of size Δτ = Δt/N. An adaptive-implicit
+    # thermodynamic scheme is passed through so the loop applies the withheld perturbation
+    # transport per substep (`residual_predictor_substep!`).
+    θ_advection = field_advection_scheme(model.advection, thermodynamic_density_name(model.formulation))
+    loop_advection = needs_implicit_solver(θ_advection) ? θ_advection : nothing
+    acoustic_rk3_substep_loop!(model, substepper, Δt, β, U⁰, loop_advection)
```
glwagner:
```suggestion
    acoustic_rk3_substep_loop!(model, substepper, Δt, β, U⁰, θ_advection)
```

then use dispatch further down the chain to distinguish between the algorithm needed for AIVA vs the vanilla algorithm.

src/TimeSteppers/acoustic_runge_kutta_3.jl
```diff
@@ -192,8 +192,17 @@ function acoustic_rk3_substep!(model::AtmosphereModel, Δt, β)
     # compute_slow_momentum_tendencies! / compute_slow_scalar_tendencies!.
     compute_flux_bc_tendencies!(model)
 
-    # Linearized acoustic substep loop: Nτ substeps of size Δτ = Δt/N.
-    acoustic_rk3_substep_loop!(model, substepper, Δt, β, U⁰)
+    # Fold the CFL-withheld transport of the stage-entry thermodynamic state into the slow
+    # tendency (a no-op unless the thermodynamic scheme is adaptive-implicit); the
+    # perturbation part is solved per substep inside the loop.
+    add_residual_base_tendency!(model)
```
glwagner: I think "residual" could be a confusing term here, better to refer explicitly to the fact that this is an IMEX split.

src/TimeSteppers/acoustic_runge_kutta_3.jl
glwagner: Done in 8cfea6a — the scheme is passed unconditionally and AIVA-vs-vanilla is now dispatch on `AdaptiveImplicitVerticalAdvection` down the chain (per-substep solve, base-tendency entry, and the post-loop demotion all lifted from value guards to methods). Fence and suite re-validated, unchanged.

src/TimeSteppers/acoustic_runge_kutta_3.jl
glwagner: Renamed throughout in 8cfea6a: `implicit_advection_substep!`, `add_implicit_advection_tendency!`, `implicit_advective_base_flux`, and the docstrings now phrase it as the implicit half of the IMEX vertical-advection split.

glwagner:
Production field validation complete. A 72 h, 3 km ERA5-nested hindcast (Dec 2025 PNW atmospheric-river event) ran with adaptive-implicit vertical advection on **every** prognostic — momentum, ρθ via this PR's fold-in, moisture via the #913/#914 fixes — with the wizard's vertical CFL term dropped:

- **Stability**: completed all 72 h at Δt pinned to the 7 s cap (mean 6.88 s, 37,700 iterations), riding 47 m/s convective spikes that force the explicit configuration to Δt ≈ 2 s. Wall clock 14h48m vs 29h09m explicit — **1.97×**.
- **Fidelity vs the explicit baseline** (matched-time, 145 snapshot pairs over 72 h): IVT pattern correlation ≥ 0.986 throughout, final-hour 0.994; amplitude ratios 0.98–1.00; precipitable water correlation 0.998. Same storm, half the time.
- **Engagement**: the implicit remainder touches a mean 0.002% (max 0.04%) of the 2 km plane — a pure outlier guard, exactly the design target.

Also mapped beyond this PR's scope: uncapped Δt equilibrates ≈ 10.6 s (horizontal CFL) but exposes a TimeStepWizard retreat-then-overshoot transient instability near Δt ≈ 9–11 — an adaptive-Δt design question independent of AIVA.

src/TimeSteppers/acoustic_substep_helpers.jl
```diff
@@ -320,6 +330,58 @@ so the explicit/implicit velocity split is consistent: the RK stage-entry predic
...
 """
+# First-order upwind flux of the stage-entry thermodynamic state carried by the implicit
+# half's velocity wⁱ = (1 - s) w, density-weighted like the implicit Center-field
+# coefficients so base + perturbation sum to the full-field operator.
+@inline function implicit_advective_base_flux(i, j, k, grid, scheme, td, W, ρθ, ρᵈ)
+    wⁱ = implicit_vertical_velocityᶜᶜᶠ(i, j, k, grid, scheme, td, W)
+    ρᶠ = ℑzᵃᵃᶠ(i, j, k, grid, ρᵈ)
+    θ⁻ = @inbounds ρθ[i, j, k-1] / ρᵈ[i, j, k-1]
+    θ⁺ = @inbounds ρθ[i, j, k]   / ρᵈ[i, j, k]
+    return Azᶜᶜᶠ(i, j, k, grid) * ρᶠ * (max(wⁱ, zero(wⁱ)) * θ⁻ + min(wⁱ, zero(wⁱ)) * θ⁺)
```
glwagner:
```suggestion
    return Azᶜᶜᶠ(i, j, k, grid) * ρᶠ * (max(wⁱ, 0) * θ⁻ + min(wⁱ, 0) * θ⁺)
```

src/Advection.jl
```diff
@@ -46,4 +50,39 @@ end
...
+# A bounds-preserving WENO whose vertical time discretization is adaptive-implicit.
+const BoundsPreservingAVIDWENO = WENO{<:Any, <:Any, <:Any, <:AdaptiveVerticallyImplicitDiscretization, <:Tuple}
+
+# Lazily s-scaled face velocity: indexing yields wᵉ = s·w, the explicit fraction of the IMEX
+# vertical-advection split, so the bounds-preserving flux functions consume the explicit
+# velocity without duplicating their reconstruction (issue #913).
+struct ExplicitVerticalVelocity{G, S, T, W}
+    grid :: G
+    scheme :: S
+    td :: T
```
glwagner: is this time_discretization?

src/Advection.jl
```diff
@@ -46,4 +50,39 @@ end
...
 
+# A bounds-preserving WENO whose vertical time discretization is adaptive-implicit.
+const BoundsPreservingAVIDWENO = WENO{<:Any, <:Any, <:Any, <:AdaptiveVerticallyImplicitDiscretization, <:Tuple}
+
+# Lazily s-scaled face velocity: indexing yields wᵉ = s·w, the explicit fraction of the IMEX
+# vertical-advection split, so the bounds-preserving flux functions consume the explicit
+# velocity without duplicating their reconstruction (issue #913).
+struct ExplicitVerticalVelocity{G, S, T, W}
+    grid :: G
+    scheme :: S
```
glwagner: is this advection_scheme?

src/Advection.jl
```diff
@@ -46,4 +50,39 @@ end
...
+    td :: T
+    w :: W
```
glwagner: suggest writing vertical_velocity

glwagner: Done in b9f635f.

src/Advection.jl
glwagner: Done in b9f635f.

src/Advection.jl
glwagner: Done in b9f635f.

src/TimeSteppers/acoustic_substep_helpers.jl
glwagner: Done in b9f635f.

glwagner:
Per review, the PR now targets **Oceananigans#main** (post-#5924) via a `[sources]` pin, so the bounds-preserving refactor and this work land compatibly in one validation cycle. When 0.112 tags, the pin drops and compat becomes a plain `0.112` bound — one-line follow-up.

Migration summary: the Δt pairing collapsed into the new single extension point (`adaptive_advection_timestep(::AcousticRungeKutta3, clock)` — upstream now owns the container iteration, FluxForm unwrapping, and composes the adaptive-Δt and limiter refreshes); `AtmosphereModel` gains an `update_advection!` orchestrator pairing each scheme with the specific field its reconstruction acts on; `BoundsPreservingAVIDWENO` tracks the new `BoundsPreservation` type. The #913 s-scaling wrapper is still required — the refactored bounded z-flux reads the raw face velocity — and upstreaming that scaling into the new structure is the planned follow-up PR to Oceananigans.

Ladder re-run in full on the new base: fence 4/4 (uniform-θ drift 4 mK), bounded-twin transport mismatch **0.0%** (tighter than on 0.111), suite **110/110 in Float64 and Float32**, Aqua 33/33, ExplicitImports 5/5.

glwagner:
Two CI-driven fixes pushed, plus a note on which remaining failures are pin artifacts.

**A real bug CI caught that local validation structurally could not.** The `update_advection!` orchestrator reached for `model.formulation.potential_temperature` — correct for the potential-temperature models this work is built around, an immediate crash for `StaticEnergyFormulation`, whose specific field is `specific_energy`. Breeze had no formulation-agnostic accessor for *the specific field an advection operator reconstructs* (as opposed to the density-weighted prognostic named by `thermodynamic_density_name`), so this adds `specific_thermodynamic_field` to the formulation interface with a method per formulation. Both formulations now time-step; the bounds-preserving limiter gets the right field on either.

**Workspace pinning.** CI activates `--project=test`, `docs`, and `benchmarking`, and a workspace sub-project does not inherit the root's `[sources]` — so those workflows were failing to *resolve* (registry 0.111, no `update_advection!`) rather than failing on anything substantive. All three now carry the pin; all of them drop together when 0.112 tags.

**Known artifact while pinned to an unreleased Oceananigans**: `Aqua.test_all` runs `test_persistent_tasks`, which builds a *fresh* temporary environment; that environment resolves Oceananigans from the registry and cannot precompile against the renamed hook. It is not fixable from this PR and clears automatically once 0.112 is released and the pins are removed. (It passed earlier in the migration only because a stale precompile cache masked it.)

Upstream status for anyone following: the 0.112 blocker was a circular compat deadlock — Oceananigans 0.112 requires `ConservativeRegridding = "0.2"`, while every registered CR 0.2.x capped Oceananigans at 0.111. CR v0.2.13 is now registered with `Oceananigans = "0.110, 0.111, 0.112, 0.113"`, so CliMA/Oceananigans#5931 needs only a CI re-run to clear its 13 resolution failures.

Project.toml
```diff
@@ -33,19 +33,22 @@ BreezeCloudMicrophysicsExt = ["CloudMicrophysics", "SpecialFunctions"]
...
+CloudMicrophysics = "0.38"
 Dates = "<0.0.1, 1"
 DocStringExtensions = "0.9.5"
 GPUArraysCore = "0.2"
 InteractiveUtils = "<0.0.1, 1"
 JLD2 = "0.5.13, 0.6"
 KernelAbstractions = "0.9"
 LazyArtifacts = "<0.0.1, 1"
-Oceananigans = "0.111"
+Oceananigans = "0.111, 0.112"
```
giordano: Based on https://github.com/NumericalEarth/Breeze.jl/pull/964#issuecomment-5574371350 sounds like this should be
```suggestion
Oceananigans = "0.112"
```
once that version is out

src/PotentialTemperatureFormulations/potential_temperature_tendency.jl
```diff
@@ -7,6 +7,9 @@ using Breeze.BoundaryConditions: theta_to_energy_bcs, materialize_atmosphere_fie
 
 const PotentialTemperatureModel = AtmosphereModel{<:Any, <:LiquidIcePotentialTemperatureFormulation}
 
+AtmosphereModels.specific_thermodynamic_field(formulation::LiquidIcePotentialTemperatureFormulation) =
+    formulation.potential_temperature
```
glwagner:
```suggestion
AtmosphereModels.specific_thermodynamic_field(formulation::LiquidIcePotentialTemperatureFormulation) = formulation.potential_temperature
```

src/StaticEnergyFormulations/static_energy_tendency.jl
```diff
@@ -13,6 +13,9 @@ const StaticEnergyModel = AtmosphereModel{<:Any, <:StaticEnergyFormulation}
 AtmosphereModels.liquid_ice_potential_temperature(model::StaticEnergyModel) = Diagnostics.LiquidIcePotentialTemperature(model, :specific)
 AtmosphereModels.liquid_ice_potential_temperature_density(model::StaticEnergyModel) = Diagnostics.LiquidIcePotentialTemperature(model, :density)
 AtmosphereModels.static_energy(model::StaticEnergyModel) = model.formulation.specific_energy
+
+AtmosphereModels.specific_thermodynamic_field(formulation::StaticEnergyFormulation) =
+    formulation.specific_energy
```
glwagner:
```suggestion
AtmosphereModels.specific_thermodynamic_field(formulation::StaticEnergyFormulation) = formulation.specific_energy
```

src/AtmosphereModels/formulation_interface.jl
```diff
@@ -86,6 +86,16 @@ Accepts a `Symbol`, `Val(Symbol)`, or formulation struct.
...
+
+Return the *specific* (per unit mass) thermodynamic field the `formulation` evolves — the
+quantity its advection operator reconstructs, as opposed to the density-weighted prognostic
+named by [`thermodynamic_density_name`](@ref). A bounds-preserving scheme limits the
+reconstruction of this field.
+"""
+function specific_thermodynamic_field end
+
 thermodynamic_density_name(formulation::Symbol) =
     thermodynamic_density_name(Val(formulation))
```
glwagner:
```suggestion
thermodynamic_density_name(formulation::Symbol) = thermodynamic_density_name(Val(formulation))
```

PR #968, opened 2026-09-04
Add single-column / column-ensemble mode for AtmosphereModel

### LINKED ISSUE #179: Add SingleColumnModel mode
It will be useful to have a special SingleColumnModel mode (similar to what we have with `Oceananigans.HydrostaticFreeSurfaceModel` that allows us to do simple performant single column simulations without solving for pressure, etc when using a grid with `topology = (Flat, Flat, Bounded)`.

glwagner:
Closes #179.

Adds a **single column mode** for `AtmosphereModel`, analogous to
`Oceananigans.HydrostaticFreeSurfaceModel`'s. On a grid with
`topology = (Flat, Flat, Bounded)` the model runs as a single vertical column;
with `Oceananigans.Grids.ColumnEnsembleSize` it runs as a *forest* of
independent columns advanced concurrently in a single kernel launch — ideal for
ensembles (parameter sweeps, closure calibration, PBL-scheme development).

## What it does

On a `(Flat, Flat, Bounded)` grid, horizontal advection/diffusion/pressure-gradient
terms already vanish (via `Flat`-topology operator dispatch) and no halo information
is exchanged between columns. For anelastic dynamics this PR additionally:

- **omits the pressure solve** — the anelastic constraint `∂z(ρᵣ w) = 0` with rigid
  lids forces `w ≡ 0`, so no solver is built (`dynamics_pressure_solver → nothing`);
- **omits vertical-velocity stepping** — `ρw` stays at zero (`w ≡ 0`); vertical
  transport comes from the turbulence closure and prescribed large-scale forcing.

Horizontal momentum and all scalars still evolve under vertical mixing, Coriolis,
microphysics, radiation, and forcing.

Column independence is verified bit-for-bit: each column of an `N₁×N₂` ensemble
reproduces a standalone single-column run exactly.

## Per-column parameters (for ensembles)

Each column can carry its own parameters, supplied as arrays indexed by column:

- **Closure** — an array of closures (`SingleColumnGrid` dispatch in
  `Breeze.TurbulenceClosures` selects each column's closure with `getclosure`).
- **Coriolis** — an array of rotations (inherited from Oceananigans).
- **Forcing** — a discrete `Forcing` with per-column array `parameters`.
- **Reference state** — array-valued `surface_pressure`/`potential_temperature`
  give each column its own adiabatic background profile.

## Notable changes

- New `const SingleColumnGrid` alias and `single_column.jl`.
- `ReferenceState` builds full `CenterField` profiles on column-ensemble grids
  (reduced `(Nothing, Nothing, Center)` fields cannot be constructed on
  `Flat`-with-size-`>1` grids); ordinary 3D and 1×1 grids keep the reduced profile.
- `anelastic_pressure_solver_analytic.jl` now exercises the anelastic tridiagonal
  solver **directly** rather than through a model, since an `AtmosphereModel` on a
  `(Flat, Flat, Bounded)` grid now runs in single-column mode and omits the solve.

## Docs & tests

- New docs page **Dynamics → Single column mode** and a **Single-column ensemble**
  example (a forest of columns with per-column vertical diffusivity).
- New `test/single_column_mode.jl`: construction/`w ≡ 0`, ensemble independence,
  per-column closure/coriolis/forcing, heterogeneous reference states, and a 3D
  no-regression check.

## Notes

- CPU-verified end to end (full test suite green). The per-column-array GPU paths
  move arrays to the grid architecture via `on_architecture` and are
  correct-by-construction, but were not runtime-tested (no GPU in the dev
  environment).


glwagner:
@ewquon this should be useful for calibrating turbulence closures in a single column context against LES data.

src/AtmosphereModels/atmosphere_model.jl
```diff
@@ -309,7 +313,14 @@ function AtmosphereModel(grid;
     # The closure's scalars — thermodynamic density, moisture, microphysical prognostic fields, user
     # tracers — in the order the vertically-implicit solve indexes them (see `closure_scalar_index`)
     scalar_names = closure_scalar_names(formulation, microphysics, tracer_names)
-    closure = Oceananigans.Utils.with_tracers(scalar_names, closure)
+    # `with_tracers` has no method for a per-column *array* of closures, so map it over the array here
+    # (in Breeze's own code — extending Oceananigans' `with_tracers` on `AbstractArray` would be piracy).
+    closure = closure isa AbstractArray ?
+        map(c -> Oceananigans.Utils.with_tracers(scalar_names, c), closure) :
+        Oceananigans.Utils.with_tracers(scalar_names, closure)
```
glwagner: How does Oceananigans handle this?

src/AtmosphereModels/single_column.jl
glwagner, line 1: let's call this "single_column_mode.jl"

src/AtmosphereModels/single_column.jl
```diff
@@ -0,0 +1,20 @@
...
+are given size greater than one — e.g. via `Oceananigans.Grids.ColumnEnsembleSize`, which forces the
+horizontal halos to zero — the grid holds a horizontally independent "forest" of columns that can be
+advanced concurrently in a single kernel launch, with no coupling between columns.
+
+`AtmosphereModel`s built on a `SingleColumnGrid` run in "single column mode": the anelastic
+dynamics omit the pressure solve and the vertical-velocity stepping (`w ≡ 0`), so vertical transport
+comes from the turbulence closure and prescribed large-scale forcing (e.g. subsidence). This mirrors
+`Oceananigans.HydrostaticFreeSurfaceModel`'s single-column / column-ensemble mode.
+"""
+const SingleColumnGrid = AbstractGrid{<:AbstractFloat, <:Flat, <:Flat, <:Bounded}
```
glwagner: this is defined in Oceananigans right? I don't think its a good idea to restrict the first parameter to AbstractFloat

src/Thermodynamics/reference_states.jl
```diff
@@ -358,6 +358,21 @@ If `profile` is a `Function`, calls `profile(z)`.
 @inline surface_value(θ::Number) = θ
 surface_value(f::Function) = _nargs(f) == 1 ? f(0) : f(0, 0, 0)
 
+# A "column ensemble" grid holds many independent columns in Flat horizontal dimensions of size > 1
+# (e.g. built with `Oceananigans.Grids.ColumnEnsembleSize`). Reduced `(Nothing, Nothing, Center)`
+# fields — the memory-efficient reference profile broadcast across the horizontal — cannot be
+# constructed on such grids (Oceananigans sizes the data by topology but indexes it by the reduced
+# location). On an ordinary 3D grid or a 1×1 single column, reduced fields work as before.
+function is_column_ensemble(grid)
```
glwagner, line 366: is this also defined in Oceananigans?

src/Thermodynamics/reference_states.jl
```diff
@@ -409,9 +424,23 @@ function ReferenceState(grid, constants=ThermodynamicConstants(eltype(grid));
                         ice_mass_fraction = nothing)
 
     FT = eltype(grid)
-    p₀ = convert(FT, surface_pressure)
     pˢᵗ = convert(FT, standard_pressure)
-    loc = (nothing, nothing, Center())
+
+    # Heterogeneous per-column reference state: array-valued `surface_pressure` and/or
+    # `potential_temperature` give each column of a column ensemble its own adiabatic profile.
+    if surface_pressure isa AbstractArray || potential_temperature isa AbstractArray
```
glwagner: predicating on this input seems a bit implicit (I wonder if we can test for SingleColumnGrid + Nx, Ny > 1 instead?)

src/TurbulenceClosures/TurbulenceClosures.jl
```diff
@@ -121,4 +121,36 @@ end
...
+@inline AtmosphereModels.∂ⱼ_𝒯₃ⱼ(i, j, k, grid::SingleColumnGrid, ρ, closures::ClosureArray, args...) =
+    AtmosphereModels.∂ⱼ_𝒯₃ⱼ(i, j, k, grid, ρ, getclosure(i, j, closures), args...)
+
+@inline AtmosphereModels.∇_dot_Jᶜ(i, j, k, grid::SingleColumnGrid, ρ, closures::ClosureArray, args...) =
+    AtmosphereModels.∇_dot_Jᶜ(i, j, k, grid, ρ, getclosure(i, j, closures), args...)
+
+# The constructor's `with_tracers` and each stage's `compute_closure_fields!` (both Oceananigans
+# functions) have no method for a bare array of closures. Rather than pirate them, Breeze handles the
+# array case in its own code: the constructor maps `with_tracers` over the array, and the
+# `compute_closure_fields!` call site skips field-less closure arrays (see AtmosphereModels).
```
glwagner: this comment can probably be deleted. it doesn't attach to any code.

glwagner:
Addressed the review comments in `bc703c7f`. Summary:

**Consolidation.** All single-column *dispatch* now lives in one file, `src/single_column_mode.jl` (a `SingleColumnMode` submodule loaded last): the `SingleColumnGrid` alias, the anelastic pressure/tendency no-ops, `dynamics_pressure_solver → nothing`, the per-column closure dispatch, and the `materialize_closure`/`materialize_coriolis` array hooks. Removed the scattered pieces from `AtmosphereModels`, `AnelasticEquations`, and `TurbulenceClosures`. Made it a submodule (rather than a top-level include) so its imports stay self-contained and out of the `Breeze`-level ExplicitImports surface.

Per-comment:

- **`single_column.jl` → `single_column_mode.jl`**: done (deleted the old file; the alias moved into the new one). `SingleColumnGrid` is now exported from `Breeze`.
- **`SingleColumnGrid` / `AbstractFloat`**: relaxed to `AbstractGrid{<:Any, <:Flat, <:Flat, <:Bounded}`. It *is* in Oceananigans (`single_column_model_mode.jl:22`) but unexported and with the `AbstractFloat` restriction, so I define our own. Note this matches the existing `CelestialMechanics` `SingleColumnGrid`, which already uses `<:Any`.
- **"How does Oceananigans handle `with_tracers` on closure arrays?"**: it doesn't, generically — it defines `with_tracers(tracers, ::ISSDVector) = [with_tracers(tracers, v) for v in vec]` per closure-vector-type (only ISSD/TISSD). So the idiom is "map over the array." I moved that into a Breeze-owned `materialize_closure` hook (generic passthrough in `AtmosphereModels`, array method in `SingleColumnMode`) instead of an inline branch — and it avoids the type piracy that a generic `with_tracers(::AbstractArray)` would trip.
- **`is_column_ensemble` in Oceananigans?**: no — it's ours (Oceananigans has `ColumnEnsembleSize` for construction but no ensemble predicate).
- **reference-state `isa AbstractArray` predicate → dispatch**: replaced with `materialize_reference_state` methods — a homogeneous method (`::Number`, `::Union{Number,Function}`) and per-column methods (`::AbstractArray`). Kept `is_column_ensemble` for the reduced-vs-full field-location choice. Left the reference-state code in `Thermodynamics` since it builds on the reference-state machinery there.
- **Dangling comment (TurbulenceClosures)**: deleted.

Green locally on CPU: `single_column_mode` 37/37, `quality_assurance` (Aqua 34/34 + ExplicitImports 5/5), `reference_states` 219 + 66, `anelastic_pressure_solver_analytic` 2/2. Rebased onto latest `main`; PR is conflict-free.

test/single_column_mode.jl
```diff
@@ -0,0 +1,235 @@
...
+
+        # Each ensemble column must match its standalone counterpart bit-for-bit: the columns
+        # are genuinely independent (no horizontal coupling, no halo exchange between columns).
+        for j in 1:N₂, i in 1:N₁
+            ensemble_column = Array(interior(prognostic_fields(ensemble)[name]))[i, j, :]
+            standalone_column = Array(interior(prognostic_fields(standalone[i, j])[name]))[1, 1, :]
+            @test ensemble_column == standalone_column
+        end
+
+        @test maximum(abs, interior(ensemble.velocities.w)) == 0
```
giordano: Is this checking all elements are zero? If so, I personally find
```suggestion
        @test all(iszero, interior(ensemble.velocities.w))
```
easier to parse

ewquon (approved):
Look forward to exercising this

glwagner:
@giordano are the L4 runners down?

glwagner:
I think we can merge this bc it does not use GPU (?) or at least it is not critical.

giordano:
> @giordano are the L4 runners down?

That's a question for @awiteck

glwagner:
ok I will merge this since the last run passed

awiteck:
Looks like L4s are running now. There was probably no on-demand capacity at the time

PR #974, opened 2026-09-07
Add `ρE` and `ρqᵗ` interface keys, and validate boundary condition and forcing names

### LINKED ISSUE #956: AtmosphereModel silently drops unknown boundary_conditions keys (bit hard across the e→s rename)
`AtmosphereModel(grid; boundary_conditions = (; ρe = ...))` on Breeze 0.10 constructs without complaint: since #926 renamed static energy `e → s`, the energy-BC interface dispatches on `:ρs` (`BoundaryConditions.jl:203`), and a `:ρe`-keyed entry matches nothing, so the model materializes default no-flux BCs as if nothing had been passed.

Downstream this cost us a full round of failed production jobs: NumericalEarth's Breeze extension passed its coupling bottom energy flux and its nested-LAM lateral θ boundary values under `ρe` (correct through 0.9). On 0.10 both were silently dropped. We were lucky the coupling path *crashed* (`Nothing has no field condition` while unwrapping the expected wrapper) — the nested lateral values alone would have produced a physically wrong but running simulation with unforced θ boundaries, which is the worst failure mode there is.

Suggestion: validate `boundary_conditions` keys at construction against the model's prognostic-field names plus the recognized interface aliases (`ρs`/`s`/`StaticEnergy`, …) and `throw(ArgumentError)` on anything unrecognized. A hard error at construction would have turned three dead GPU jobs and a debugging session into a one-line message at submit time — and it makes renames like e→s safe to roll out, since stale keys announce themselves.

(NumericalEarth side is fixed by keying on `pkgversion(Breeze)`; this issue is about making the failure loud for the next consumer.)

glwagner:
Closes #956.

`AtmosphereModel(grid; boundary_conditions = (; ρe = ...))` constructed silently. Since the `e → s` rename in #926 the energy-BC interface dispatches on `:ρs`, so a `:ρe`-keyed entry matched nothing and default no-flux conditions were materialized as if nothing had been passed — which is what cost NumericalEarth three dead GPU jobs.

## Validate names at construction

Every key of `boundary_conditions` and `forcing` must now name something that can carry them — a prognostic field, a velocity component where velocities are prognostic, or an interface key. Anything else raises an `ArgumentError` listing the valid names, instead of being merged in and never looked up.

```julia
julia> AtmosphereModel(grid; boundary_conditions=(; ρe=bcs))
ERROR: ArgumentError: Invalid boundary_conditions: (:ρe,) do not name anything that carries boundary conditions!
Boundary conditions may be set on (:ρu, :ρv, :ρw, :ρθ, :ρqᵛ, :ρE, :ρqᵗ).
```

## Interface keys that do not depend on the model

Two prognostic names are set by choices made elsewhere in the model. The thermodynamic variable follows `formulation` (`ρθ` or `ρs`), and the moisture variable follows `microphysics` (`ρqᵛ` under non-equilibrium cloud formation, `ρqᵉ` under saturation adjustment) — which `BulkMicrophysics` forwards to its `cloud_formation` option, so a choice about *condensation* silently renames the key an *evaporative flux* has to use.

Keying an input by those names therefore ties a setup to one formulation and one scheme. This PR adds two keys that name the physical input instead:

| Key | Input | Specific alias | Applied to | Conversion |
|---|---|---|---|---|
| `ρE` | energy | `E` | thermodynamic prognostic | ÷ `cᵖᵐ` (flux), ÷ `cᵖᵐ Π` (forcing) for `ρθ`; none for `ρs` |
| `ρqᵗ` | water | `qᵗ` | moisture prognostic | none — a pure re-key |

Neither is itself prognostic, so both are unambiguous, and the same setup now runs under any formulation and any microphysics:

```julia
boundary_conditions = (; ρE = FieldBoundaryConditions(bottom=FluxBoundaryCondition(Q)),
                         ρqᵗ = FieldBoundaryConditions(bottom=FluxBoundaryCondition(E)))

θ_model = AtmosphereModel(grid; boundary_conditions)
s_model = AtmosphereModel(grid; formulation=:StaticEnergy, boundary_conditions)
```

The specific names stay valid where they are genuinely prognostic, which is what you want when you mean a flux of that variable rather than an energy or water input. Using one where it is not prognostic gets a message pointing at the interface key, and supplying both an interface key and its target is an error — the two would be summed into one flux on the same field. A new formulation that has not said how an energy flux enters its prognostic variable gets a clear error rather than a silent wrong answer.

## Documentation

Boundary conditions and surface fluxes had no manual coverage outside the examples, so this adds **AtmosphereModel documentation → Boundary conditions and forcing**: valid keys and the error for an unrecognized one, why the two prognostic names depend on the model, the interface keys and their conversions, surface fluxes from bulk formulae, and density-versus-specific forcing keys. The existing `AtmosphereModel` section is retitled *AtmosphereModel documentation*.

`notation.md` gains rows for `E`, `ρE`, and `ρqᵗ`; `AGENTS.md` and `.agents/physics-debugging.md` are updated. Examples, `docs/src/index.md`, and the DCMIP2016 TC validation script migrate to `ρE`/`ρqᵗ`.

## A live instance of the same bug, caught by the new validation

`examples/radiative_convection.jl` keyed its surface vapor flux `ρqᵗ` while the `SaturationAdjustment` moisture prognostic is `ρqᵉ`. Both that boundary condition and its `scalar_advection` entry were silently dropped, so the example has been running without its surface moisture flux. Fixed — and `ρqᵗ` is now the correct key it was reaching for.

## Follow-up, deliberately not in this PR

`scalar_advection` has the same hole: `AtmosphereModel(grid; scalar_advection=(; ρbogus=WENO()))` constructs and drops the entry. The validator is a couple of lines from covering it.

## Testing

`forcing_and_boundary_conditions`, `specific_forcing`, `atmosphere_model_construction`, and `boundary_conditions` pass (537/537), including four new testsets covering name rejection, `ρE` → thermodynamic routing, `ρqᵗ` → moisture routing, and the duplicate-key errors. Verified numerically that a `ρE` forcing gives `Δρs = F Δt` on a static-energy model and `Δρθ = F Δt / (cᵖᵐ Π)` on a potential-temperature model, that `ρqᵗ` gives `Δρqᵛᵉ = F Δt` under both `nothing` and `SaturationAdjustment`, and that the specific aliases pick up the reference density. Every `@example` block on the new docs page was executed, and a Documenter build of the page reports no unresolved references.

https://claude.ai/code/session_01SNUCtZ1LaKGL4rZbMNb2pa


glwagner:
Downstream verification from the consumer that motivated #956 — this works, and `ρE` is exactly the right shape.

NumericalEarth's Breeze extension has an unusual usage worth recording, since it's the case a formulation-agnostic key could plausibly have broken: for a nested LAM it merges **lateral Value BCs** (parent θ boundary values) and a **bottom energy Flux BC** (coupler surface enthalpy flux) onto the *same field* through one key. Checked out `glw/energy-key-validation` and ran that exact pattern on a `LiquidIcePotentialTemperature` model:

- bottom: `EnergyFluxBoundaryConditionFunction` ✓ (wrapped for the `cᵖᵐ` conversion)
- west/east: `Value` classification, values **300.0 preserved** ✓ (passed through unconverted — not divided by `cᵖᵐ`, which would have been silently wrong by ×10³)
- `ρe` key: **rejected** with the new `ArgumentError` ✓ — precisely the failure mode that cost us three GPU jobs

The `energy_to_theta_bc(bc) = bc` fallback is what makes the mixed-classification case work; worth keeping in mind if that dispatch is ever tightened, since "energy key carrying Value BCs of the prognostic variable" is a slightly odd thing to rely on. (The cleaner long-run shape for us is lateral values keyed as `ρθ` alongside an `ρE` bottom flux, which `validate_energy_bcs` currently forbids as double-specification — reasonable as-is; flagging only so the constraint is a known one.)

For NumericalEarth: NumericalEarth#668 currently version-gates `ρe`/`ρs`. Once this lands I'll retarget it to `ρE`, which ends the gating permanently — that's the real value of a formulation-agnostic key, thank you for adding it.

.agents/physics-debugging.md
```diff
@@ -6,12 +6,28 @@
...
+An energy input (W/m² at a boundary, W/m³ in the interior) goes under `ρE` — or the specific
+key `E` for forcings — and Breeze converts it for the prognostic variable: divided by `cᵖᵐ`
+(fluxes) or `cᵖᵐ Π` (forcings) for `ρθ`, unconverted for `ρs`.
+
+A water input (kg/m²/s at a boundary, kg/m³/s in the interior) goes under `ρqᵗ` — or the
+specific key `qᵗ` — and is applied to the prognostic moisture unconverted, since water added
+there is water added to `qᵗ` under every scheme.
+
+The specific names are keys only where they are actually prognostic: `ρs`/`s` when static
+energy is the thermodynamic variable, `ρqᵛ` or `ρqᵉ` depending on the microphysics. Supplying
```
ewquon:
```suggestion
Each variable's own name is a key only where that variable is prognostic: `ρs`/`s` when static
energy is the thermodynamic variable, `ρqᵛ` or `ρqᵉ` depending on the microphysics. Supplying
```

src/AtmosphereModels/formulation_interface.jl
```diff
@@ -89,6 +89,25 @@ function thermodynamic_density_name end
...
+    total_energy_density_name
+
+The key under which an *energy* input — a surface heat flux, a diabatic heating rate — is
+supplied to `AtmosphereModel`, `:ρE`, along with its specific alias `:E` for forcings.
+
+``E`` denotes total energy (`total_energy`), so the key names the physical input without
+committing to the variable that carries it: a `boundary_conditions` or `forcing` entry keyed
+`ρE` is routed onto whichever thermodynamic density the formulation evolves (see
+[`thermodynamic_density_name`](@ref)), converted as that variable requires — divided by
+``cᵖᵐ`` (fluxes) or ``cᵖᵐ Π`` (forcings) for ``ρθ``, and passed through unconverted for
```
ewquon, line 110: `Π` inconsistency https://github.com/NumericalEarth/Breeze.jl/issues/976

src/AtmosphereModels/atmosphere_model.jl
```diff
@@ -486,6 +487,92 @@ function field_names(dynamics, formulation, microphysics, tracer_names)
...
+    if name ∈ (:ρs, :s)
+        return string('\n', "An energy flux or forcing is supplied under ", total_energy_density_name,
+                      " (or E) and applied to the prognostic thermodynamic variable; ", name,
+                      " is a key only when static energy is prognostic (formulation = :StaticEnergy).")
+    elseif name ∈ (:ρqᵛ, :qᵛ, :ρqᵉ, :qᵉ)
+        return string('\n', "A water flux or forcing is supplied under ", total_moisture_density_name,
+                      " (or qᵗ) and applied to the prognostic moisture variable; ", name,
+                      " is a key only under the microphysics that makes it prognostic, which for ",
+                      "BulkMicrophysics is set by `cloud_formation`.")
+    end
```
ewquon, line 511:
```suggestion
    elseif name ∈ (:ρe, :e)
        return string('\n', "An energy flux or forcing is supplied under ", total_energy_density_name,
                      " (or E); ", name, " names turbulent kinetic energy, so it is a key only under ",
                      "a prognostic-TKE closure.")
    end
```

src/StaticEnergyFormulations/static_energy_tendency.jl
```diff
@@ -68,6 +70,8 @@ end
              - buoyancy_flux
              - ∇_dot_Jᶜ(i, j, k, grid, ρ_field, closure, closure_fields, id, specific_energy, clock, model_fields, closure_buoyancy)
              + ρs_forcing(i, j, k, grid, clock, model_fields)
+             # An energy forcing needs no conversion: static energy *is* an energy per unit mass
+             + ρE_forcing(i, j, k, grid, clock, model_fields)
```
ewquon, line 77:
```suggestion
             # An energy forcing (note: ρs and ρE are mutually exclusive) is an energy per unit mass
             # and needs no conversion
             + ρs_forcing(i, j, k, grid, clock, model_fields)
             + ρE_forcing(i, j, k, grid, clock, model_fields)
```

ewquon:
@glwagner In addition to fixing the docs build and tests, I've made a few changes to the logic to how
atmosphere keys are filtered and merged.

- Under `formulation = :StaticEnergy`, guard against `(; ρs, ρE)` so that `ρs` forcing isn’t double counted.
- Allow at most one source per weighting (density-weighted vs specific) for a given *quantity*.
- As a rule, `ρqᵗ` → `ρqᵛᵉ` so specifying `(; ρqᵗ, ρqᵛᵉ)` == `(; ρqᵛᵉ, ρqᵛᵉ)` is *not* allowed. Pairing it across weightings now *is* allowed: e.g., on the default scheme, `(; ρqᵗ, qᵛ)` and `(; qᵗ, ρqᵛ)` are legal, which reverses two assertions that expect them to throw.
- The energy interface key is guarded depending on the formulation. For `:StaticEnergy`, `ρE` → `ρs` and no conversion is applied. For `:LiquidIcePotentialTemperature`, `ρE` → `ρθ` and is divided by `cᵖᵐ Π`, so an energy source and a θ source are different inputs and both are allowed (e.g., `(; θ, E)` in BOMEX).
- Validate BCs on each side separately. E.g., enables lateral Dirichlet value of the prognostic thermodynamic variable with a surface energy flux.

The macOS test is timing out, and shouldn’t be a result of this PR.

glwagner:
~~docs is still failing?~~ nevermind!

PR #529, opened 2026-02-27
Implement checkpointing for `AtmosphereModel`

ewquon:
Addresses https://github.com/NumericalEarth/Breeze.jl/issues/443

Following https://github.com/CliMA/Oceananigans.jl/blob/51c67b5dc9445fe3b9015b3ccad4ebdfb89258f1/docs/src/simulations/checkpointing.md, I've independently verified that Oceananigans restarts work flawlessly -- bitwise agreement.

The same workflow applied to the free convection demo (https://numericalearth.github.io/BreezeDocumentation/stable/#Quick-Start)
```julia
simulation = Simulation(model, Δt=10, stop_time=2hours)
simulation.output_writers[:checkpointer] = Checkpointer(model, schedule=IterationInterval(100))
conjure_time_step_wizard!(simulation, cfl=0.7)
run!(simulation, checkpoint_at_end=true)
```
--vs--
```julia
simulation = Simulation(model, Δt=10, stop_iteration=2000)
simulation.output_writers[:checkpointer] = Checkpointer(model, schedule=IterationInterval(100))
conjure_time_step_wizard!(simulation, cfl=0.7)
run!(simulation)
# ...
simulation = Simulation(model, Δt=10, stop_time=2hours)
simulation.output_writers[:checkpointer] = Checkpointer(model, schedule=IterationInterval(100))
conjure_time_step_wizard!(simulation, cfl=0.7)
run!(simulation, pickup="checkpoint_iteration2000.jld2", checkpoint_at_end=true)
```
gives qualitatively identical results. The fields differ by:
```
ρu.data DIFFER!? max abs/rel diff : 0.0002235832290651274 0.027438062136977973
ρw.data DIFFER!? max abs/rel diff : 0.0001403733258555917 0.03648357117818405
ρθ.data are approximately equal
ρqᵗ.data DIFFER!? max abs/rel diff : 6.206636658928621e-7 4.801171466458838e-5
```

PR #978, opened 2026-09-09
Implement checkpointing for `AtmosphereModel`

### LINKED ISSUE #443: Checkpointing for AtmosphereModel
We need it, obviously!

Checkpointing was refactored in Oceananigans recently and this should be relatively straightforward.

ewquon:
Closes #443. Supersedes #529.

`AtmosphereModel` now implements Oceananigans' `prognostic_state` /
`restore_prognostic_state!` protocol, so the stock `Checkpointer` and
`run!(simulation; pickup=...)` work with no Breeze-specific machinery. Neither
method existed before, so `Checkpointer` fell through Oceananigans' untyped
fallback and pickup threw.

Restarts are bitwise, on CPU and on GPU. Getting there also required pinning the
null-space component of the anelastic pressure solve, whose underlying cause is
upstream and is reported in CliMA/Oceananigans.jl#5960. That one line is a
workaround, not a requirement: with it reverted and the fix proposed in #5960
applied, the tests below still pass on CPU and on a T4, so it can be dropped
once upstream lands.

`test/checkpointing.jl` requires exact equality between a continuous run and a
checkpoint/restart run, across both dynamics (compressible in explicit and
split-explicit form), equilibrium and non-equilibrium microphysics at two and
five prognostic species, and the closure-free, diagnostic-closure and
prognostic-closure cases. The checkpointed field set changes with each of those
knobs. 92 assertions, all exact rather than approximate — an approximate
comparison passes on a contaminated restart.

---

### DIFF
--- a/src/AnelasticEquations/anelastic_pressure_solver.jl
+++ b/src/AnelasticEquations/anelastic_pressure_solver.jl
@@ -83,7 +83,7 @@ end
 
 function solve_for_anelastic_pressure!(pₙ, solver, ρŨ, Δt)
     compute_anelastic_source_term!(solver, ρŨ, Δt)
-    fill!(solver.storage, 0)
+    fill!(solver.storage, 0)  # workaround for bitwise reproducibility — see Oceananigans #5960
     solve!(pₙ, solver)
     return pₙ
 end
--- a/src/AnelasticEquations/anelastic_pressure_solver.jl
+++ b/src/AnelasticEquations/anelastic_pressure_solver.jl
@@ -83,6 +83,7 @@ end
 
 function solve_for_anelastic_pressure!(pₙ, solver, ρŨ, Δt)
     compute_anelastic_source_term!(solver, ρŨ, Δt)
+    fill!(solver.storage, 0)
     solve!(pₙ, solver)
     return pₙ
 end
--- a/src/TurbulenceClosures/tke_based_turbulence_closure.jl
+++ b/src/TurbulenceClosures/tke_based_turbulence_closure.jl
@@ -538,3 +538,6 @@ function Base.show(io::IO, closure::TKEBasedTurbulenceClosure)
               "├── minimum_tke: ", prettysummary(closure.minimum_tke), '\n',
               "└── negative_tke_damping_time_scale: ", prettysummary(closure.negative_tke_damping_time_scale))
 end
+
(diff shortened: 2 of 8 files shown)


src/AnelasticEquations/anelastic_pressure_solver.jl
```diff
@@ -83,6 +83,7 @@ end
 
 function solve_for_anelastic_pressure!(pₙ, solver, ρŨ, Δt)
     compute_anelastic_source_term!(solver, ρŨ, Δt)
+    fill!(solver.storage, 0)
```
glwagner: i think this is a relatively large cost to pay to fix checkpointing. do we need to just zero out the solver storage when restoring from a checkpoint?

src/TurbulenceClosures/tke_based_turbulence_closure.jl
```diff
@@ -538,3 +538,6 @@ function Base.show(io::IO, closure::TKEBasedTurbulenceClosure)
               "├── minimum_tke: ", prettysummary(closure.minimum_tke), '\n',
               "└── negative_tke_damping_time_scale: ", prettysummary(closure.negative_tke_damping_time_scale))
 end
+
+Oceananigans.prognostic_state(::TKEClosureFields) = nothing
```
glwagner, line 544: perhaps add a note that this may have to change in the future

glwagner (approved):
seems legit

src/AnelasticEquations/anelastic_pressure_solver.jl
ewquon: This is the bitwise reproducibility fix until https://github.com/CliMA/Oceananigans.jl/issues/5960 lands.

ewquon:
```suggestion
    fill!(solver.storage, 0)  # workaround for bitwise reproducibility — see Oceananigans #5960
```

src/TurbulenceClosures/tke_based_turbulence_closure.jl
ewquon, line 544:
```suggestion
# Kᵘ, Kᶜ, Kᵉ and Lᵉ are recomputed exactly by update_state! from the prognostic state alone
# but this may change in the future if the closure becomes dependent on the previous state.
Oceananigans.prognostic_state(::TKEClosureFields) = nothing
```

PR #979, opened 2026-09-09
Add the Supersaturation diagnostic

glwagner:
`Supersaturation(model)` returns `RelativeHumidity(model) - 1` as an operation, and `SupersaturationField(model)` the corresponding `Field`.

Supersaturation is what decides whether a droplet grows, activates, or evaporates, and in a cloud it is a small residual of two much larger numbers, so it is the field one wants to write out rather than the humidity itself. Cloud-chamber and activation work reaches for it constantly; this saves every such simulation from spelling out the subtraction.

Tests cover both branches of the diagnostic: subsaturated air, where it is negative everywhere and exactly one less than the relative humidity, and saturated air, where it vanishes wherever saturation adjustment has produced condensate. The docstring carries a doctest, and the notation table records the constructor next to the existing `supersaturation` function.

https://claude.ai/code/session_01CQKnHK63dDbpppr1ixUm6Y

### DIFF
--- a/test/diagnostics.jl
+++ b/test/diagnostics.jl
@@ -184,164 +184,22 @@ end
     grid = RectilinearGrid(default_arch; size=(2, 2, 8), extent=(100, 100, 1000))
     model = AtmosphereModel(grid; microphysics=SaturationAdjustment())
 
-    # Subsaturated: 𝒮 = ℋ - 1 < 0 everywhere, and exactly one less than the humidity
+    # Subsaturated air: the supersaturation is negative everywhere, and is exactly one less than
+    # the relative humidity
     set!(model, θ=300, qᵗ=0.005)
     𝒮 = Supersaturation(model)
     @test 𝒮 isa Oceananigans.AbstractOperations.AbstractOperation
     𝒮_field = SupersaturationField(model)
     ℋ_field = RelativeHumidityField(model)
-    @test all(isfinite.(interior(𝒮_field)))
-    @test all(interior(𝒮_field) .< 0)
-    @test interior(𝒮_field) ≈ interior(ℋ_field) .- 1
+    @test all(isfinite, 𝒮_field)
+    @test all(<(0), 𝒮_field)
+    @test maximum(abs, 𝒮_field - ℋ_field + 1) < eps(FT)
 
-    # Saturated: the supersaturation vanishes wherever saturation adjustment has made condensate
+    # Moist enough to condense: saturation adjustment pins the supersaturation to zero wherever it
+    # makes condensate, and leaves the rest of the column subsaturated
     set!(model, θ=300, qᵗ=0.03)
     𝒮_saturated = SupersaturationField(model)
-    qˡ = model.microphysical_fields.qˡ
-    @allowscalar begin
-        for k in 1:8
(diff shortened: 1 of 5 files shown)


test/diagnostics.jl
```diff
@@ -179,6 +179,34 @@ end
...
+    model = AtmosphereModel(grid; microphysics=SaturationAdjustment())
+
+    # Subsaturated: 𝒮 = ℋ - 1 < 0 everywhere, and exactly one less than the humidity
+    set!(model, θ=300, qᵗ=0.005)
+    𝒮 = Supersaturation(model)
+    @test 𝒮 isa Oceananigans.AbstractOperations.AbstractOperation
+    𝒮_field = SupersaturationField(model)
+    ℋ_field = RelativeHumidityField(model)
+    @test all(isfinite.(interior(𝒮_field)))
+    @test all(interior(𝒮_field) .< 0)
```
giordano: Hoping this works on GPU, the functional form is faster because (1) it doesn't have to allocate the broadcast arrays and (2) it shortcircuits at the first false element  
```suggestion
    @test all(isfinite, interior(𝒮_field))
    @test all(<(0), interior(𝒮_field))
```

glwagner: Ah yes that is better. Also, I think we should try to avoid the `interior` shortcut.

glwagner: Taken, thank you — and it does work on GPU: I ran the diagnostics suite on an A100 and all 61 tests pass. The functional forms are in as `all(isfinite, 𝒮_field)` and `all(<(0), 𝒮_field)`.

glwagner: Done in e70fdb9: no `interior` anywhere in the testset now. The equality against the humidity is `maximum(abs, 𝒮_field - ℋ_field + 1) < eps(FT)`, and the saturated case is judged by reductions rather than a scalar loop over the column — `maximum(qˡ) > 0` that saturation adjustment made condensate, `abs(maximum(𝒮_saturated)) < 1e-3` that it pinned the supersaturation to zero there, and `minimum(𝒮_saturated) < 0` that the rest of the column stayed subsaturated. That also drops the `@allowscalar` block.

glwagner:
Heads-up on something this PR turned up, since it bears on the functional-reduction suggestion above.

`all(isfinite, field)` and `all(<(0), field)` fail on Julia 1.13:

```
MethodError: no method matching initarray!(::SubArray{Bool, 3, Array{Bool, 3}, ...}, ::typeof(identity), ...)
```

The reduction builds a `Bool` reduced field, and Oceananigans' `initarray!` has no method for one. It passes on 1.11 and 1.12, including on the GPU runners, so it only showed up once CI moved the Linux jobs to 1.13 (#988). The failing worker also interrupted the other 59 testsets in that job, which is why the run looked far worse than it was.

The testset now reduces over floats instead, which asserts the same things: `maximum(abs, 𝒮_field) < Inf` for finiteness, since a NaN anywhere fails it, and `maximum(𝒮_field) < 0` for subsaturation. Still no `interior`, still no scalar indexing.

Worth an Oceananigans issue on its own: boolean reductions over fields are a natural thing to write, and they are silently version-dependent.

PR #985, opened 2026-09-10
Weight compressible momentum forces by the dry-air mass fraction

kaiyuan-cheng:
## Summary

The compressible core carries dry-coupled momentum (`ρu = ρᵈ u`: `set_velocity!` writes `ρᵈ u`, `_compute_velocities!` recovers `u = ρu / ℑ(ρᵈ)`, continuity is `∂t ρᵈ = -∇·(ρu)`) but applied the pressure gradient and gravity to it unweighted. Those are force densities on the **total** mass `ρ = ρᵈ + Σρˣ`, so moist air was accelerated by `1/(1 - qᵗ)` too much. On a uniform column with `reference_state = nothing`:

```
qᵗ = 0.02:  dw/dt = -10.010204081632654    (should be -9.81; 2.04% too strong)
qᵗ = 0.0 :  dw/dt = -9.81                  (dry control, unaffected)
```

Dry air hides the bug, which is why the existing suites did not catch it.

## Fix

Weighting the mixture momentum equation by `ρᵈ/ρ` and folding in dry continuity gives

```math
∂_t(ρ^d \boldsymbol{u}) + ∇·(ρ^d \boldsymbol{u}\boldsymbol{u}) = - q^d ∇p - ρ^d g \hat{\boldsymbol{z}} ,
\qquad q^d = ρ^d/ρ = 1 - q^t
```

the height-coordinate form of the `μᵈ α` and `-g μᵈ` factors in [Skamarock & Klemp (2008)](https://doi.org/10.1016/j.jcp.2007.01.037).

New `coupling_mass_fractionᶠᶜᶜ/ᶜᶠᶜ/ᶜᶜᶠ` on the dynamics interface, defaulting to `1` (exact wherever `total_density === dynamics_density`, so anelastic and kinematic paths are untouched). `CompressibleDynamics` overrides it as a ratio of *face-interpolated* densities, so the discrete gravitational force is exactly `-g ℑᶻ(ρᵈ)` — the same face density that recovers `w`. Applied to the pressure gradient and buoyancy on:

- **explicit** — `x/y/z_momentum_tendency`
- **split-explicit** — slow vertical assembly, horizontal step, vertical RHS, and the three tridiagonal coefficients (matrix and RHS share the stage-frozen weight)
- **terrain** — contravariant projection, with horizontal forces weighted at their own x/y faces before interpolation so `ρw̃` stays in step with the `(ρu)′` update

Damping and the sponge take no factor: they act on `(ρw)′`, not on mixture mass. Zero total density falls back to weight `1` via `safe_divide`. Dry runs are bit-identical.

## Tests

New `test/moist_momentum_coupling.jl` (149 assertions): free fall under both time discretizations, horizontal acceleration `-∂ₓp/ℑˣ(ρ)`, the weight values, and a moist split-explicit bubble run.

All passing:

```
moist_momentum_coupling (new)         149     terrain_following_split_explicit  512
acoustic_substepping_components       117     acoustic_substepping_stability     35
acoustic_substepping_open_boundaries   35     adiabatic_balance_set              57
balance_adiabatically                  40     dynamics                           80
substepper_structural                  22     substepper_rest_state              18
quality_assurance (Aqua + ExplicitImports + Core.Box)                            40
```

Two existing tridiagonal-coefficient tests needed the new argument threaded through.

## Docs

`compressible_dynamics.md` described the prognostic mass as the total density, which is not what the code advances. Corrected there, in the `CompressibleEquations` module docstring, in the linearized perturbation equations, and in `dycore_equations_algorithms.md` and `notation.md`.


docs/src/compressible_dynamics.md
```diff
@@ -1,34 +1,53 @@
...
+&\text{Momentum:} && ∂_t(ρ^d \boldsymbol{u}) + ∇·(ρ^d \boldsymbol{u} \boldsymbol{u}) = - q^d ∇ p - ρ^d g \hat{\boldsymbol{z}} + ρ^d \boldsymbol{f} + ∇·\boldsymbol{\mathcal{T}} ,\\
 &\text{Thermodynamic:} && ∂_t χ + ∇·(χ \boldsymbol{u}) = Π \, ∇·\boldsymbol{u} + S_χ ,\\
 &\text{Moisture:} && ∂_t(ρ q^t) + ∇·(ρ q^t \boldsymbol{u}) = S_q .
 \end{aligned}
 ```
 
+Dry air is the prognostic mass because its continuity equation has no sedimentation or phase-change
+source; the momentum carrier is therefore also dry, and velocity is recovered as
+``\boldsymbol{u} = (ρ^d \boldsymbol{u})/ρ^d``. Pressure gradient and gravity, however, act on the
+*total* mass. Weighting the single-velocity mixture equation
```
glwagner: I wonder if this language is a bit misleading/confusing. I agree that gravity "acts" on the total mass, but the more important point here is that the momentum balance incorporates the total gravitational force `rho g` where `rho` is the total density. In other words the central question is not "how gravity works" but rather "how our momentum equation is formulated".

docs/src/compressible_dynamics.md
```diff
@@ -1,34 +1,53 @@
...
+&\text{Dry mass:} && ∂_t ρ^d + ∇·(ρ^d \boldsymbol{u}) = 0 ,\\
+&\text{Momentum:} && ∂_t(ρ^d \boldsymbol{u}) + ∇·(ρ^d \boldsymbol{u} \boldsymbol{u}) = - q^d ∇ p - ρ^d g \hat{\boldsymbol{z}} + ρ^d \boldsymbol{f} + ∇·\boldsymbol{\mathcal{T}} ,\\
 &\text{Thermodynamic:} && ∂_t χ + ∇·(χ \boldsymbol{u}) = Π \, ∇·\boldsymbol{u} + S_χ ,\\
 &\text{Moisture:} && ∂_t(ρ q^t) + ∇·(ρ q^t \boldsymbol{u}) = S_q .
 \end{aligned}
 ```
 
+Dry air is the prognostic mass because its continuity equation has no sedimentation or phase-change
+source; the momentum carrier is therefore also dry, and velocity is recovered as
+``\boldsymbol{u} = (ρ^d \boldsymbol{u})/ρ^d``. Pressure gradient and gravity, however, act on the
```
glwagner: should we note that `u` is also the velocity of dry air (an approximation) because of the form of the continuity equation that we use? `u` is not the barycentric velocity.

src/AtmosphereModels/dynamics_interface.jl
```diff
@@ -177,6 +177,24 @@ overrides it with a diagnosed total-density field, distinct from the coupling de
 """
 total_density(dynamics) = dynamics_density(dynamics)
 
+"""
+    coupling_mass_fractionᶠᶜᶜ(i, j, k, grid, dynamics)
+    coupling_mass_fractionᶜᶠᶜ(i, j, k, grid, dynamics)
+    coupling_mass_fractionᶜᶜᶠ(i, j, k, grid, dynamics)
```
glwagner: I was confused at first seeing three docstrings combined into one... I suggest writing just one function signature and then leaving a remark that the other two functions are the same. Also shouldn't we use DOCSTRINGSEXTENSIONS?

I'm going to make the same comment as previously that it isn't exactly clear to me why we call this a "coupling" term. I think we used a term "carrier" elsewhere, right?

src/AtmosphereModels/dynamics_interface.jl
```diff
@@ -478,6 +496,16 @@ Adapt.adapt_structure(to, s::HorizontalSlowMode) = HorizontalSlowMode(adapt(to,
...
 
+# Whichever forces the wrappers let through still act on the wrapped formulation's momentum.
+const SlowMomentumMode = Union{SlowTendencyMode, HorizontalSlowMode}
+
+@inline coupling_mass_fractionᶠᶜᶜ(i, j, k, grid, s::SlowMomentumMode) =
+    coupling_mass_fractionᶠᶜᶜ(i, j, k, grid, s.dynamics)
+@inline coupling_mass_fractionᶜᶠᶜ(i, j, k, grid, s::SlowMomentumMode) =
+    coupling_mass_fractionᶜᶠᶜ(i, j, k, grid, s.dynamics)
+@inline coupling_mass_fractionᶜᶜᶠ(i, j, k, grid, s::SlowMomentumMode) =
+    coupling_mass_fractionᶜᶜᶠ(i, j, k, grid, s.dynamics)
```
glwagner: consider making these single line

src/AtmosphereModels/dynamics_interface.jl
kaiyuan-cheng: We ended up using "dynamics_", so I renamed it to `dynamics_mass_fraction` here.

docs/src/compressible_dynamics.md
kaiyuan-cheng: Revised this paragraph.

docs/src/appendix/notation.md
```diff
@@ -35,14 +35,14 @@ The following table also uses a few conventions that suffuse the source code and
...
-* Because momentum is coupling-weighted, the *mixture* forces — the pressure gradient and gravity, which act on the total mass ``ρ`` — enter the momentum tendency multiplied by ``qᵈ = ρᵈ/ρ`` (`coupling_mass_fractionᶠᶜᶜ` and its ``ᶜᶠᶜ``/``ᶜᶜᶠ`` counterparts, identically 1 on the anelastic core). Advection, Coriolis and the stress divergence take no factor.
+* The mixture momentum equation includes the full pressure-gradient force and the total gravitational force ``ρ g``. Because the prognostic momentum is coupling-weighted (``ρu = ρᵈ u``), these forces are multiplied by ``qᵈ = ρᵈ/ρ`` in the tendency, using `dynamics_mass_fractionᶠᶜᶜ` and its ``ᶜᶠᶜ``/``ᶜᶜᶠ`` counterparts; this factor is identically 1 on the anelastic core. Advection, Coriolis and stress-divergence terms are not multiplied by ``qᵈ``.
```
glwagner, line 38: If `ρᵈ u` is "coupling weighted momentum" then what is `ρ u` ? I would have thought `ρᵈ u` is the dry air momentum. 🤔

kaiyuan-cheng: `AM.momentum` holds `ρᵈu` on the compressible core and `ρᵣu` on the anelastic core; "coupling-weighted" was only meant as the umbrella for those two. I've revised it a bit and hopefully it is clearer now.

PR #986, opened 2026-09-10
Fill the halos the terrain acoustic substep reads horizontally

kaiyuan-cheng:
Two fields in the terrain acoustic path are written over interior cells and then read at `i±1`, so their halos held stale values: zeros on a stage's first substep, and always wrong at an internal partition boundary of a distributed grid.

- **Predictor `(ρθ)′★`** — `_build_vertical_rhs!` reads it through `∇ᶻp′`, whose terrain method adds the slope correction `slopeₓ·∂ₓ(Cᴸ(ρθ)′★) + slopeᵧ·∂ᵧ(Cᴸ(ρθ)′★)`. Both writers cover the interior only, so the exchange goes after `implicit_advection_substep!`. It is unconditional, matching every other fill in the loop: on a height-coordinate grid `∇ᶻp′` is a pure vertical difference, so nothing reads the filled halos and results stay bit-identical.
- **Tendencies `Gⁿρu`, `Gⁿρv`** — `_assemble_terrain_slow_vertical_momentum_tendency!` interpolates them to `(Center, Center, Face)`, reading face `Nx+1`/`Ny+1`, which the tendency kernels never write. Filled at this consumer, not the producer: `Gⁿρu` has three writers per RK3 stage and this is the last point before the substep loop.

## Test

Terrain and initial condition exactly invariant under a shift of `Nx/2` cells, so the discrete solution must be too (`mod(x, Lx/2)` on power-of-two centers makes the shifted arguments bitwise equal).

| | predictor west halo vs. interior wrap | relative shift asymmetry in `w` |
|---|---|---|
| before | `0.0` vs `1.654e-2` | `3.7e-4` |
| after | equal | `0.0` (bitwise) |

All four assertions fail on `main`. Verified: `terrain_following_split_explicit.jl` 516/516, Aqua 34/34, ExplicitImports 5/5, plus `terrain_following_latlon.jl` and `acoustic_substepping_components.jl`.

### DIFF
--- a/src/CompressibleEquations/acoustic_substepping.jl
+++ b/src/CompressibleEquations/acoustic_substepping.jl
@@ -1520,6 +1520,10 @@ function acoustic_rk3_substep_loop!(model::AtmosphereModel, substepper, Δt, β_
         # solve and recovery substitution see a transport-consistent ρθ′★ (issue #897).
         implicit_advection_substep!(model, substepper, advection, Δτ)
 
+        # On terrain `∇ᶻp′`'s slope correction reads ρθ′★ at i±1, j±1, and both writers above
+        # cover interior cells only. Must follow the implicit substep, which rewrites ρθ′★.
+        fill_halo_regions!(substepper.density_potential_temperature_predictor)
+
         launch!(arch, grid, KernelParameters(1:size(grid, 1), 1:size(grid, 2), 1:size(grid, 3) + 1),
                 _build_vertical_rhs!,
                 substepper.vertical_solver_source_term,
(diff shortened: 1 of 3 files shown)


src/CompressibleEquations/acoustic_substepping.jl
```diff
@@ -1520,6 +1520,10 @@ function acoustic_rk3_substep_loop!(model::AtmosphereModel, substepper, Δt, β_
         # solve and recovery substitution see a transport-consistent ρθ′★ (issue #897).
         implicit_advection_substep!(model, substepper, advection, Δτ)
 
+        # On terrain `∇ᶻp′`'s slope correction reads ρθ′★ at i±1, j±1, and both writers above
+        # cover interior cells only. Must follow the implicit substep, which rewrites ρθ′★.
+        fill_halo_regions!(substepper.density_potential_temperature_predictor)
```
glwagner, line 1525: Is there an algorithm that would avoid this? This will have a severe cost for distributed architectures. Ultimately we might hope to eliminate halo filling completely from the substepping.

PR #969, opened 2026-09-05
Standardize P3 parameter and sedimentation notation

glwagner:
## Summary

- reserve blackboard C for calibratable empirical coefficients, written as ℂ in formula-local Julia variables and as mathbb C with relation and coefficient subscripts in the theory documentation
- preserve descriptive public parameter fields while documenting their one-to-one mathematical symbols
- restore cᵃᶜ for acoustic sound speed so it is not mistaken for a calibratable coefficient
- replace overloaded V fall-speed notation with the Breeze sedimentation symbol 𝕎 throughout P3 source comments, docstrings, theory, and tests

## Testing

- P3 process tests
- P3 reference process-routing tests
- acoustic-substepping component tests
- quality assurance

All selected groups pass: 776/776 tests.

PR #987, opened 2026-09-10
(0.11.0) Bump version

ewquon:


giordano:
Do we want to wait for some other [breaking changes](https://github.com/NumericalEarth/Breeze.jl/pulls?q=is%3Apr+state%3Aopen+label%3A%22breaking+change+%3Abroken_heart%3A%22) to be merged first? No objection from me otherwise.

ewquon:
Good call, since #975 is also related to the turbulence closure let's wait on that. Also gives me a push to pick up some backburnered work.

glwagner:
> Do we want to wait for some other [breaking changes](https://github.com/NumericalEarth/Breeze.jl/pulls?q=is%3Apr+state%3Aopen+label%3A%22breaking+change+%3Abroken_heart%3A%22) to be merged first? No objection from me otherwise.

I propose to do this one since (based on what I know or rather based on what I know I need to do for them) those others may not go in until next week

glwagner:
except #893. That might be ready

giordano:
I think #957 is also ready?

glwagner:
> I think #957 is also ready?

true, that will conflict with #959, but I think #959 requires extended analysis anyways, so no big deal. I'll merge 957.

glwagner:
Actually I was confused and thinking of #969

glwagner:
which is also a breaking change

glwagner:
I wonder if we should have a grace period for very new things (like P3), so that we don't end up with 99 version bumps as we break the interface in the process of making a completely new feature more mature.

PR #988, opened 2026-09-10
[CI] Move non-GPU jobs to Julia v1.13.0

giordano:


giordano:
I may need to tweak something, but if there are no objections I'm seriously inclined to merge this, macOS tests with Julia v1.13 are _a lot_ faster than with v1.12, they finished in less than 20 minutes, when they usually take 45+ with Julia v1.12.  Also, precompilation is generally faster.  ***Edit***: I think the exceptionally fast macOS job was due to somehow getting a machine with a lot more memory than usual, ParallelTestRunner indicated almost 8 GiB available when the VM should regularly have 7 in total.  If anything, this confirm that when the macOS machine has enough memory it runs super fast, otherwise it struggled quite a bit.  ***Edit 2***: [it turns out](https://github.com/NumericalEarth/Breeze.jl/actions/runs/34538836007/job/103076610724?pr=988#step:8:12) there are M2 chips in the pool now:
```
Julia Version 1.13.0
Commit d1c37793dd2 (2026-09-09 19:00 UTC)
Build Info:
  Official https://julialang.org/ release
Platform Info:
  OS: macOS (arm64-apple-darwin25.6.0)
  CPU: 5 × Apple M2 Pro (Virtual)
  WORD_SIZE: 64
  LLVM: libLLVM-20.1.8 (ORCJIT, apple-m2)
  GC: Built with stock GC
Threads: 1 default, 1 interactive, 1 GC (on 5 virtual cores)
```

PR #996, opened 2026-09-15
Minor improvements to test suite

giordano:
Mainly removing repeated work, and useless for loops


test/p3_lookup_table_reader.jl
```diff
@@ -11,6 +11,11 @@ using Breeze.Microphysics.PredictedParticleProperties:
     prepare_interpolation,
     rain_terminal_velocities
 
+# Building the P3 scheme parses a multi-megabyte lookup table and runs the rain
+# quadrature: about 0.13 s and 130 MiB each time. The scheme is immutable and the
+# testsets below only read from it, so they share one default instance.
+const DEFAULT_P3 = PredictedParticlePropertiesMicrophysics(Float64)
+
```
giordano: I'm not convinced always defaulting to the `Float64` version (here and in all other P3 tests) is a good idea, I'll have to think more about this one. ***Edit***: I reverted this change for the time being. I think the base idea (avoid rebuilding the microphysics every single time) is correct, but the implementation is not as good (it's always forcing the same float type).

Issue #777, opened 2026-06-10
Energy-flux bottom BC on `CompressibleDynamics` errors: `ef.density === nothing` (BC materialized before `materialize_dynamics`)


ewquon:
## Summary

An `AtmosphereModel` with `CompressibleDynamics` and a bottom **energy-flux** boundary condition (the surface-coupling path — a `FluxBoundaryCondition` on `ρe`, which Breeze wraps via `energy_to_theta_bc`) errors on the first `time_step!`: the energy→θ-flux converter `𝒬_to_Jᶿ` indexes `ef.density`, which is `nothing`. The same setup with `AnelasticDynamics` works.

The root cause is a construction-ordering issue: the energy-flux BC is materialized *before* `materialize_dynamics` allocates the compressible prognostic density field, so it captures the stub `density === nothing`.

## Minimal reproduction (CPU, verified)

```julia
using Breeze, Oceananigans
using Breeze: CompressibleDynamics, SaturationAdjustment, WarmPhaseEquilibrium

grid = RectilinearGrid(size = (8, 8, 16), x = (0, 8e3), y = (0, 8e3), z = (0, 8e3),
                       halo = (5, 5, 5), topology = (Periodic, Periodic, Bounded))

Jᵉ  = Field{Center, Center, Nothing}(grid); set!(Jᵉ, 100)             # surface energy flux [W m⁻²]
bcs = (; ρe = FieldBoundaryConditions(bottom = FluxBoundaryCondition(Jᵉ)))

model = AtmosphereModel(grid;
                        dynamics            = CompressibleDynamics(surface_pressure = 1e5),
                        microphysics        = SaturationAdjustment(equilibrium = WarmPhaseEquilibrium()),
                        boundary_conditions = bcs)
set!(model; ρ = 1.2, θˡⁱ = 288.0, qᵗ = 0.0)
time_step!(model, 0.05)        # ERROR
```

Swapping `dynamics` to `AnelasticDynamics(ReferenceState(grid, ThermodynamicConstants(); surface_pressure = 1e5, potential_temperature = 288))` (and dropping `ρ = 1.2` from `set!`) runs cleanly.

## Error

```
MethodError: no method matching getindex(::Nothing, ::Int64, ::Int64, ::Int64)
Stacktrace:
 [1] 𝒬_to_Jᶿ   @ src/BoundaryConditions/thermodynamic_variable_bcs.jl:67   # ρ = @inbounds ef.density[i, j, k]
[...]

PR #997, opened 2026-09-16
Fix moisture partitioning and unify moist-state conversions

kaiyuan-cheng:
## Summary

Total moisture `qᵗ` is converted to a scheme's prognostic moisture `qᵛᵉ` by one generic
method per argument form, driven by `condensate_field_names`. Those names are the
complement of the prognostic moisture, the same partition `total_condensate_density`
already uses, so a single expression recovers vapor for non-equilibrium schemes and
equilibrium moisture `qᵉ` for saturation adjustment, with no per-scheme method.

This replaces a P3-specific `μ::NamedTuple` override and seven hand-written state-based
overloads (Kessler, P3, WP1M, MP1M, WPNE1M, MPNE1M, WPNE2M). They were a second source of
truth and had already drifted: five clamped the result at zero and two did not.

## Results change

This is a correctness fix rather than a refactor, and existing runs will produce different
numbers:

- **Saturation-adjustment schemes.** Retaining precipitation changes the diagnosed
  temperature and the diagnosed cloud liquid, and removes the spurious cooling that
  accompanied every autoconversion and accretion event.
- **Moist surface fluxes.** The bulk sensible heat flux changes wherever the near-wall air
  is moist, and both bulk scalar fluxes now sample the wall at the same time as the air.
- **Anything relying on the default `filter_timescale = Inf`.** The filtered surface state
  was frozen after initialization and now follows the current state, so bulk fluxes
  respond to the wind and the wall temperature.

Initial conditions also change wherever `qᵗ` was supplied together with condensate, since
that water was previously counted twice.

## Bugs fixed

- **Condensate was double-counted in initial conditions.** `set!` guarded the `qᵗ`
  conversion with `hasmethod`, and only P3 had a four-argument method. Every other scheme
  left `qᵗ` in the `qᵛᵉ` slot while the microphysical fields held the same water again.
- **Saturation adjustment discarded precipitation.** The adjusted state was rebuilt from
  `qᵉ` alone, so the diagnosed cloud liquid was cloud minus rain, and every autoconversion
  or accretion event appeared as a spurious cooling of `ℒΔq/cᵖ` because `qᵉ` fell while
  `θˡⁱ` stayed fixed. Precipitation now enters the heat capacity, gas constant, and latent
  terms, the convention the non-equilibrium schemes already followed.
- **The bulk sensible heat flux used a dry Exner exponent.** `wall_potential_temperature`
  formed `θ₀` with `Rᵈ/cᵖᵈ` and differenced it against a moist `θˡⁱ`. It now uses the
  near-wall composition, as `wall_static_energy` already did.
- **The parcel `θ → T` conversion was dry.** `set!(::ParcelModel)` now sets moisture first;
  the relative-humidity path, which needs `T` to diagnose saturation, takes a preliminary
  pass and repeats the conversion.
- **`filter_timescale = Inf`, the default, froze the filtered surface fields.** It drove
  `ϵ = Δt/τ` to zero, so a bulk flux never saw the wind or the wall temperature change
  after initialization. It now samples the current state, meaning "no temporal averaging"
  rather than "no update at all".
- **The two bulk scalar fluxes observed the wall at different times.** Both now filter the
  complete near-wall difference, wall value included, so with a time-varying surface
  temperature the heat and vapor fluxes agree about when the wall was sampled.
- **The parcel state ignored condensate its prognostics carried.** Water is partitioned
  before the static energy is formed, so a parcel holding rain starts at the environmental
  temperature with the right heat capacity and latent term instead of waiting for the first
  substep to rewrite it.

## Other changes

- The pressure-based and density-based saturation adjustments collapse into one secant
  iteration, the state selecting the saturation constraint and the residual. This removes
  the duplicate `LiquidIceDensityState` implementation added in #765.
- The conversion preserves total water without clipping, so a bad initial condition
  surfaces rather than silently gaining water. `set!` validates it against a tolerance
  relative to the water in each cell; under Reactant the validating reduction is a traced
  Boolean, so it is checked in a runtime callback rather than while tracing.
- `fix_negative_moisture!` gains a parcel method keeping condensate nonnegative and within
  the conserved total water.
- `specific_field_name` builds strings, which allocates and does not constant-fold, so the
  density-to-specific name mapping is generated at compile time rather than derived inside
  the kernels that use it.
- Bulk boundary conditions are materialized before the dynamics (#777), so the sensible
  heat flux captures whatever `total_density` returned on the dynamics stub. Nothing
  enforced that, and a dynamics allocating its density during materialization would hand
  the boundary condition a stale field with no error. It is checked at `initialize!` and
  the invariant is recorded in the `materialize_dynamics` docstring.
- `temperature_from_potential_temperature` and its inverse accept full
  `MoistureMassFractions`, so condensate enters both the Exner function and the latent term.


docs/src/developer/microphysics/overview.md
```diff
@@ -66,22 +66,56 @@ Velocity components are interpolated from cell faces to cell centers and passed
...
 The argument `qᵛᵉ` is the scheme-dependent specific moisture: vapor (``qᵛ``) for
-non-equilibrium schemes, or equilibrium moisture (``qᵉ = qᵛ + qᶜˡ``) for saturation
+non-equilibrium schemes, or equilibrium moisture (``qᵉ = qᵛ + qᶜˡ + qᶜⁱ``) for saturation
 adjustment schemes.
 
 **Note**: Non-equilibrium schemes don't need `𝒰` to build their state (they use prognostic fields).
 Saturation adjustment schemes override `grid_moisture_fractions` directly since they read cloud
 condensate from diagnostic fields.
 
+### Total-Water Conversion
```
glwagner:
```suggestion
### Total Water Conversion
```

ext/BreezeCloudMicrophysicsExt/one_moment_microphysics.jl
```diff
@@ -845,11 +829,12 @@ end
 @inline AM.maybe_adjust_thermodynamic_state(𝒰₀, bμp::NonEquilibrium1M, qᵛ, constants) = 𝒰₀
 
 # Saturation adjustment (warm-phase and mixed-phase)
-@inline function AM.maybe_adjust_thermodynamic_state(𝒰₀, bμp::Union{WP1M, MP1M}, qᵉ, constants)
-    q₁ = MoistureMassFractions(qᵉ)
+@inline function AM.maybe_adjust_thermodynamic_state(𝒰₀, bμp::Union{WP1M, MP1M}, qᵉ, constants, μ, ρ)
+    qʳ = μ.ρqʳ / ρ
+    qˢⁿ = get(μ, :ρqˢⁿ, zero(ρ)) / ρ
```
glwagner, line 834: can we use dispatch here, not `get` ?

src/AtmosphereModels/microphysics_interface.jl
```diff
@@ -385,13 +385,29 @@ the specific (per-mass) name. For example, `:ρqᶜˡ` → `:qᶜˡ`.
 """
 specific_field_name(name::Symbol) = (s = string(name); Symbol(s[nextind(s, 1):end]))
 
+# Resolve names to literals at compile time: string processing cannot run inside GPU kernels.
+# QuoteNode keeps a scalar Symbol literal from being interpreted as a variable name.
+@generated specific_field_name(::Val{name}) where name = QuoteNode(specific_field_name(name))
+@generated specific_field_names(::Val{names}) where names = :($(map(specific_field_name, names)))
```
glwagner, line 392: what's the purpose of this?

src/AtmosphereModels/microphysics_interface.jl
```diff
@@ -636,20 +657,67 @@ $(TYPEDSIGNATURES)
...
+@inline specific_prognostic_moisture_from_total(microphysics, qᵗ, ℳ::AbstractMicrophysicalState) =
+    moisture_less_condensate(qᵗ, ℳ, specific_condensate_names(microphysics))
 
-# Generic fallback: no condensate prognostics → all moisture is vapor/equilibrium.
+# An empty state has no condensate to subtract, even if the scheme names condensate species.
 @inline specific_prognostic_moisture_from_total(microphysics, qᵗ, ::NothingMicrophysicalState) = qᵗ
 
+@inline moisture_less_condensate(qᵗ, ℳ, ::Tuple{}) = qᵗ
+@inline moisture_less_condensate(qᵗ, ℳ, names::Tuple{Symbol, Vararg}) =
+    qᵗ - sum_microphysical_components(ℳ, names)
```
glwagner:
```suggestion
@inline moisture_less_condensate(qᵗ, ℳ, names::Tuple{Symbol, Vararg}) = qᵗ - sum_microphysical_components(ℳ, names)
```

glwagner: I don't totally understand the names. "moisture less condensate" is q^t - q^c ? is that different from the vapor mass fraction?

src/AtmosphereModels/microphysics_interface.jl
```diff
@@ -636,20 +657,67 @@ $(TYPEDSIGNATURES)
...
+per-scheme method needed.
+
+```jldoctest
+using Breeze
+using Breeze.AtmosphereModels: specific_prognostic_moisture_from_total
+using Breeze.Microphysics: DCMIP2016KesslerMicrophysics
+
+microphysics = DCMIP2016KesslerMicrophysics()
+μ = (ρqᶜˡ=0.0012, ρqʳ=0.0024)
+specific_prognostic_moisture_from_total(microphysics, 0.02, μ, 1.2)
```
glwagner: why do we need the suffix `_from_total` here? Is there another competing method `specific_prognostic_moisture` that we need to distinguish?

src/AtmosphereModels/microphysics_interface.jl
```diff
@@ -636,20 +657,67 @@ $(TYPEDSIGNATURES)
...
+# output
+0.017
+```
+"""
+@inline specific_prognostic_moisture_from_total(microphysics, qᵗ, μ::NamedTuple, ρ) =
+    moisture_less_condensate(qᵗ, μ, ρ, condensate_field_names(microphysics))
+
+@inline moisture_less_condensate(qᵗ, μ, ρ, ::Tuple{}) = qᵗ
+@inline moisture_less_condensate(qᵗ, μ, ρ, names::Tuple{Symbol, Vararg}) =
+    qᵗ - sum_microphysical_components(μ, names) / ρ
```
glwagner:
```suggestion
@inline moisture_less_condensate(qᵗ, μ, ρ, names::Tuple{Symbol, Vararg}) = qᵗ - sum_microphysical_components(μ, names) / ρ
```

glwagner: does this imply that `sum_microphysical_components` returns a density? If so it should be in the name. It seems like this could be `condensate_density`.

src/AtmosphereModels/microphysics_interface.jl
```diff
@@ -743,6 +811,15 @@ end
     return ρqˣ + sum_microphysical_densities(i, j, k, microphysical_fields, Base.tail(names))
 end
 
+# Sum scalars, whole fields, or state components. Stop at the last component to avoid
+# adding a scalar zero to lazy field expressions; moisture_less_condensate handles empty tuples.
+@inline sum_microphysical_components(μ, names::Tuple{Symbol}) = getproperty(μ, first(names))
```
glwagner: this method doesn't seem specific to microphysics

src/ParcelModels/parcel_dynamics.jl
```diff
@@ -925,18 +972,56 @@ function ssp_rk3_parcel_substep!(model::ParcelModel, U⁰::ParcelInitialState, 
...
+end
 
+function update_parcel_moisture!(model)
+    AtmosphereModels.fix_negative_moisture!(model)
+    state = model.dynamics.state
+    microphysics = model.microphysics
+    qᵛᵉ, q = parcel_moisture_partition(microphysics, state.ρ, state.μ, state.qᵗ, state.𝒰)
+    𝒰 = with_moisture(state.𝒰, q)
+    state.𝒰 = AtmosphereModels.maybe_adjust_thermodynamic_state(
+        𝒰, microphysics, qᵛᵉ, model.thermodynamic_constants, state.μ, state.ρ)
```
glwagner:
```suggestion
    state.𝒰 = AtmosphereModels.maybe_adjust_thermodynamic_state(𝒰, microphysics, qᵛᵉ, model.thermodynamic_constants, state.μ, state.ρ)
```

src/AtmosphereModels/microphysics_interface.jl
kaiyuan-cheng: This helper subtracts the species listed by the microphysics scheme:

  - Non-equilibrium microphysics: subtract all hydrometeors. The result is vapor mass fraction.
  - Saturation adjustment: subtract only precipitating hydrometeors. The result is equilibrium moisture, qᵉ = qᵛ + qᶜˡ + qᶜⁱ, which still includes cloud droplet and cloud ice.

We can come up with a better name.

src/AtmosphereModels/microphysics_interface.jl
kaiyuan-cheng: Yes. There is `specific_prognostic_moisture(model)`, which retrieves the model’s existing qᵛ or qᵉ field.

`specific_prognostic_moisture_from_total` instead converts supplied total water into qᵛ or qᵉ by subtracting the appropriate condensates.

We could use dispatch so that we don't need the suffix.

src/AtmosphereModels/microphysics_interface.jl
kaiyuan-cheng: I will go with `subtract_condensate`. Feel free to suggest a new name.

src/AtmosphereModels/microphysics_interface.jl
kaiyuan-cheng: You are right. Moved this into `Breeze.Utils`.

PR #998, opened 2026-09-16
Add column batching for the RRTMGP all-sky solver

kaiyuan-cheng:
## Summary

Radiative transfer is column-local, so the RRTMGP solver workspace does not have to cover the
whole domain at once. Adds `column_batch_size` to the all-sky `RadiativeTransferModel`
constructor: the workspace is sized for one batch of columns, which `solve_radiation_batches!`
sweeps across the domain. The state stays full-size and each batch solves a `view` of it.

## Measurements

3 km CONUS over the HRRR footprint, 2628 x 1152 x 50 = 151.4M cells (3.03M columns) in `Float32`,
on one H100 80GB. Median of 5 solves after a warmup.

| `column_batch_size` | columns/batch | batches | RRTMGP memory | saved | solve | vs unbatched |
|---|---|---|---|---|---|---|
| `nothing` | 3,027,456 | 1 | 33.19 GiB | - | 16.95 s | 1.000x |
| 262144 | 336,384 | 9 | 13.84 GiB | 19.34 GiB | 16.98 s | **1.002x** |
| 65536 | 84,096 | 36 | 12.03 GiB | 21.16 GiB | 19.74 s | 1.165x |
| 16384 | 21,024 | 144 | 11.59 GiB | 21.59 GiB | 26.01 s | 1.535x |

The memory column is the GPU delta across `RadiativeTransferModel` construction. It splits cleanly
into a **21.8 GiB workspace that batches** and a **11.4 GiB state that does not**: predicted vs
measured agrees to 0.01 GiB at every batch size. So the reachable saving is the workspace, and
batching recovers 89% of it at 9 batches and 99% at 144.

**Cost is an occupancy tradeoff, not a constant overhead.** The solver runs one thread per column,
so a batch that still saturates the GPU is free: at 336k columns/batch the solve is within 0.2% of
unbatched. Below roughly 10^5 columns/batch the kernel stops saturating and time scales with the
batch count.

**Results are bitwise identical.** Every batch size reproduces the unbatched fluxes exactly, over
all five output fields (`lw_up`, `lw_dn`, `sw_up`, `sw_dn`, `flux_divergence`), 154M elements each,
zero differing elements. Two solves from an identical state also agree bitwise, so the McICA cloud
sampler is per-column and independent of how columns are grouped.

## Design notes

- **Batches are whole latitude rows.** `rrtmgp_column_index` is `i + (j - 1) * Nx`, so a j-slab is
  a unit-stride `view`. On the GPU a view of a `CuArray` over such a range is itself a `CuArray`,
  so the solver kernels see what they would see unbatched.
- **Batches must be equal-width.** `TransposedStateCache` is sized from `RRTMGPGridParams(; ncol)`
  and refreshed with `permutedims!`, which throws `DimensionMismatch` on a short final batch.
  `resolve_column_batch_rows` rounds up to a divisor of `Ny` and warns when that lands more than 2x
  above the request.
- **The unbatched default is unchanged.** `column_batch_size = nothing` short-circuits to RRTMGP's
  own `update_lw_fluxes!` / `update_sw_fluxes!`, which dispatch on `radiation_method`. Avoids a
  second specialization of the RRTMGP kernel stack for `SubArray` state on CPU.
- **Column views are explicit.** A generic `Adapt` traversal was rejected: the `TwoStream*RTE`
  buffers are shared batch-width workspace while `columns` is a global range, and `VmrGM.vmr` is
  per-gas. Two `TODO`s mark the upstream asks that would remove the need.
- `LwBCs.inc_flux` and `SwBCs.inc_flux_diffuse` are `(ncol, ngpt)`, column-*first*, so they get a
  separate `incident_flux_view`. Both are `nothing` in Breeze today; slicing them on the trailing
  dimension would have silently sliced g-points.

## Not covered

- No test exercises the batched path (all all-sky tests are `Nx = Ny = 1`).
- Only the workspace is batched; the 11.4 GiB of per-column state stays `O(Nx Ny Nz)`.
- Clear-sky is not batched: `solve_radiation_batches!` hardcodes the all-sky lookup triple.


ext/BreezeRRTMGPExt/rrtmgp_column_batching.jl
glwagner, line 1: this code seems general for other radiation solvers too, is it specific to RRTMGP?

ext/BreezeRRTMGPExt/rrtmgp_column_batching.jl
```diff
@@ -0,0 +1,190 @@
...
+#####
+##### Column views of RRTMGP's state types
+#####
+#
+# Each rebuilds the struct around `view`s of the column dimension; non-column members pass
+# through unchanged.
+#
+# TODO: these transcribe RRTMGP's positional constructors, coupling us to its field order.
+# RRTMGP has no column-subsetting API; the fix is upstream — a way to rebind `bcs` on an existing
+# workspace, or a column-range argument on `update_lw_fluxes!`/`update_sw_fluxes!`.
```
glwagner, line 67: I'm surprised this helps. Can you explain why?

PR #1003, opened 2026-09-17
Correct the documented dimensions of the P3 effective radius

masonlee277:
The entry for `effective_radius` in the `IceBulk` docstring lists

```
r_e = ∫A·N'dD / ∫N'dD
```

which is the number-weighted mean projected **area**, not a radius: with `A` in m² and `N'` in m⁻⁴, the ratio evaluates to m². (`A(D)` is projected area elsewhere in the package — see the sweep-out integral `∫ 𝕎(D) A(D) N'(D) dD` documented in `lookup_table_format.jl`.)

The three neighbouring entries are dimensionally correct (`mean_diameter` → m, `mean_density` → kg/m³, `reflectivity` → m³), so only this one changes.

I did not substitute a corrected integral. `effective_radius` is read from the P3 lookup table rather than integrated at runtime (`lookup_table_reader.jl` fills it from `ice_4d[:effective_radius]`), and the integral that generates that column lives in the table-generation source outside this repository — so rather than guess at it, the entry now names the quantity and its units.

Docstring only; no behaviour change.

Supersedes #981, which was closed; this is the same one-line change rebased onto current `main`.


src/Microphysics/PredictedParticleProperties/ice_properties.jl
```diff
@@ -161,7 +161,7 @@ size distribution. They are used for radiation, radar, and diagnostics.
 
 **Diagnostic integrals:**
 
-- `effective_radius`: Radiation-weighted radius ``r_e = ∫A·N'dD / ∫N'dD``
+- `effective_radius`: Radiative effective radius [m], tabulated
```
kaiyuan-cheng, line 164:
```suggestion
- `effective_radius`: Area-weighted mean radius ``(3 / (4 ρ_i*)) · ∫m(D)N'(D)dD / ∫A(D)N'(D)dD``
```

masonlee277:
The red CI here is not from this PR — it is an Aqua regression released a few hours ago.

**Root cause.** Aqua **v0.8.17** was published today at 16:46 UTC and adds a new `manifest_entries` function in `persistent_tasks.jl` (it does not exist in v0.8.16). It walks the dependency graph and does:

```julia
for (name, dep_uuid) in get(prj, "deps", Dict{String,Any}())
    srcpath = Base.locate_package(PkgId(UUID(dep_uuid), name))
    srcpath === nothing &&
        error("Unable to locate `$name`, a dependency of `$(prj["name"])`")
```

Many packages list a weak dependency in **both** `[deps]` and `[weakdeps]` — the standard pattern for supporting Julia < 1.9 alongside extensions. Those are not installed, so `locate_package` returns `nothing` and Aqua errors rather than skipping. In this tree that fires on:

- `ChainRulesCore`, a weakdep of `SpecialFunctions`
- `ChangesOfVariables`, a weakdep of `LogExpFunctions`
- `RecipesBase`, a weakdep of `Rotations`

**Evidence it is not this PR.** This branch is `+1/-1` in a docstring against `b177dcda`, whose own `main` CI passed at 14:42 — before the Aqua release. Comparing the runs:

| run | Aqua | result |
|---|---|---|
| `main` @ b177dcda, 14:42 | — | pass |
| `glw/numerical-radiation`, 19:04 | **v0.8.16** (cached) | pass |
| this PR, 19:20 | **v0.8.17** | fail |

In-repo branches still had v0.8.16 in their cache; this branch resolved fresh and picked up v0.8.17.

The single Aqua error then interrupts ~46 other testsets under the parallel test runner, which is why the check list looks far worse than it is. The same regression is failing CliMA/Oceananigans.jl PRs right now, where the unit-test job reports **4,013,579 passed, 0 failed, 1 errored** — that one error being this.

Pinning `Aqua = "~0.8.16"` in `test/Project.toml` would unblock CI until the upstream fix lands. Happy to open that as a separate PR if you would like it.

PR #1004, opened 2026-09-17
Declare transport bounds in each microphysical quantity's units

masonlee277:
P3 carries mass fractions, number concentrations, a rime volume and a signed supersaturation. All but the last are non-negative, and for a positivity-preserving limiter that is the whole of what matters.

| quantity | bound |
|---|---|
| `ρqᶜˡ, ρqʳ, ρqⁱ, ρqᶠ, ρqʷⁱ, ρnᶜˡ, ρnʳ, ρnⁱ, ρnᵃ, ρbᶠ` | `(0, 1)` |
| `ρsᵛ⁺ˡ` | `nothing` |

### Why the signed one is the point

`ρsᵛ⁺ˡ` is `qᵛ - qᵛ⁺ˡ` and is legitimately negative wherever the air is subsaturated. A lower bound of zero limits a correct field hard: a cell mean of `-0.01` reconstructing to `-0.1` gives `θ ≈ 0.11`. Every other P3 prognostic wants positivity; this one must be left alone. That single exception is why the method exists — without it a blanket `(0, 1)` over `prognostic_field_names` would be very nearly right.

### Correcting my own framing

An earlier revision of this PR claimed a blanket `(0, 1)` "clips a number concentration of order 1e8 kg⁻¹ to 1". **That is false and I should not have written it.** The limiter forms `θ = min(θᵐᵃˣ, θᵐⁱⁿ, 1)` and rescales toward the cell mean; with `cᵐᵃˣ = 1` far below the mean, `θᵐᵃˣ` is enormous and the `min` discards it. Nothing is clipped, and mass is conserved exactly.

What the *lower* bound does is the part worth keeping — it is the positivity. Over 60 steps of a 1e8-amplitude field, unlimited advection reaches `-6.0e4` while bounded stays at `+6.1e-9`.

So a second revision, which returned `nothing` for the number concentrations to avoid an `Inf` the consumer cannot accept, was a regression: it gave up positivity on exactly the fields whose negatives are the problem. This revision restores it.

The nominal upper bound of `1` on a number concentration is honest about being nominal. Since `θᵐᵃˣ` reduces to about `cᵢ / (M - cᵢ)`, it is inert unless the reconstruction overshoots the cell mean by more than a factor of two — and where it does engage, it is damping a greater-than-100% overshoot that a positivity-preserving scheme wants damped anyway.

### The real limitation, which is upstream

None of this would need saying if `bounds` could express "non-negative, no upper bound". It cannot:

- `(0, Inf)` is `Tuple{Int, Float64}`, and `(0, Inf) isa NTuple{2}` is `false`, so `WENO` rejects it at construction.
- `(0.0, Inf)` constructs and the primal is clean, but `Inf` in `abs((cᵐᵃˣ - cᵢ)/(M - cᵢ + ε₂))` gives a NaN reverse-mode gradient — confirmed with Enzyme, and it NaNs even in the `θ = 1` branch where no limiting happens at all.

So a one-sided `bounds` in Oceananigans would let every entry in that table become the positivity constraint it actually is, and the per-quantity table would collapse to a single rule plus the one signed exception. I will open that upstream separately.

### Unrelated, worth flagging upstream

In `Float32`, the hardcoded `ε₂ = 1e-20` squares to `1e-40` in the reverse rule for `a/b²` and overflows, so `bounds` already produces NaN gradients wherever the reconstruction is flat, independently of anything here.

### Tests

46 assertions: both aerosol settings (the prognostic list differs, so `ρnᶜˡ` and `ρnᵃ` are otherwise never reached), the `:ρqᵛ` and unknown-name paths, the generic fallback, and that everything reported is accepted by `WENO(; bounds)`.

Supersedes #984, which was closed.


masonlee277:
The red CI here is not from this PR — it is an Aqua regression released a few hours ago.

**Root cause.** Aqua **v0.8.17** was published today at 16:46 UTC and adds a new `manifest_entries` function in `persistent_tasks.jl` (it does not exist in v0.8.16). It walks the dependency graph and does:

```julia
for (name, dep_uuid) in get(prj, "deps", Dict{String,Any}())
    srcpath = Base.locate_package(PkgId(UUID(dep_uuid), name))
    srcpath === nothing &&
        error("Unable to locate `$name`, a dependency of `$(prj["name"])`")
```

Many packages list a weak dependency in **both** `[deps]` and `[weakdeps]` — the standard pattern for supporting Julia < 1.9 alongside extensions. Those are not installed, so `locate_package` returns `nothing` and Aqua errors rather than skipping. In this tree that fires on:

- `ChainRulesCore`, a weakdep of `SpecialFunctions`
- `ChangesOfVariables`, a weakdep of `LogExpFunctions`
- `RecipesBase`, a weakdep of `Rotations`

**Evidence it is not this PR.** This branch is `+1/-1` in a docstring against `b177dcda`, whose own `main` CI passed at 14:42 — before the Aqua release. Comparing the runs:

| run | Aqua | result |
|---|---|---|
| `main` @ b177dcda, 14:42 | — | pass |
| `glw/numerical-radiation`, 19:04 | **v0.8.16** (cached) | pass |
| this PR, 19:20 | **v0.8.17** | fail |

In-repo branches still had v0.8.16 in their cache; this branch resolved fresh and picked up v0.8.17.

The single Aqua error then interrupts ~46 other testsets under the parallel test runner, which is why the check list looks far worse than it is. The same regression is failing CliMA/Oceananigans.jl PRs right now, where the unit-test job reports **4,013,579 passed, 0 failed, 1 errored** — that one error being this.

Pinning `Aqua = "~0.8.16"` in `test/Project.toml` would unblock CI until the upstream fix lands. Happy to open that as a separate PR if you would like it.

kaiyuan-cheng:
I think a better solution is to have a high-order positive-definite advection.

masonlee277:
Reworked after review. The previous revision returned `(0, Inf)` for number concentrations, which **is not accepted by the consumer** — `WENO` validates `bounds isa NTuple{2}` and `(0, Inf) isa NTuple{2}` is `false`. Forcing it homogeneous would have been worse: `Inf` in `abs((cᵐᵃˣ - cᵢ) / (M - cᵢ + ε₂))` is discarded by the `min` in the primal but leaves a `0 * Inf` NaN in the adjoint.

Now only two-sided bounds are reported, the P3 method no longer throws on `:ρqᵛ` (which a caller iterating advected names will ask about), and the tests assert the returned bounds actually construct a `WENO`.

masonlee277:
Yeah, I think you're right — and having gone and measured it, most of the evidence is sitting in this PR.

Two things I got wrong in my original write-up:

**The motivating claim was false.** I said a blanket `(0, 1)` would clip a number concentration of ~1e8 to 1. It doesn't. `θᵐᵃˣ` blows up when `cᵐᵃˣ` sits far below the cell mean and the `min` throws it away — nothing gets clipped and mass is conserved bitwise. I've corrected the PR body.

**And to your point:** the only thing those `(0, 1)` entries are actually doing is enforcing positivity. Since `θᵐᵃˣ` reduces to about `cᵢ/(M - cᵢ)`, the upper bound is inert unless the reconstruction overshoots the mean by more than 2×. So the per-quantity table is really a workaround for `bounds` needing two sides when I only ever wanted one. If it could express positivity directly, the table collapses to "everything except supersaturation is non-negative" and most of this PR goes away.

I did try to say it directly and it doesn't work today:

- `(0, Inf)` fails the `bounds isa NTuple{2}` check — it's `Tuple{Int, Float64}`.
- `(0.0, Inf)` constructs fine and the primal is clean, but the `Inf` in `abs((cᵐᵃˣ - cᵢ)/(M - cᵢ + ε₂))` NaNs the reverse-mode gradient. It NaNs even in the `θ = 1` branch where nothing is being limited at all. That one matters to us since we differentiate through this.

The bit that survives whatever the scheme is: `ρsᵛ⁺ˡ` is signed (`qᵛ - qᵛ⁺ˡ`), so positivity is simply wrong for it — a cell mean of `-0.01` reconstructing to `-0.1` gives `θ ≈ 0.11` on a perfectly correct field. A positive-definite scheme would limit it just as wrongly as `(0, 1)` does, so something still has to say "not this one". That's the part I'd like to keep, and it's what the current default in NumericalEarth gets wrong — it bounds every microphysical prognostic including that one.

So I'd suggest: keep this narrow as a Breeze-side trait, and I'll raise the one-sided `bounds` question upstream in Oceananigans, which is the real version of what you're describing.

One unrelated thing I hit while measuring, since it's in the same operator: in `Float32` the hardcoded `ε₂ = 1e-20` squares to `1e-40` in the `a/b²` reverse rule and overflows, so `bounds` already produces NaN gradients wherever the reconstruction is flat. Nothing to do with this PR, but worth someone knowing.

PR #1005, opened 2026-09-17
Borrow between species in partial densities, without a round trip through q

masonlee277:
The species-borrowing correction moves mass between reservoirs that are all partial densities — `total_condensate_density` sums them with a bare `+`, no density weighting. It nevertheless divided each one by the local density, transferred in mass-fraction units, and multiplied back.

Those two steps cancel exactly in real arithmetic. They do not in floating point: `ρ * (ρq / ρ)` does not return `ρq`. So

- a deficit its donors **fully cover** settles *near* zero instead of *on* it, and
- a donor drained to empty can be pushed **slightly negative** — a new negative created in a field that was strictly positive.

That is the postcondition of this routine being violated by the routine itself. The result is also density-dependent (same physical state, different `dynamics_density`, different answer) and not idempotent.

### Evidence

Replicating both versions over the test's cases, at five reference densities:

| | not exactly zero | spurious negative | density-dependent | non-idempotent |
|---|---|---|---|---|
| `main`, Float32 | 4 | 4 | 8 | 4 |
| `main`, Float64 | 5 | 1 | 7 | 1 |
| this PR | **0** | **0** | **0** | **0** |

Float64 fails too, so this is not a Float32-only artifact. Over 400k randomized cells, 7.6% of Float32 cells kept a spurious negative (worst `-9.3e-10` kg/m³) and 2.3% had a donor driven negative; both go to zero here, and bitwise-exact conservation becomes *more* frequent (72.4k/200k → 82.0k/200k), not less.

### The change

Transfer directly in partial densities. The conversion was pure overhead, so `ρ` stops being read anywhere in the same-level chain and is dropped from it; vertical borrowing still takes `ρ₀`, so the kernel signature is unchanged. Two divisions per field per donor also go away.

One detail worth naming: the recipient is written as `ifelse(mass < 0, zero(mass) - remaining, mass)`. Writing `zero(mass) - remaining` rather than `-remaining` keeps a fully funded deficit at `+0.0` instead of `-0.0`, and is exactly as precise for a partially funded one.

### Scope note

The recipient is written from a value read before the donor sweep, which is correct only because the recipient is not also one of its own donors and never aliases vapor. That holds for P3's chain and for any sensible ordering, but it is now an implicit requirement rather than an accident of `+=`.

### Test

`test/species_borrowing_roundoff.jl`, 254 assertions over both precisions, checking against a `BigFloat` oracle: conservation, exact zero on a fully funded deficit, no spurious negatives, and idempotency. It asserts `iszero` rather than a tolerance, because a tolerance is precisely what would hide this.

The test calls `same_level_borrow!` directly on the CPU rather than through a launched kernel — it is pointwise and architecture-independent, and this avoids adding `KernelAbstractions` as a test dependency for one file. The P3 integration tests cover it on whichever architecture they run.

Supersedes #982, which was closed. Rebased onto current `main`; the earlier revision described this as a units error, which was wrong — the old code is dimensionally consistent, and the defect is the inexact round trip.


masonlee277:
Extended after review — the vertical phase had the **same defect**, and it runs last, so it had the final say on the sign.

```julia
deficit = ifelse(qᵛ < 0, -ρqᵛ_k * Δz_k, zero(ρqᵛ_k))
ρqᵛᵉ[i, j, k] += deficit / Δz_k          # -> 0 when deficit > 0
```

`ρqᵛ + fl(fl(-ρqᵛ Δz) / Δz)` is not zero unless Δz is a power of two, so the level it had just repaired kept a residual negative. Fixing only the species phase would have left the routine still emitting negatives on its last pass.

Also: the sign tests there divided by ρ purely to ask `qᵛ < 0`. ρ is positive, so `ρqᵛ < 0` answers it identically — the same round trip this PR removes, three lines below where it removed it. With those gone **nothing in the correction needs the density**, so `ρ₀` is dropped from `vertical_borrow!`, `apply_vertical_correction!`, the column kernel, and `fix_negative_moisture!`.

One honest note on test quality: my first draft of the vertical test **passed against the old code**. With uniform Δz the multiply and divide cancel exactly, so the defect is undetectable. The committed version uses spacings and columns found by searching for cases the old formulation actually gets wrong — it fails 2 of 4 in Float32 and 3 of 4 in Float64 before the change, and none after.

282 assertions total, all passing.

PR #1006, opened 2026-09-17
Apply the boundary conditions given for P3 prognostic fields

masonlee277:
`AtmosphereModel` collects boundary conditions for every prognostic name, validates them, and passes the regularized inventory to `materialize_microphysical_fields`. P3's method accepts that argument and never reads it:

```julia
function AM.materialize_microphysical_fields(p3::P3, grid, bcs)
    ρqᶜˡ = CenterField(grid)  # Cloud liquid
    ρqʳ  = CenterField(grid)  # Rain mass
    ...
```

Because the P3 names are legitimate — they come from `prognostic_field_names`, so `validate_boundary_condition_names` accepts them — a prescribed surface rain flux or a Dirichlet value on any P3 prognostic was **silently dropped**. No error, no warning, no effect. `DCMIP2016Kessler` gets this right one file over (`dcmip2016_kessler.jl`):

```julia
ρqᶜˡ = CenterField(grid, boundary_conditions=boundary_conditions.ρqᶜˡ)
```

The optional field groups (`aerosol_activation_fields`, `supersaturation_fields`) need the inventory for the same reason. A name that is absent falls back to the default, because diagnostics may materialize these fields with an empty tuple — `predicted_particle_properties_integration.jl` and `..._processes.jl` both do.

### A second problem this exposes

Giving P3 prognostics real boundary conditions surfaces a latent bug. `compute_auxiliary_thermodynamic_variables!` refills the halos of the *whole* microphysical inventory — which holds the prognostic densities as well as the diagnostics it just computed — but did so without `boundary_condition_args(model)`:

```julia
fill_halo_regions!(model.microphysical_fields)
```

A constant condition does not notice. A function-valued one is evaluated with the wrong signature. That fill now passes the same arguments as the prognostic fill in `update_state!`, which is what the three other `fill_halo_regions!` calls in that file already do.

I kept this here rather than in its own PR because it is only reachable once P3 prognostics carry conditions, so the two belong to the same change.

### Scope

This is the first of three pieces split out of the closed #983, which bundled them behind a P3-shaped title. This piece is the P3 plumbing only. The remaining two concern translating a `Value` on a prognostic *density* onto the *specific* field that transport actually reconstructs — which is not P3-specific and is a live bug for moisture and `ρθ` under every scheme today. That deserves its own review and its own model-level test, so it will come separately.

### Test

A P3 model built with a constant flux on `ρqʳ` and a function-valued `Value` on `ρqⁱ`: the conditions are on the fields, an unspecified prognostic keeps its default, and `update_state!` runs and leaves the field finite. The function-valued condition is the part that exercises the halo-fill fix. 6/6 passing on GPU.


glwagner:
hmm may be a good catch

kaiyuan-cheng:
This is a real bug, but it has remained latent because we do not currently have a use case that requires explicit specification of microphysical fluxes.

kaiyuan-cheng:
The first bug is not P3-specific. 1M and 2M are also having the same bug. We may want to come up with a generic fix.

glwagner:
> This is a real bug, but it has remained latent because we do not currently have a use case that requires explicit specification of microphysical fluxes.

don't we simply need to add a test to the test suite? do you mean that such a case does not exist among all Breeze users (who is "we")?

PR #1011, opened 2026-09-18
Decouple prognostic aerosol from prognostic droplet number in P3

kaiyuan-cheng:
## Summary

P3 gated both the cloud droplet number `ρnᶜˡ` and the unactivated aerosol reservoir `ρnᵃ`
on a single check, `p3.aerosol === nothing`. Predicting droplet number from Morrison &
Grabowski (2007) activation therefore forced an aerosol budget along with it, with no way
to get the former without the latter.

This adds `prognostic_aerosol` to `AerosolActivation`, making the two independent:

```julia
aerosol = AerosolActivation(AerosolMode())                            # fixed population (default)
aerosol = AerosolActivation(AerosolMode(); prognostic_aerosol = true) # depleting reservoir
```

|         | `aerosol = nothing` | `prognostic_aerosol = false` | `prognostic_aerosol = true` |
|---------|---------------------|------------------------------|-----------------------------|
| `ρnᶜˡ`  | —                   | prognostic                   | prognostic                  |
| `ρnᵃ`   | —                   | —                            | prognostic                  |

With a fixed population the `min(N_act, nᶜˡ + nᵃ)` cap cannot bind, because
`total_activated_number` is already capped at the distribution total, so activation reduces
to `max(0, N_act(S) - nᶜˡ) / ℂᶠᵒʳᵐ₄` — a plain relaxation toward the equilibrium count.
That path routes through the pre-existing
`prognostic_ccn_activation_rate(aerosol, nᶜˡ, qᵛ, qᵛ⁺ˡ, T)` method, which already defaulted
the pool to `sum_aerosol_number`, so `ℳ.nᵃ` keeps a single meaning: the prognostic
reservoir, or zero.

## Implementation

The switch is a type parameter, not a field, so every gate keyed on it folds to a
compile-time constant — the same reason `ProcessRate` carries `predict_supersaturation` in
its type, and a requirement here because a `Union` return from `prognostic_field_names`
would make the GPU `extract_microphysical_prognostics` recursion allocate. Keeping it on
`AerosolActivation` rather than on the scheme also makes `prognostic_aerosol = true` with
`aerosol = nothing` unrepresentable, so no constructor-time validation is needed.

Gated on the parameter: `aerosol_prognostic_names`, `aerosol_correction_fields`,
`aerosol_activation_fields` (allocation), `grid_aerosol_number`,
`write_aerosol_diagnostic!`, `add_p3_aerosol_tendency!`, `p3_aerosol_tendency_fields`, and
`compute_ccn_activation`.

## What `prognostic_aerosol = false` saves

Four arrays, not one: the `ρnᵃ` field, its `nᵃ` diagnostic mirror, the `Gⁿ` tendency slot,
and the SSP-RK3 `U⁰` stage copy. On a 64³ Float64 grid with halo 3 that is 10.5 MiB of
149.5 MiB (~7%); it scales as `4 × (Nx+2H)(Ny+2H)(Nz+2H) × sizeof(FT)`, so ~1.05 GiB at
512×512×256 in Float32. It also drops one tracer from advection and halo exchange per RK
stage.

## Behavior change

`aerosol = AerosolActivation(AerosolMode())` written against the previous code now gets one
fewer prognostic and no aerosol depletion. This is the intended new default, but it is not
additive — worth a release note.

## Testing

All green locally: `aerosol_activation` 89 (36 new), P3 atmosphere integration 130, P3
processes 433, P3 integrals 101, reference process routing 66, trace regressions 34,
`parcel_dynamics`, `number_concentration`, `doctests`, and all five ExplicitImports checks.

Physics is unchanged for both configurations: a 40-step two-configuration run reproduces
`nᶜˡ`, `nᵃ` and `qᶜˡ` bit-for-bit against the pre-refactor code.


docs/src/appendix/notation.md
```diff
@@ -79,10 +79,10 @@ The following table also uses a few conventions that suffuse the source code and
 | ``ρqˢⁿ``                            | `ρqˢⁿ` | `AM.microphysical_fields.ρqˢⁿ`      | Snow density                                                                   |
 | ``n^{cl}``                          | `nᶜˡ`  | `AM.microphysical_fields.nᶜˡ`       | Cloud droplet number per unit mass (1/kg); P3 allocates this diagnostic only with aerosol activation |
 | ``n^r``                             | `nʳ`   | `AM.microphysical_fields.nʳ`        | Rain drop number per unit mass (1/kg)                                          |
-| ``n^a``                             | `nᵃ`   | `AM.microphysical_fields.nᵃ`        | Aerosol number per unit mass (1/kg); allocated only with aerosol activation    |
+| ``n^a``                             | `nᵃ`   | `AM.microphysical_fields.nᵃ`        | Aerosol number per unit mass (1/kg); only with `prognostic_aerosol`            |
```
glwagner, line 82:
```suggestion
| ``n^a``                             | `nᵃ`   | `AM.microphysical_fields.nᵃ`        | Aerosol number per unit mass (1/kg)            |
```

Since the symbol is valid regardless of whether or not aerosol is prognostic.

docs/src/appendix/notation.md
```diff
@@ -79,10 +79,10 @@ The following table also uses a few conventions that suffuse the source code and
...
 | ``\rho n^{cl}``                     | `ρnᶜˡ` | `AM.microphysical_fields.ρnᶜˡ`      | Cloud droplet number density (1/m³), prognostic (absent in P3's prescribed-Nᶜˡ path) |
 | ``\rho n^r``                        | `ρnʳ`  | `AM.microphysical_fields.ρnʳ`       | Rain drop number density (1/m³), prognostic                                    |
-| ``\rho n^a``                        | `ρnᵃ`  | `AM.microphysical_fields.ρnᵃ`       | Aerosol number density (1/m³), prognostic (only with aerosol activation)        |
+| ``\rho n^a``                        | `ρnᵃ`  | `AM.microphysical_fields.ρnᵃ`       | Aerosol number density (1/m³), prognostic (only with `prognostic_aerosol`)      |
```
glwagner, line 85:
```suggestion
| ``\rho n^a``                        | `ρnᵃ`  | `AM.microphysical_fields.ρnᵃ`       | Aerosol number density (1/m³)      |
```

src/Microphysics/PredictedParticleProperties/aerosol_activation.jl
```diff
@@ -174,8 +183,19 @@ length(aerosol.modes)
...
+```jldoctest
+using Breeze.Microphysics.PredictedParticleProperties: AerosolActivation, AerosolMode
+summary(AerosolActivation(AerosolMode()))
+
+# output
+"AerosolActivation(1 mode, fixed reservoir)"
+```
 """
 function AerosolActivation(mode1::AerosolMode{FT}, rest::AerosolMode{FT}...;
+                           prognostic_aerosol = false,
```
glwagner, line 198:
```suggestion
                           prognostic = false,
```

just an idea --- since this is already the `AerosolActivation` type, it looks like this

```julia
activation = AerosolActivation(prognostic=true)
```

src/Microphysics/PredictedParticleProperties/ccn_activation_rates.jl
```diff
@@ -83,8 +84,17 @@ Returns `(; mass, number)` named tuple.
     return (; mass, number = zero(FT))
 end
 
-@inline function compute_ccn_activation(aerosol::AerosolActivation, p3, qᶜˡ, nᶜˡ, nᵃ,
-                                        qᵛ, qᵛ⁺ˡ, T, ρ, constants)
+# Fixed population: `ℳ.nᵃ` is zero here, as in the prescribed-Nᶜˡ path, so the rate is
+# taken against the whole distribution instead.
+@inline function compute_ccn_activation(aerosol::AerosolActivation{FT, false}, p3, qᶜˡ, nᶜˡ, nᵃ,
+                                        qᵛ, qᵛ⁺ˡ, T, ρ, constants) where FT
+    result = prognostic_ccn_activation_rate(aerosol, nᶜˡ, qᵛ, qᵛ⁺ˡ, T)
```
glwagner, line 91: do we have the concept of "activation" separate from "ccn" (cloud condensation nuclei)?

src/Microphysics/PredictedParticleProperties/p3_microphysical_state.jl
```diff
@@ -178,13 +183,14 @@ AM.negative_moisture_correction(p3::P3) = p3.negative_moisture_correction
 # aerosol container so each tuple folds to a constant. The prescribed-Nᶜˡ path has no
 # `ρnᶜˡ`/`ρnᵃ` fields at all, so there is nothing to repair.
 @inline cloud_number_correction_pairs(::Nothing, μ) = ()
-@inline cloud_number_correction_pairs(_, μ) = ((μ.ρnᶜˡ, μ.ρqᶜˡ),)
+@inline cloud_number_correction_pairs(::AerosolActivation, μ) = ((μ.ρnᶜˡ, μ.ρqᶜˡ),)
```
glwagner, line 186:
```suggestion
@inline cloud_number_correction_pairs(::AerosolActivation, μ) = tuple((μ.ρnᶜˡ, μ.ρqᶜˡ))
```

a bit easier to read, esp with the nested tuple

src/Microphysics/PredictedParticleProperties/p3_microphysical_state.jl
```diff
@@ -178,13 +183,14 @@ AM.negative_moisture_correction(p3::P3) = p3.negative_moisture_correction
...
 
 @inline cloud_number_correction_fields(::Nothing, μ) = ()
-@inline cloud_number_correction_fields(_, μ) = (μ.ρnᶜˡ,)
+@inline cloud_number_correction_fields(::AerosolActivation, μ) = (μ.ρnᶜˡ,)
```
glwagner, line 189:
```suggestion
@inline cloud_number_correction_fields(::AerosolActivation, μ) = tuple(μ.ρnᶜˡ)
```

src/Microphysics/PredictedParticleProperties/p3_microphysical_state.jl
```diff
@@ -323,14 +331,19 @@ end
...
 
-@inline aerosol_activation_fields(_, grid) =
-    (; ρnᶜˡ = CenterField(grid),        # Cloud number density [1/m³]
-     ρnᵃ = CenterField(grid),         # Unactivated aerosol number density [1/m³]
-     nᶜˡ = CenterField(grid),         # Cloud number concentration [kg⁻¹]
-     nᵃ = CenterField(grid))          # Unactivated aerosol [kg⁻¹]
+@inline aerosol_activation_fields(aerosol::AerosolActivation, grid) =
+    merge((; ρnᶜˡ = CenterField(grid),  # Cloud number density [1/m³]
+           nᶜˡ = CenterField(grid)),    # Cloud number concentration [kg⁻¹]
+          aerosol_reservoir_fields(aerosol, grid))
```
glwagner, line 340:
```suggestion
           aerosol_reservoir_fields(aerosol, grid))
```

src/Microphysics/PredictedParticleProperties/p3_microphysical_state.jl
```diff
@@ -323,14 +331,19 @@ end
...
+
+@inline aerosol_reservoir_fields(::AerosolActivation{FT, false}, grid) where FT = (;)
```
glwagner, line 342: what does "reservoir" means here? should it just be `aerosol_fields` or are there other types of aerosol fields?

src/Microphysics/PredictedParticleProperties/p3_microphysical_state.jl
```diff
@@ -323,14 +331,19 @@ end
...
-# or `ρnᵃ`.
+# or `ρnᵃ`. A fixed population needs neither `ρnᵃ` nor the `nᵃ` diagnostic mirroring it.
 @inline aerosol_activation_fields(::Nothing, grid) = (;)
 
-@inline aerosol_activation_fields(_, grid) =
-    (; ρnᶜˡ = CenterField(grid),        # Cloud number density [1/m³]
-     ρnᵃ = CenterField(grid),         # Unactivated aerosol number density [1/m³]
-     nᶜˡ = CenterField(grid),         # Cloud number concentration [kg⁻¹]
-     nᵃ = CenterField(grid))          # Unactivated aerosol [kg⁻¹]
+@inline aerosol_activation_fields(aerosol::AerosolActivation, grid) =
```
glwagner, line 337: is this technically correct to think of cloud number as an aerosol activation field? In principle we can run completely aerosol-free initial value problems / freely-evolving simulations, right? Also, isn't homogeneous droplet nucleation possible?

src/Microphysics/PredictedParticleProperties/p3_microphysical_state.jl
```diff
@@ -408,11 +421,14 @@ end
...
-@inline grid_cloud_droplet_number(p3::P3, _, μ, i, j, k, ρ) =
+@inline grid_cloud_droplet_number(p3::P3, ::AerosolActivation, μ, i, j, k, ρ) =
     @inbounds μ.ρnᶜˡ[i, j, k] / ρ
 
 @inline grid_aerosol_number(::Nothing, μ, i, j, k, ρ) = 0 * ρ
-@inline grid_aerosol_number(_, μ, i, j, k, ρ) = @inbounds μ.ρnᵃ[i, j, k] / ρ
+@inline grid_aerosol_number(::AerosolActivation{FT, false}, μ, i, j, k, ρ) where FT = 0 * ρ
+
+@inline grid_aerosol_number(::AerosolActivation{FT, true}, μ, i, j, k, ρ) where FT =
+    @inbounds μ.ρnᵃ[i, j, k] / ρ
```
glwagner, line 431:
```suggestion
@inline grid_aerosol_number(::AerosolActivation{FT, true}, μ, i, j, k, ρ) where FT = @inbounds μ.ρnᵃ[i, j, k] / ρ
```

src/Microphysics/PredictedParticleProperties/p3_microphysical_state.jl
```diff
@@ -800,8 +824,15 @@ end
...
-@inline function add_p3_cloud_number_tendencies!(G, i, j, k, _, result)
+@inline function add_p3_cloud_number_tendencies!(G, i, j, k, aerosol::AerosolActivation, result)
     @inbounds G.ρnᶜˡ[i, j, k] += result.tendency_ρnᶜˡ
+    add_p3_aerosol_tendency!(G, i, j, k, aerosol, result)
+    return nothing
+end
+
+@inline add_p3_aerosol_tendency!(G, i, j, k, ::AerosolActivation{FT, false}, result) where FT = nothing
+
+@inline function add_p3_aerosol_tendency!(G, i, j, k, ::AerosolActivation{FT, true}, result) where FT
```
glwagner, line 835: what role does `FT` play here?

src/Microphysics/PredictedParticleProperties/p3_microphysical_state.jl
```diff
@@ -800,8 +824,15 @@ end
 # Configurations without a prognostic have no `Gⁿ` slot to add to.
 @inline add_p3_cloud_number_tendencies!(G, i, j, k, ::Nothing, result) = nothing
 
-@inline function add_p3_cloud_number_tendencies!(G, i, j, k, _, result)
+@inline function add_p3_cloud_number_tendencies!(G, i, j, k, aerosol::AerosolActivation, result)
     @inbounds G.ρnᶜˡ[i, j, k] += result.tendency_ρnᶜˡ
+    add_p3_aerosol_tendency!(G, i, j, k, aerosol, result)
```
glwagner, line 829: I'm wary of omitting `grid` (eg used to seeing `i, j, k, grid` for these functions). It can be a good idea to keep a convention

src/Microphysics/PredictedParticleProperties/p3_microphysical_state.jl
```diff
@@ -569,9 +585,17 @@ end
 # Configurations without the prognostic have no field to write.
 @inline write_cloud_number_diagnostics!(μ, i, j, k, ::Nothing, ℳ) = nothing
 
-@inline function write_cloud_number_diagnostics!(μ, i, j, k, _, ℳ)
+@inline function write_cloud_number_diagnostics!(μ, i, j, k, aerosol::AerosolActivation, ℳ)
```
glwagner, line 588:
```suggestion
@inline function write_cloud_number_diagnostics!(μ, i, j, k, grid, aerosol::AerosolActivation, ℳ)
```

let's consider using this convention. the grid can also be used to extract `eltype(grid)` if nothing else

src/Microphysics/PredictedParticleProperties/p3_microphysical_state.jl
glwagner, line 829:
```suggestion
    add_aerosol_tendency!(G, i, j, k, grid, aerosol, result)
```

we could consider standardization of how this is done, to reduce the number of function names. Then this function dispatches on the type of `aerosol`. this could reduce code duplication.

glwagner (approved):

src/Microphysics/PredictedParticleProperties/aerosol_activation.jl
kaiyuan-cheng: Done.

src/Microphysics/PredictedParticleProperties/ccn_activation_rates.jl
kaiyuan-cheng: Here two concepts are mixed up. prognostic cloud droplet number and prognostic aerosol (CCN). I've revised the code to make it clear.

src/Microphysics/PredictedParticleProperties/p3_microphysical_state.jl
kaiyuan-cheng: Currently, the prognostic cloud droplet number is always coupled to aerosol, regardless it's prescribed or prognostics. However, you are right, cloud droplet number doesn't have to be coupled with aerosol activation field. I made cloud number no longer filed under aerosol activation.

src/Microphysics/PredictedParticleProperties/p3_microphysical_state.jl
kaiyuan-cheng: "Reservoir" was only there to distinguish the unactivated pool from the droplet-number fields. I've cleaned the code and renamed it to `aerosol_fields`, and the droplet-number half is now `cloud_number_fields`.

kaiyuan-cheng:
close #1003

